PR #3, opened 2021-08-01
Pauses in the blips to match the pauses in the text (e.g. commas, stutters, emphasizing individual words)

notexplosive:
This makes it so the sound effect only plays when the character being displayed is alphanumeric.

I highly recommend you follow the commits to understand where each change is coming from. Changes noted as "Refactor" are behavior preserving. One of the refactors I did was to add `this.` on all the field. I did _not_ do this manually. My formatter did it automatically. It seems we're inconsistent on when we use `this.` and when we don't. Consistency is better and I prefer making `this.` obvious.

### DIFF
--- a/Assets/Scripts/TextArchitect.cs
+++ b/Assets/Scripts/TextArchitect.cs
@@ -1,7 +1,7 @@
 ﻿using System.Collections;
 using System.Collections.Generic;
-using UnityEngine;
 using TMPro;
+using UnityEngine;
 public class TextArchitect {
     /// <summary>
     /// A dictionary keeping tabs on all architects present in a scene. Prevents multiple architects from influencing the same text object simultaneously.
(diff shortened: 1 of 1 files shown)


JuiceGrape:
Why all the refactoring with this. And movement of existing code? It doesn't seem to serve any purpose but change an already established coding style to fit your own, and changes the git blame of lots of lines to yours. I'd recommend changing it back and adhering to the existing coding standard in the future.

IsaacLaquerre:
works for me ez

JuiceGrape:
I'm gonna close this pr for now, you should still have the commits on your fork. This is because this is a feature I would rather see in the develop branch instead of the master branch. The other 2 commits we have right now should be added to master before we make a develop branch, so we'll start with that

PR #4, opened 2021-08-01
Song Title Toast on song transition

notexplosive:
## The Problem ##
Denny and Arnold didn't acknowledge the music the team worked so hard on!

## The Solution ##
When a new track plays, a "Toast" scrolls into view at the top of the screen that displays the name of the currently playing song. This draws the players attention briefly to the fact that the song has changed, and hints to what the song is based on.

Example Gif

## Future Work ##
Currently the song title is displayed as just the asset name in the editor, this isn't a good player-facing name. So we'd either need to commit to naming songs in a customer-facing way in the editor files, or have some mapping between the asset and the "actual" song name.

This also doesn't support long strings, so if a song name were something super long like: `Tutorial Boy's Tragically Generic Reminiscence`, this will do the wrong thing. I imagine it like a scrolling ticker for long strings. I think TextMeshPro has some out-of-the-box way of doing that but I'm not sure what it is.

Graphics-wise: it looks _really_ rough. Gof and Disco said they'd be willing to polish it up, so this will probably need a new asset. We might be able to do some clever asset reuse with the text box assets but... we'll see.

### DIFF
--- a/Assets/Scripts/AudioManager.cs
+++ b/Assets/Scripts/AudioManager.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using UnityEngine;
@@ -21,6 +22,8 @@ public class AudioManager : MonoBehaviour
         }
     }
 
+    public event Action<string> SongStarted;
+
     // MY VARIABLES -------------------------------------------------------------------------------------------------------
     private AudioSource activeSong = null;
 
@@ -40,9 +43,15 @@ public class AudioManager : MonoBehaviour
 
     public void PlaySong(string songName, float volume = 1f, bool loop = true)
     {
+        OnSongStarted(songName);
         AudioClip songToPlay = Resources.Load("Audio/Music/" + songName) as AudioClip;
         SetActiveSong(songToPlay, volume, loop);
         Play();
+    }
+
+    private void OnSongStarted(string songName)
+    {
+        SongStarted?.Invoke(songName);
     }
 
     public void SetActiveSong(AudioClip song, float volume = 1f, bool loop = true)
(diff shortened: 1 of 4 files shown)


Assets/Scripts/AudioManager.cs
```diff
@@ -21,6 +22,8 @@ void Awake()
         }
     }
 
+    public event Action<string> SongStarted;
```
JuiceGrape, line 25: Please use unity events instead

Assets/Scripts/MusicToast.cs
```diff
@@ -0,0 +1,113 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+/// <summary>
+/// This is the "Now Playing" thing that appears at the top of the screen when the music track changes.
+/// 
+/// It's called a "Toast" because it pops out like toast from a toaster.
```
JuiceGrape, line 9: Seems a bit obscure. Better to keep naming a bit more serious

JuiceGrape (approved):
Please do check out my comments. Code looks good. Still don't agree with the code changes, but we can merge them for now until we make a proper decision

Assets/Scripts/MusicToast.cs
notexplosive: Toast is what Android calls them: https://developer.android.com/guide/topics/ui/notifiers/toasts

notexplosive:
FYI - I will probably end up closing this PR-- the idea of a Song-title-toast was something I pitched in the Discord and used this to demonstrate.

This is mostly just an exploration/proposal in how difficult it would be to implement.

PR #5, opened 2021-08-01
Upgraded Unity to 2020 LTS version. Slightly untested

JuiceGrape:
Updated to 2020 LTS for stability and to make it easier to work together. Should support all required features and seems to work well


notexplosive:
I haven't tested this but I can't imagine we'll run into any major issues. I think the one thing I'd like you to do is update README.md to point to 2020 LTS.

notexplosive (changes requested):
Maybe I'm dumb but I couldn't figure out how to fetch this branch, I think it's because the branch actually lives on your fork and not the repo itself (you can just your remote origin url to this one now if you want).

So what I did was a downloaded the zip straight from GitHub, downloaded Unity LTS and loaded the project and.... error! Unity wanted me to open the project in Safe Mode and then I saw this:

image

Digging a bit deeper I think the problem is in GitHub for Unity.

image

Maybe we just need to upgrade GitHub for Unity? Or maybe I messed up somehow.

If you wanna repro try downloading the zip from GitHub (via the "clone" dropdown), unzip and load it.

TooManyMoths (approved):
Opens without errors on Unity version 2010.3.15f2
SampleScene plays without errors
I opened the project by clicking Open Width > Open In Desktop, then fetching the files in Github Desktop and opening through Unity Hub.

PR #18, opened 2021-08-16
Feature/input system

PhoebeMitchell:
## Summary

Adds the Input System package, along with code to create an Input Actions object (PlayerInput.cs)
Unity's new Input System will make supporting other devices easier (gamepads/mobile), and allow controls to be configurable at runtime.
Also updates the .gitignore to ignore Rider's cache directory.

## Testing instructions

Check scenes with event systems run as expected. Event systems may need to be updated to use the new input system.

## Additional information

- `[ ]` Changes UI


notexplosive:
How do I review this? I don't really see any code changes? I'm inclined to just say "yeah that's probably right" and hit approve.

lyndsiWilliams:
> How do I review this? I don't really see any code changes? I'm inclined to just say "yeah that's probably right" and hit approve.

I think that's fine, there won't always be something wrong. Just your stamp of "Hey I looked through this and it looks good to me" is a good review, it still shows that another pair of eyes checked the code and _didn't_ find anything 😁

Assets/Scripts/Input/Controls.cs
```diff
@@ -0,0 +1,147 @@
...
+    ""controlSchemes"": [
+        {
+            ""name"": ""Mouse and Keyboard"",
+            ""bindingGroup"": ""Mouse and Keyboard"",
+            ""devices"": []
+        }
+    ]
+}");
+        // Player
+        m_Player = asset.FindActionMap("Player", throwIfNotFound: true);
```
lyndsiWilliams, line 54: Member variables should start with just `_`, the m should be removed. [Code style doc reference](https://docs.google.com/document/d/1zN4Yx62PpyXhu1g_AhtTZSC8k9e4zqWbzQPt1Hlw9Ag/edit#heading=h.bhrnxu6997n8)

lyndsiWilliams (changes requested):
Just a code style nit 😁

Assets/Scripts/Input/Controls.cs
bas-mentjot: This is an auto generated script, best to leave it alone

PhoebeMitchell: This class is generated by Input System, so I think its best to leave it as is.

PhoebeMitchell:
> How do I review this? I don't really see any code changes? I'm inclined to just say "yeah that's probably right" and hit approve.

It just adds the Input System package, so pretty much just check if there are any errors.
PlayerInput is the only script I created.

Assets/Scripts/Input/Controls.cs
lyndsiWilliams: Oooo are the `m`s automatic? That could get confusing, maybe we should revisit the `m_` naming rule

lyndsiWilliams (approved):

PhoebeMitchell:
Closed because it is now part of feature/Main-Menu

Issue #19, opened 2021-08-18
The current playing song can change via the Audio Controller


notexplosive:
## Description

Build an hook up the AudioController

## Acceptance Criteria

This story will be finished when:
- A song can be played via the audio manager
- The current track can be changed with the audio manager, fading out the current track and fading in the new one

PR #21, opened 2021-08-19
#19 - Audio Controller

notexplosive:
## Summary

Originally this was meant to be just the "Music Player" side of the Audio Controller, but playing sound effects was pretty trivial by comparison so I included that too.

#### Music
Music is played as a single audio source attached directly to the Controller's GameObject, when you play a new track it fades out (over 1 second), then switches track, then fades in (also over 1 second). This does not (yet) support crossfade (fade out A while fading in B). Although I could make some adjustments to get us there if the sound team wants it.

There's also a "SettingsVolume" which is currently a field on AudioController but should eventually come from some core "Settings" object. This scales the total volume tween for the fader, so the fade always takes 1 second, but the amount that it fades in that time will vary depending on the settings volume.

#### Sound Effects
~~Sound Effects work slightly differently, since you apparently need an AudioSource (and therefore an entire game object) to play a sound. I create a child object on AudioController for each sound effect played. However, the game objects stick around after the sound is played, and they're reused if the sound is played multiple times.~~

SoundEffects use PlayOneShot on the existing AudioSource

#### Debug Stuff

Currently there's some clearly marked debug logic that starts out on a particular track and changes the track when you press A or B, and plays a sound when you press C. I can remove this before the final submit, but I left it there to make it easy for people to test.

AudioControllerExample

## Testing instructions

Load up the `AudioController - Test Scene` scene and press `A` and `B` to switch between two tracks. Move the SettingsMusicVolume slider up and down while the music is playing. Try adjusting it while it's fading too and see if you get the expected result.

## Additional information

- `[ ]` Changes UI
- `[X]` Introduces new feature
- `[X]` Has associated resource:  #19


Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -0,0 +1,137 @@
...
+        }
+        else
+        {
+            _directorActionDecoder.SetAudioController(this);
+        }
+
+        _musicFader = new MusicFader();
+
+        _audioSource = GetComponent<AudioSource>();
+        Debug.Assert(_audioSource != null, "AudioController is missing missing AudioSource Component");
```
lyndsiWilliams: Is this meant to have "missing" twice?

lyndsiWilliams (approved):

Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -0,0 +1,137 @@
...
+{
+    [Tooltip("Attach the action decoder object here")]
+    [SerializeField] DirectorActionDecoder _directorActionDecoder;
+    /// <summary>
+    /// One day this will come from the "Settings," but for now it lives on a field
+    /// </summary>
+    [Tooltip("Maximum allowed volume by the game")]
+    [Range(0f, 1f)]
+    [SerializeField] private float _settingsMusicVolume = 1f;
+    [SerializeField] private bool _isTransitioningMusicTracks;
```
PhoebeMitchell: Does this need to be serialized? Would the user ever need to set it manually?

Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -0,0 +1,137 @@
...
+        if (IsCurrentlyPlayingMusic())
+        {
+            yield return _musicFader.FadeOut();
+        }
+
+        SetCurrentTrack(songName);
+        yield return _musicFader.FadeIn();
+
+        _isTransitioningMusicTracks = false;
+        yield return null;
```
PhoebeMitchell: Unnecessary yield. Coroutine is already over.

Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -0,0 +1,137 @@
...
+        {
+            var audioSourceGameObject = new GameObject(string.Format("SFX_{0}", soundEffectName));
+            audioSourceGameObject.transform.SetParent(transform);
+            var source = audioSourceGameObject.AddComponent<AudioSource>();
+            _cachedAudioSources.Add(soundEffectName, source);
+            source.clip = soundEffectClip;
+        }
+
+        _cachedAudioSources[soundEffectName].Play();
+    }
```
PhoebeMitchell, line 87: Making a new audio source for every sound effect feels like an over-complication. Sound effects can be played using PlayOneShot without interrupting the currently playing audio clip. Perhaps make one audio source devoted to sound effects in case they need to make use of AudioSource settings such as pitch.

Assets/Scripts/AudioController/MusicFader.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    /// <summary>
+    /// To be used in a coroutine to fade in the track
+    /// </summary>
+    /// <returns>WaitUntil object that returns control to the coroutine when it's finished</returns>
+    public WaitUntil FadeIn()
+    {
+        return new WaitUntil(() =>
+        {
+            NormalizedVolume += Time.deltaTime;
```
PhoebeMitchell: Maybe make a multiplier and have a serialized field in AudioController so the fade time can be adjusted.

PhoebeMitchell (changes requested):
Some suggestions, and a request to rework how sound effects are played.

Assets/Scripts/AudioController/MusicFader.cs
notexplosive: That's... what it already does? AudioController takes the "normalized" value and multiplies it by the volume level scalar.

Assets/Scripts/AudioController/AudioController.cs
notexplosive: I fixed fixed the typo typo

Assets/Scripts/AudioController/AudioController.cs
notexplosive: It was visible so it was easier to monitor while testing. Un-setting.

Assets/Scripts/AudioController/MusicFader.cs
notexplosive: Done. I went with a divisor instead of multiplier so you're setting a "duration" and not a "speed," I think it's more intuitive this way.

PhoebeMitchell (approved):

PR #26, opened 2021-08-23
Feature/actor controller

### LINKED ISSUE #25: Actor Controller can change active actor and set their emotion
This story will be complete when the Actor Controller component can change active (currently visible) actor and change their emotion. Both must be achieved by taking a string input.

PhoebeMitchell:
## Summary

Implements methods in ActorController (excluding SetActiveSpeaker). Created ActorData scriptable object which holds data about each actor. ActorController loads instances of these objects for each actor and passes it to an Actor class which handles animations and sprites for individual actors. ActorData instances are loaded from an ActorDictionary scriptable object which lists all the actors available.

Modifies DirectorActionDecoder so that the onNewLine event is invoked by the methods tasked with carrying out actions, rather than being invoked by the OnNewActionLine method. Added the PLAY_ANIMATION. This works the same way as the EMOTION actions, however it waits for the animation finish before moving on to the next action. Animations without an OnAnimationComplete event will play indefinitely.

Implemented the WAIT action. This will calls start a coroutine on the scene controller which calls an event after a specified time to call the next line. Calling the next line in the dialogue controller will call the onNextLine event which cancels the current wait coroutine.

Created a template animator controller of blank animations. Actors have animator overrides which they use to play their own animations.

dialoguedemo3
The above GIF showcases the PLAYANIMATION and WAIT actions.

## Testing instructions

Open "Actor- TestScene" and play. Press space to progress through the demo.

## Additional information

- `[ ]` Changes UI


Assets/Resources/InkDialogueScripts/ActorTestInk.ink
```diff
@@ -0,0 +1,23 @@
...
+
+&EMOTION:Normal
+&WAIT:1
+
+&ACTOR:Jory
+&SPEAK:Jory
+Uh... um... Dan I...
+
+&ACTOR:Dan
+&EMOTION:Lean
```
JuiceGrape, line 23: Forgot a &SPEAK:Dan

Assets/Scripts/Interfaces/IActorController.cs
```diff
@@ -1,5 +1,7 @@
 public interface IActorController
 {
+    bool Animating { get; set; }
```
JuiceGrape: Please make sure that this is only gettable, not settable publicly. If the system is showing an animation should only ever be set by the ActorController itself

Assets/Scripts/ActorController.cs
```diff
@@ -17,32 +29,60 @@ void Start()
...
     {
-        Debug.LogWarning("HideActor not implemented");
+        _spriteRenderer.enabled = false;
     }
-
+    
     public void SetActiveActor(string actor)
     {
-        Debug.LogWarning("SetActiveActor not implemented");
+        _activeActor = Resources.Load<ActorData>($"Actors/{actor}");
```
JuiceGrape: I'd prefer we have a list of these in editor instead of loading them on runtime like this, but I'd like a second opinion on this from a secondary reviewer.

Assets/Scripts/ActorController.cs
```diff
@@ -17,32 +29,60 @@ void Start()
...
+        if (_activeActor == null)
+        {
+            Debug.LogError($"Actor \"{actor}\" could not be found.");
```
JuiceGrape: Kudos on the error logging though. Happy to see this

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -36,6 +36,12 @@ public void OnNewActionLine(string line)
         string action = actionAndParam[0];
         string parameters = actionAndParam[1];
 
+        StopAllCoroutines();
+        StartCoroutine(PerformActions(action, parameters));
+    }
+
+    private IEnumerator PerformActions(string action, string parameters)
```
JuiceGrape: Please remove the coroutine functionality from this, as we will probably set this class to be fully separate from unity once we get a script checker in editor.

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -53,7 +59,8 @@ public void OnNewActionLine(string line)
             case "CAMERA_SET": SetCameraPosition(parameters); break;
             case "SHAKESCREEN": ShakeScreen(parameters); break;
             case "SCENE": SetScene(parameters); break;
-            case "WAIT": Wait(parameters); break;
+            case "WAIT_FOR_ANIMATION": yield return StartCoroutine(WaitForAnimation()); break;
```
JuiceGrape: Please move this functionality to the ActorController

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -53,7 +59,8 @@ public void OnNewActionLine(string line)
...
+            case "WAIT": yield return StartCoroutine(Wait(parameters)); break;
```
JuiceGrape: Please move this functionality to the SceneController

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -36,6 +36,12 @@ public void OnNewActionLine(string line)
         string action = actionAndParam[0];
         string parameters = actionAndParam[1];
 
+        StopAllCoroutines();
```
JuiceGrape: General remark for this file: onActionDone should only be invoked on actions that are performed instantly. Probably best to move it to the functions that also handle the decoding, so external controllers can just call OnNextLine in the dialogueController themselves

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -36,6 +36,12 @@ public void OnNewActionLine(string line)
...
+        StartCoroutine(PerformActions(action, parameters));
```
JuiceGrape: General remark: Please revert the changes the formatter made. I'd hate for this file to be re-done fully by every programmer that touches it by the coding style their formatter is set to.
Side note: Turn off your formatter

Assets/Scripts/ActorController.cs
```diff
@@ -1,14 +1,26 @@
...
-    void Start()
+    public bool Animating { get; set; }
+    
+    private SpriteRenderer _spriteRenderer;
+    private Animator _animator;
+    private ActorData _activeActor;
+
+    private void Awake()
     {
+        _spriteRenderer = GetComponent<SpriteRenderer>();
```
JuiceGrape: Get the active gameobject representing the actor from the scene controller, and then apply actor changes an animations to that instead, so we can have multiple gameobjects in parallel, with only the one being shown being active,

Assets/Scripts/ActorController.cs
```diff
@@ -17,32 +29,60 @@ void Start()
         {
             _directorActionDecoder.SetActorController(this);
         }
-
     }
 
     public void ShowActor()
     {
-        Debug.LogWarning("ShowActor not implemented");
+        _spriteRenderer.enabled = true;
```
JuiceGrape: With the changes to the active gameobject for the actor being set in the scene manager, showing the actor might be better suited for the scene controller instead of the actor controller. Please move it there and then leave it to be implemented when the scene controller gets implemented.

Assets/Resources/InkDialogueScripts/ActorTestInk.ink
```diff
@@ -0,0 +1,23 @@
...
+
+&SPEAK:Dan
+Hey good lookin'.
+
+&EMOTION:Lean
+&SPEAK:Dan
+Watch this!
+
+&EMOTION:AirGuitar
+&WAIT_FOR_ANIMATION:
```
JuiceGrape: Seeing as this has no parameters, should we maybe change the decoder to handle a function without a ':', or should we make sure this has some parameters? (I think the first one is better)

JuiceGrape (changes requested):
Very good start, but gonna need some changes. As you mentioned, there were one or two things that could have been prevented if I had the scrum board in order, so my apologies for that. 
Kudos for the error handling and catching the "wait" and "wait_for_end_of_animation" functionality though

Assets/Scripts/ActorController.cs
PhoebeMitchell: I also made a version of this where all the actors were listed in a scriptable object. But was unsure whether it was better to load all the actors into memory like that, or just load them as needed.

alipapapupu: Yeah, I agree. Having hard reference to the objects should be used whenever possible. It makes sure the project stays clean, makes it easier to see all the ActorDatas the game is currently using, keeps the build size smaller and is overall just better option to use on this.

Assets/Scripts/ActorData.cs
```diff
@@ -0,0 +1,14 @@
+using UnityEngine;
+
+[CreateAssetMenu(fileName = "New Actor Data", menuName = "Actors/Actor Data")]
+public class ActorData : ScriptableObject
+{
+    [field: SerializeField, Tooltip("The actor's sprite that will appear in the profiles menu.")]
+    public Sprite Profile { get; private set; }
```
alipapapupu: Since these are public, they don't necessarily need the SerializeField. Not wrong to have them, just a remark.

Assets/Scripts/ActorController.cs
```diff
@@ -17,32 +29,60 @@ void Start()
         {
             _directorActionDecoder.SetActorController(this);
         }
-
     }
 
     public void ShowActor()
```
alipapapupu: Why isn't the method commented properly?

alipapapupu (approved):
Overall seems good, didn't really find too many improvements that werent mentioned already.

Assets/Scripts/ActorController.cs
JuiceGrape: General remark for this code

Assets/Scripts/ActorController.cs
bas-mentjot: A scriptable object seems a bit overkill. What about a prefab?

Assets/Scripts/ActorData.cs
PhoebeMitchell: They are properties so need [field: SerializeField] to be serialized. They are properties to prevent their value being changed by scripts on runtime.

Assets/Scripts/ActorController.cs
PhoebeMitchell: Done. Also moved HideActor to SceneController, updated interfaces, and updated SetActorVisibility method to reflect the change.

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
PhoebeMitchell: Done

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
PhoebeMitchell: Done

Assets/Scripts/ActorController.cs
PhoebeMitchell: Changed back to the scriptable object solution.

Assets/Scripts/Interfaces/IActorController.cs
PhoebeMitchell: Done

Assets/Resources/InkDialogueScripts/ActorTestInk.ink
PhoebeMitchell: Added

Assets/Scripts/ActorController.cs
PhoebeMitchell: Added summaries to methods with more than one line.

Assets/Resources/InkDialogueScripts/ActorTestInk.ink
PhoebeMitchell: Actions without parameters no longer require a ':'

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
PhoebeMitchell: Moved onActionDone to the methods called in the switch statement, and OnNextLine is called directly when animations are finished playing.

Assets/Scripts/ActorController.cs
PhoebeMitchell: ActorController now has an ActiveActor property which can be set, and communicates with an actor class which handles playing animations.

JuiceGrape (approved):

Issue #20, opened 2021-08-18
Appearing dialog when a character is speaking


alipapapupu:
## Description
Make dialog appear to the player in controlled manner.
...

## Acceptance Criteria

This story will be finished when:
- The appearing dialog can be controlled with an inky file.
- The speed of dialog and individual letters can be controlled in inky file as well as Editor
- Custom letters can be set with their own speed and actions in Editor
- The code gives information back to caller that dialog has ended
- The player is able to speed up the dialog with input

PR #32, opened 2021-08-26
Feature/20 appearing dialog

alipapapupu:
## Summary

Created script that makes letters appear one at a time to the screen. You are able to change the speed of punctuation, dialog, or overall speed from the inky file. The code has function ready to speed up the dialog when player gives input, and this feature can be disabled/enabled from the inky file. You can prolong the dialog with next one after players input, or make the dialogs skip to next one without players input.

Does not change the speakers name or background color yet. For this I would request new PR implementing Scriptable Object containing the name, color etc. information about the speaker, and this could be used to changed all information on the screen.
 
Example of the dialog appearing
AppearingDialog

## Testing instructions

Open AppearingText-TestScene and hit play. The textbox contains Lorem Ipsum, but after you press space, the dialog starts. Take a look at the AppearingDialogInk-file while testing the dialog to see how they affect the way dialog appears.

## Additional information

- `[x]` Changes UI
- `[x]` Has associated resource:  #20


Assets/Resources/InkDialogueScripts/AppearingDialogInk.ink
```diff
@@ -0,0 +1,46 @@
...
+&CLEAR_SPEED:
+&OVERALL_SPEED:0.02
+&SPEAK:Dan
+DON'T JUST GO FLASHING YOUR HEDGEHOG AROUND!
+
+&SPEAK:Arin
+BUT I DON'T WANNA WEAR PANTS!
+
+&OVERALL_SPEED:0.03
+&FORCE_NEXT_DIALOG:
```
JuiceGrape: I think it's better to replace this with "AUTO_SKIP:True" and "AUTO_SKIP:False". Just having auto skip enabled until we disable it seems like a more logical way of doing this than defining it for every line

Assets/Resources/InkDialogueScripts/AppearingDialogInk.ink
```diff
@@ -0,0 +1,46 @@
...
+&SPEAK:Arin
+FUCK YOU YOU'RE NOT MY REAL MOM!
+
+&FORCE_NEXT_DIALOG:
+&SPEAK:Dan
+GOOD BECAUSE IF I WAS, I WOULD MAKE YOU GO TO BED WITHOUT ANY SHNITZEL!
+
+&CLEAR_SPEED:
+&DISABLE_SKIPPING:TRUE
```
JuiceGrape, line 37: What would DISABLE_SKIPPING:False do? For that matter, what does DISABLE_SKIPPING:True do?

Assets/Scripts/TextDecoder/DialogueController.cs
```diff
@@ -67,12 +71,27 @@ public void OnNextLine()
...
     private bool IsAction(string line)
     {
         return line[0] == ACTION_TOKEN;
         //TODO: Check if line is action
     }
+
+    ///<summary>
+    ///When the dialog has been written, makes the code be able to continue to next dialog
+    ///</summary>
+    public void StopWritingDialog()
```
JuiceGrape: Good way of solving this externally. However, button logic will not be handled in here, as it being in here was for debug purposes only (as denoted by the comment). Maybe create an "inputHandler" class which takes button presses and then triggers functions (through events) in the correct places? 
Basically I think it's best to have all input handling logic in it's own class, separate from the DialogueController

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -13,6 +13,10 @@ public class DirectorActionDecoder : MonoBehaviour
     private IAudioController _audioController;
     private IEvidenceController _evidenceController;
 
+    [Header("Basic Values")]
```
JuiceGrape: Please remove these and add them like we have before. This class needs to be completely agnostic from anything external

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -143,7 +155,7 @@ void FadeInScene(string seconds)
 
         float timeInSeconds;
 
-        if(float.TryParse(seconds, out timeInSeconds))
+        if (float.TryParse(seconds, out timeInSeconds))
```
JuiceGrape: Please refrain from changes like this as they fog up git. Revert if you can

JuiceGrape: In other places as well, I trust your judgement on finding them

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -469,4 +481,84 @@ private bool HasEvidenceController()
...
     #endregion
+
+    #region DialogStuff
+
+
+    /// <summary>
+    /// Checks if the decoder has an appearing dialog controller attached, and shows an error if it doesn't
+    /// </summary>
+    /// <returns>Whether an appearing dialog controller is connected</returns>
+    private bool HasAppearingDialogController()
```
JuiceGrape, line 494: Move to controller stuff

Assets/Scripts/UI/AppearingDialogController.cs
```diff
@@ -0,0 +1,517 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using TMPro;
+using UnityEngine.Events;
+using UnityEngine.UI;
+
+public class AppearingDialogController : MonoBehaviour
```
JuiceGrape: Add interface

Assets/Scripts/UI/AppearingDialogController.cs
```diff
@@ -0,0 +1,517 @@
...
+    [SerializeField, Tooltip("When player is giving correct input to the game, how much faster should the game go.")]
+    private float _speedMultiplierFromPlayerInput = 2;
+
+    [Header("Events")]
+    [SerializeField, Tooltip("Events that should happen after completing a dialog.")]
+    private UnityEvent _dialogDoneEvent = new UnityEvent();
+    ///<summary>
+    ///Events that should happen after completing a dialog. You have to manually add and remove listeners, and they will happen every time dialog is complete.
+    ///</summary>
+    public UnityEvent DialogDoneEvent { get { return _dialogDoneEvent; } }
```
JuiceGrape: Why does this exist? We should not be linking events in code, unless in special cases. This has no reason to be public

Assets/Scripts/TextDecoder/DialogueController.cs
JuiceGrape: Quick thought: We should probably make it a state machine

Assets/Scripts/UI/AppearingDialogController.cs
```diff
@@ -0,0 +1,517 @@
...
+    }
+
+    ///<summary>
+    ///Checks how long the number continues from the start for string.
+    ///</summary>
+    /// <param name="dialogText">The dialog we check the number from.</param>
+    /// <param name="previousTime">How fast the dialog was going previously, in case the number should increase or decrease the current speed.</param>
+    /// <param name="newAppearTime">Gives out the new float containing new appear time rule.</param>
+    /// <returns>Returns true if numeric value was retrieved, or false it there was an error or the command didn't contain numeric value.</returns>
+    static private bool GetFloatFromString(string dialogText, float previousTime, out float newAppearTime)
```
JuiceGrape: This description makes no sense to me. What does this function do? The name is kinda bad too, cause getting a float from a string is just float.parse(string)

Assets/Scripts/UI/AppearingDialogController.cs
```diff
@@ -0,0 +1,517 @@
...
+    public void ContinueDialog()
+    {
+        _continueDialog = true;
+    }
+
+
+    ///<summary>
+    ///Small class containing if the current WaitTimer is in use, and how long to wait if it is.
+    ///</summary>
+    private class WaitInformation
```
JuiceGrape: Probably better as a struct?

Assets/Scripts/UI/AppearingDialogController.cs
```diff
@@ -0,0 +1,517 @@
...
+        _allWaiters[waiterTypeToChange].waitTime = newAppearTime;
+    }
+
+    ///<summary>
+    ///Should the next speaken dialog continue after this
+    ///</summary>
+    public void ContinueDialog()
+    {
+        _continueDialog = true;
+    }
```
JuiceGrape: Rest of the class looks great, keep up the good work.

JuiceGrape (changes requested):
Needs some work. Particularly sticking to existing code structures and moving logic out of existing classes where it doesn't belong.

Assets/Resources/InkDialogueScripts/AppearingDialogInk.ink
alipapapupu: DISABLE_SKIPPING:True makes it so player cant speed up or skip the dialog at all, rather, it will appear in constant speed. Good if dialog needs to be timed to appear with the animation, otherwise the player could speed up the dialog and screw things up. DISABLE_SKIPPING:False just makes the speeding possible again. If you have better names for this, I would be glad to hear them.

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
alipapapupu: Sure, my bad. I tend to auto format everything, and its in my muscle memory, but I'll see what I can do.

Assets/Scripts/UI/AppearingDialogController.cs
alipapapupu: Legacy code from my original version, I can just delete the whole function.

Assets/Scripts/TextDecoder/DialogueController.cs
PhoebeMitchell: UnityInputSystemInterface is now in the develop branch if you need an input class.

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -469,4 +481,84 @@ private bool HasEvidenceController()
...
+    ///<summary>
+    ///Toggles skipping on or off
+    ///</summary>
+    ///<param name = "disable">Should the text skipping be disabled or not</param>
+    private void DisableTextSkipping(string disabled)
+    {
+        if (!HasAppearingDialogController())
+            return;
+
+        bool value = bool.Parse(disabled);
```
PhoebeMitchell: Consider using TryParse for error checking, and to be consistent with the rest of the script.

Assets/Scripts/UI/AppearingDialogController.cs
```diff
@@ -0,0 +1,517 @@
...
+using UnityEngine;
+using TMPro;
+using UnityEngine.Events;
+using UnityEngine.UI;
+
+public class AppearingDialogController : MonoBehaviour
+{
+    [Header("Basic Values")]
+    [SerializeField, Tooltip("TextMeshPro-component all the dialog should appear in.")]
+    private TextMeshProUGUI _controlledText = null;
```
PhoebeMitchell: nit: Do not need to initialise fields if the value is the field's default value.

Assets/Scripts/UI/AppearingDialogController.cs
```diff
@@ -0,0 +1,517 @@
...
+    {
+        foreach (UnityAction ua in dialogDoneActions)
+        {
+            _temporaryDialogDoneEvent.AddListener(ua);
+        }
+
+        yield return StartDialogCoroutine(dialog);
+
+        _temporaryDialogDoneEvent?.Invoke();
+        _temporaryDialogDoneEvent.RemoveAllListeners();
```
PhoebeMitchell: Add ? for null check, like on previous line.

Assets/Scripts/UI/AppearingDialogController.cs
```diff
@@ -0,0 +1,517 @@
...
+    ///<summary>
+    /// Displays dialog one letter at a time.
+    ///</summary>
+    /// <param name="dialog">The dialog that should be showed to player.</param>
+    /// <param name="dialogDoneActions">All the actions that should be called after the dialog has written everything. These actions happen only once, and will be removed afterwards.</param>
+    public IEnumerator StartDialogCoroutine(string dialog, params UnityAction[] dialogDoneActions)
+    {
+        foreach (UnityAction ua in dialogDoneActions)
+        {
+            _temporaryDialogDoneEvent.AddListener(ua);
```
PhoebeMitchell: Could you just loop through the dialogDoneActions and invoke each one? Instead of subscribing them all to an event.

Assets/Scripts/UI/AppearingDialogController.cs
```diff
@@ -0,0 +1,517 @@
...
+        _currentLetterNum++;
+
+        //Make sure the _currentLetterNum isn't going to be too small.
+        if (_currentLetterNum >= 2)
+        {
+            //_currentLetterNum-2 is to make the text wait longer AFTER the punctuation has been written, rather it taking long time for the punctuation itself to appear.
+            WaiterTypes newWaiter = GetCurrentWaiter(_currentDialog[_currentLetterNum - 2]);
+
+            if (newWaiter == WaiterTypes.letter)
+                _allWaiters[WaiterTypes.letter].inUse = false;
```
PhoebeMitchell: Add braces to if statement.

Assets/Scripts/UI/AppearingDialogController.cs
```diff
@@ -0,0 +1,517 @@
...
+            f += previousTime;
+        else if (decrease)
+            f = previousTime - f;
+
+        newAppearTime = f;
+        return true;
+    }
+
+    ///<summary>
+    ///Read the commands from dialog and seperates them. Prototype, currently not in use.
```
PhoebeMitchell: spelling: separates

Assets/Scripts/UI/AppearingDialogController.cs
```diff
@@ -0,0 +1,517 @@
...
+            //Check if the command character is found
+            if (dialog[i] == _commandCharacter)
+            {
+                insideCommand = !insideCommand;
+
+                if (insideCommand)
+                {
+                    startCommand = i;
+                }
+                else if (!insideCommand)
```
PhoebeMitchell: Use else instead of else if

Assets/Scripts/UI/AppearingDialogController.cs
PhoebeMitchell: By prototype do you mean these methods will be used in future, or that they are no longer used? Can they be removed?

Assets/Scripts/UI/AppearingDialogController.cs
```diff
@@ -0,0 +1,517 @@
...
+        //Make sure the _currentLetterNum isn't going to be too small.
+        if (_currentLetterNum >= 2)
+        {
+            //_currentLetterNum-2 is to make the text wait longer AFTER the punctuation has been written, rather it taking long time for the punctuation itself to appear.
+            WaiterTypes newWaiter = GetCurrentWaiter(_currentDialog[_currentLetterNum - 2]);
+
+            if (newWaiter == WaiterTypes.letter)
+                _allWaiters[WaiterTypes.letter].inUse = false;
+
+            _currentAppearTime = _allWaiters[newWaiter].waitTime;
```
PhoebeMitchell: Punctuation speed affects things it probably shouldn't such as apostrophes, and symbols such as $ and £.

Assets/Scripts/UI/AppearingDialogController.cs
```diff
@@ -0,0 +1,517 @@
...
+        }
+
+        //Is this WaitInformation in use now.
+        public bool inUse = false;
+        //What is the current wait time of this information.
+        public float waitTime = 0;
+        //Is this information used only with punctuations.
+        public bool useOnlyWithPunctuations = false;
+        //Is this information used with everything.
+        public bool useWithEverything = false;
```
PhoebeMitchell: These should be properties, and in PascalCase

Assets/Scripts/UI/AppearingDialogController.cs
```diff
@@ -0,0 +1,517 @@
...
+        _continueDialog = true;
+    }
+
+
+    ///<summary>
+    ///Small class containing if the current WaitTimer is in use, and how long to wait if it is.
+    ///</summary>
+    private class WaitInformation
+    {
+        public WaitInformation() { }
```
PhoebeMitchell: Remove duplicate constructor.

Assets/Scripts/UI/WaiterTypes.cs
```diff
@@ -0,0 +1,16 @@
...
+///Contains all different special actions used in code. Order matters, as topmost value is most important, and lowest least important.
+///</summary>
+public enum WaiterTypes
+{
+    letter,
+    punctuation,
+    defaultPunctuation,
+    dialog,
+    overall,
+    defaultValue,
```
PhoebeMitchell: Consider changing WaiterTypes to WaiterType and using PascalCase for values to follow C# conventions.
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces

PhoebeMitchell (approved):
Some suggestions, but overall the code looks functional, and runs as expected. AppearingDialogController is very big though. Could it be split up? It should also be be Dialogue instead of Dialog, as it's referring to conversation and not computer dialog boxes.

Not sure if having a mask to colour the textbox is necessary, when you could just set the colour of the mask image itself. I feel it would be easier to position text if the name box and the speech box were separate sprites. Then you could just get the speech text to fill the whole size of the speech box, and it would allow for easier position of the name text too.

Also correct the spelling of general in GerenralSpeechPanel. Then make the changes requested by JuiceGrape and it should be good.

Assets/Resources/InkDialogueScripts/AppearingDialogInk.ink
JuiceGrape: Ah good thing to have indeed

Assets/Scripts/UI/AppearingDialogController.cs
JuiceGrape: Legacy code should be removed unless you specifically intent to refactor or re-use it later

Assets/Scripts/UI/AppearingDialogController.cs
JuiceGrape: What is the alternate course for a ? null check? If it's just ignoring the line, I'd prefer to make those null checks explicit in it's own line which aborts any future functionality

Assets/Scripts/UI/AppearingDialogController.cs
JuiceGrape: Also, this seems like state behaviour, so it should be an enum instead of 2 bools. If neither of these can be true at the same time that is

Assets/Scripts/UI/AppearingDialogController.cs
alipapapupu: I know, but Unity used to give warnings in the editor whenever a SerializeField or public value wasn't initialised in the code even though it would get initialised in the editor. They did fix this like a year ago, but it's already in my muscle memory to initialise values with the fields default value, and I will probably keep doing it even though it's a bit pointless.

Assets/Scripts/UI/AppearingDialogController.cs
alipapapupu: It depends on what we decide and what the designer team wants. Right now the speeds can be changed only before the dialog happens. Things markes as prototype would give full control over how the dialog is shown, giving the possibility to change speed values in the middle of the dialog and thus giving possibility to make more it complex, speeding and slowing when wanted. If we deem this unnecessary, I can remove everything marked as prototype, but I would like to hear yours and @JuiceGrape opinion on it, do we/design/writer team need that much control?

Assets/Scripts/UI/AppearingDialogController.cs
PhoebeMitchell: For regular C# events it checks if the event has any listeners and is the same as checking "event != null".
I found out today though it's not necessary for Unity Events. '?' checks if the UnityEvent object itself is null, which it probably isn't if it's serialised.

Assets/Scripts/UI/AppearingDialogController.cs
PhoebeMitchell: I think this feature would be cool, but ask the design team what they're expecting. No point putting the effort in if they don't want it.

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -468,5 +477,112 @@ private bool HasEvidenceController()
...
+    ///<param name = "value">The read bool value given out.</param>
+    ///<returns>True if the string contained bool value, and false if it didn't.</returns>
+    private bool ParseBool(string parameter, out bool value)
+    {
+        if (bool.TryParse(parameter, out value))
+        {
+            return true;
+        }
+
+        Debug.LogError("Bool value wasn't found from command. Please fix.");
```
JuiceGrape: Nondescript error message doesn't help anyone. Put the error message in the decode function, which sorta eliminates the need for having this a separate function.

Assets/Scripts/TextDecoder/DialogueController.cs
alipapapupu: Fixed the issue. DialogueAppearController takes care of the autoSkip feature, and calls the OnNextLine when autoskip should happen. DialogueController also has bool that is changed by DialogueAppearControllers events to make sure dialog doesn't overlap, and it's not connected to the input in any way anymore.

Assets/Scripts/UI/AppearingDialogController.cs
alipapapupu: made private, and deleted extra event

Assets/Scripts/UI/AppearingDialogController.cs
alipapapupu: Refactored the method to work smarter way, and deleted unnecessary legacy code.

Assets/Scripts/UI/AppearingDialogController.cs
alipapapupu: Since the inUse and waitTime-values of the class is changed constantly, this being a struct won't work because that would require recreating the vakkye every time these values need to change.

Assets/Scripts/UI/AppearingDialogController.cs
alipapapupu: I'll leave these in for now, since there might be need for them later, but what we have right now seemed to be enough for people.

Assets/Scripts/UI/AppearingDialogController.cs
alipapapupu: Removed the whole function since it was quite unnecessary

JuiceGrape (approved):

Issue #33, opened 2021-08-29
[BUG] AudioController issues


notexplosive:
## Describe the bug
- Songs are not looping
- Sound effects do not play when volume is zero

PR #35, opened 2021-08-29
Bugfix - AudioController issues

notexplosive:
## Summary
Fixes for #33 

On `AudioController.Start` we create a child game object and parent it to the AudioController, that GameObject has its own SoundSource that is used for playing SFX while the parent object's sound source is for Music.

### Before
- SFX volume was the same as music volume
- Songs wouldn't loop

### After
- SFX volume is now its own value separate from music volume
- Songs now loop

## Testing instructions
In the `AudioController - Test Scene` you now have separate volume sliders for SFX and Music. Press A and B to change tracks, press C to play a sound effect.

## Additional information

- `[ ]` Changes UI
- `[/]` Introduces new feature
  #33


Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -5,17 +5,27 @@ public class AudioController : MonoBehaviour, IAudioController
 {
```
JuiceGrape, line 5: Add a RequireComponent flag to the class to make sure it always has an audio source

Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -32,8 +42,13 @@ void Start()
 
         _musicFader = new MusicFader();
 
-        _audioSource = GetComponent<AudioSource>();
-        Debug.Assert(_audioSource != null, "AudioController is missing AudioSource Component");
+        _musicAudioSource = GetComponent<AudioSource>();
+        Debug.Assert(_musicAudioSource != null, "AudioController is missing AudioSource Component");
+        _musicAudioSource.loop = true;
+
+        var sfxGameObject = new GameObject("SFX Player");
```
JuiceGrape: Why is only the SFX source a child? Probably best to make them both a child and turn this into a function

JuiceGrape (approved):

Assets/Scripts/AudioController/AudioController.cs
JuiceGrape: In which case the other comment is no longer applicable

Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -32,8 +42,13 @@ void Start()
...
+        sfxGameObject.transform.parent = this.transform;
+        _sfxAudioSource = sfxGameObject.AddComponent<AudioSource>();
```
TooManyMoths, line 51: Suggestion: Use "new GameObject("SFX Player", typeof(AudioSource));" to create a new game object with an audio source.

PhoebeMitchell (approved):
Tested in editor. Sound effects play when transitioning songs, and songs loop.

PR #46, opened 2021-09-03
Feature/45 speech options

### LINKED ISSUE #45: More speach options
## Description

When a character is talking, the mouth should only move when the current speaking actor is the same as the actor shown on the screen.
Also, when the spoken actor is none or when the think action is performed instead of the speak action, the mouth should also stay still.

## Acceptance Criteria

This story will be finished when the think action is implemented, an actor doesn't move their mouth if they aren't the actor speaking, and the none character is added so the nameplate can hide itself and mouth movement is done.

JuiceGrape:
## Summary
Added all new speech options (no speaking actor, actor talking if it is the active actor, think instead of talk which makes mouths not move)

NewSpeechOptions

## Testing instructions
Open inky test and play the current added dialogue (advance with x)

## Additional information

- `[ ]` Changes UI

closes #45


Assets/Scripts/Actor/Actor.cs
```diff
@@ -80,4 +80,9 @@ public void SetTalking(bool isTalking)
     {
         _animator.SetBool("Talking", isTalking);
     }
+    
+    public bool isActor(ActorData actor)
```
notexplosive: Nit: name is confusing. Based on how you're using it maybe you want something like: `MatchesActor` or `HasActorData`

Also, capitalization is wrong. `IsActor`
Also, missing summary. 😛

lyndsiWilliams: ✨This method is missing a summary ✨

lyndsiWilliams (approved):
Just one (1) summary! ✨

Assets/Scripts/Actor/Actor.cs
```diff
@@ -80,4 +80,9 @@ public void SetTalking(bool isTalking)
...
+    {
+        return _actorData == actor;
```
notexplosive, line 91: Consider: Is `==` safe here? Does ActorData overload `==` to do the right thing? If ActorData is a ScriptableObject maybe unity does some magic thing for us. Or maybe we're just comparing addresses so we don't need to overload it.

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -124,12 +125,14 @@ private void SetActorVisibility(string showActor)
...
     /// <param name="actor">Actor to make the speaker</param>
-    private void SetSpeaker(string actor)
+    /// <param name="speakingType">Type of speaking to speak the text with</param>
+    private void SetSpeaker(string actor, SpeakingType speakingType)
     {
         if (!HasActorController())
             return;
 
         _actorController.SetActiveSpeaker(actor);
+        _actorController.SetSpeakingType(speakingType);
```
notexplosive, line 135: Consider: Should SetActiveSpeaker and SetSpeakingType be combined into one method? Is there ever a time where we'd want one and not the other?

Assets/Scripts/Actor/Actor.cs
JuiceGrape: Not anymore

notexplosive (approved):
LGTM, few tiny nits and thoughts

Assets/Scripts/Actor/Actor.cs
JuiceGrape: Scriptable objects don't change, so a == always work, they're always a reference to the original object

Assets/Scripts/Actor/Actor.cs
JuiceGrape: Changed to MatchesActorData

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
JuiceGrape: Seemed logical to separate em because there was no proper function name that didn't end up with SetActiveSpeakerAndSpeakingType. At that point, make it 2 functions.

Assets/Scripts/Actor/ActorController.cs
```diff
@@ -157,4 +170,13 @@ public void OnAnimationDone()
         }
         
     }
+
+    /// <summary>
+    /// Sets the speaking type of the sentences shown, so the actor can react appropiately.
```
ViMaSter, line 175: Nit, typo: `appropriately` instead of `appropiately`

PR #49, opened 2021-09-06
Feature/31 witness interrogation

### LINKED ISSUE #31: Witness interrogation
## Description

The first minigame we are implementing is the witness interrogation. It works like it does in the lua build:
- The witness tells their story in full.
- The witness repeats their story, with a possible objection or press for info with every line.
- Only one specific line will have the correct objection, as defined in the inky file
- Every line will have their own press for info

## Dev Guide
- Each interrogation is it's own inky script
- Mark the start and end of a witness story with an action, so the rest can be handled in code
- Probably mark the correct line and correct evidence with an action as well (probably parameters for the start story action)

## Acceptance Criteria

This story will be finished when witness interrogation works in an example scene with 3 example inky files for 3 separate interrogations that follow each other. Animations do not have to be working yet.

JuiceGrape:
## Summary
Added cross examinatin mode as part of the dialogue controller. Also added sub stories as a way of handling failure states, so these can be small self contained inky stories that we can complete before continuing with the main story. Also added the handling of presented evidence / actors.

substory
oim9

## Testing instructions
Open the inky test scene and run through the witness interrogation, using z to open the evidence menu, x to continue normally, and c to press the witness.

## Additional information

- `[ ]` Changes UI
closes #31


lyndsiWilliams (dismissed):
This looks great! Nice work! 🔥

Assets/Scripts/Actor/ActorController.cs
```diff
@@ -118,7 +118,7 @@ public void SetActiveSpeaker(string actor)
         catch (KeyNotFoundException exception)
         {
             _currentSpeakingActor = null;
-            Debug.Log($"{exception.GetType().Name}: Actor was not found in actor dictionary");
+            Debug.Log($"{exception.GetType().Name}: Actor " + actor + " was not found in actor dictionary");
```
PhoebeMitchell: Use string interpolation: $"{actor}"

Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -54,8 +54,11 @@ void Start()
...
     void Update()
     {
-        _musicAudioSource.volume = _musicFader.NormalizedVolume * _settingsMusicVolume;
-        _sfxAudioSource.volume = _settingsSFXVolume;
+        if (_musicAudioSource != null && _musicFader != null)
+            _musicAudioSource.volume = _musicFader.NormalizedVolume * _settingsMusicVolume;
+
+        if (_sfxAudioSource != null)
+            _sfxAudioSource.volume = _settingsSFXVolume;
     }
```
PhoebeMitchell, line 62: I know it's not really relate to this PR but maybe this would be better in an OnValidate method so its only called when a value is changed in the inspector. Or an #if UNITY_EDITOR directive.

Assets/Scripts/TextDecoder/DialogueController.cs
```diff
@@ -11,30 +27,60 @@ public class DialogueController : MonoBehaviour
...
+    [SerializeField] private DialogueControllerMode _dialogueMode = DialogueControllerMode.Dialogue;
+
+    [SerializeField] private FailureStoryList _failureList;
+
+    [SerializeField] private GameObject _dialogueControllerPrefab;
+
     [Header("Events")]
 
     [Tooltip("Attach a scene controller to this so it can cancel 'wait' actions on new lines.")]
     [SerializeField] private UnityEvent _onNewLine; // TODO make this into a custom event
```
PhoebeMitchell: Might as well remove this TODO now the events below are now just regular UnityEvents.
The event itself is unused now we decided the WAIT action cannot be skipped so maybe remove it? Unless you think it will be useful.

Assets/Scripts/TextDecoder/DialogueController.cs
```diff
@@ -124,8 +351,57 @@ private void OnNextLine()
...
+    //Sub story stuff
+
+    /// <summary>
+    /// Starts a new sub story based on the provided inky dialogue script. The sub story always goes before the main story. When the sub story is finished, the main story continues on the line after the one it left off on.
+    /// </summary>
+    /// <param name="subStory">Inky dialogue script to be set as the sub story</param>
+    public void StartSubStory(TextAsset subStory)
+    {
+        GameObject obj = GameObject.Instantiate(_dialogueControllerPrefab);
+        _subStory = obj.GetComponent<DialogueController>();
```
PhoebeMitchell: Micro-optimisation: Make _dialogueControllerPrefab a DialogueController to eliminate this GetComponent.

PhoebeMitchell (dismissed):
Looks good to me. Just a few suggestions.

Assets/Scripts/AudioController/AudioController.cs
JuiceGrape: Probably a bit too early to be thinking about optimizations like that. This was never gonna be a big impact anyway, three checks per frame is pretty easy for a cpu

Assets/Scripts/TextDecoder/DialogueController.cs
JuiceGrape: I'd end up having to get the gameobject from the prefab though, so it ends up the same ;)
But yeah good way to make sure it has a dialogue controller, so will add

Assets/Scripts/Actor/ActorController.cs
JuiceGrape: Added. Didn't even know this was a thing

Assets/Scripts/TextDecoder/DialogueController.cs
JuiceGrape: Removed

Assets/Scripts/TextDecoder/DialogueController.cs
JuiceGrape: Oh actually, I want to target the newly created instance, not just the the one I got earlier, so the get component is still necessary. However, providing a DialogueController is better in this case to make sure it has one

PhoebeMitchell (approved):

lyndsiWilliams (approved):
Lookin' great part 2 electric boogaloo! 🕺

PR #52, opened 2021-09-11
Player can access the profile menu

### LINKED ISSUE #11: Player can access the Profile Menu
Blocked by: #13 

## Description

When in the evidence menu, the player should be able to show the profile menu by pressing the 'c' button (by default). The profile menu works exactly the same as the evidence menu does, but with profiles instead of evidence. (Look at #13)
The profiles get added over the course of the game, and every unity scene should be able to have a list of default profiles that are already added. Whenever a new character is introduced, it gets added to the profile menu.

## Dev Guide

- Profiles are the same scriptable objects as actors are. Please use those instead of making new scriptable objects.
- Try to re-use most of the code the evidence uses. Maybe use a shared base class, or an interface?

## Acceptance Criteria

This story will be finished when the sample scene for the evidence is extended with the profile menu, which can be shown by pressing 'c' while in the evidence menu. It should have 9 profiles to show.

PhoebeMitchell:
## Summary

Updates the evidence menu to support showing profiles. Actor dictionaries now use the same system as evidence (dictionary created on a monobehaviour instead of in the scriptable object). The logic from EvidenceDictionary is now in its own class called ObjectDictionary which the ActorDictionary and EvidenceDictionary monobehaviours use. ActorData and Evidence now implement ICourtRecordObject which allows their names, icons, and descriptions to be accessed. Also features tests for the evidence menu and ObjectDictionary class.

profilesdemo2

## Testing instructions

Open a scene with the StandardInclusion asset and enter play mode. Press z to open the evidence menu, then press c to open the profiles menu. Navigate with the arrow keys (and check for spelling mistakes in each bio).
Open the unity test runner and run the play mode and edit mode tests.
The screenshots below show the tests passing.
Screenshot 2021-09-16 at 14 30 48
Screenshot 2021-09-16 at 14 30 55

## Additional information

- `[x]` Changes UI
  
#11


Assets/Scripts/Actor/ActorDictionary.cs
```diff
@@ -1,25 +1,12 @@
-using System.Collections;
-using System.Collections.Generic;
-using UnityEngine;
-
-[CreateAssetMenu(fileName = "New Actor Dictionary", menuName = "Actors/Actor Dictionary")]
-public class ActorDictionary : ScriptableObject
+public class ActorDictionary : ObjectDictionaryInterface<ActorData, ActorList>, ICourtRecordObjectDictionary
```
JuiceGrape: Kudos: Nice one! Sexy ass implementation, shit like this makes me happy

Assets/Scripts/Actor/ActorData.cs
```diff
@@ -21,4 +33,24 @@ public class ActorData : ScriptableObject
...
     public RuntimeAnimatorController AnimatorController { get; private set; }
+
+    public string CourtRecordName => GenerateNameWithAge();
+    public Sprite Icon => Profile;
+    public string Description => Bio;
+
+    private string GenerateNameWithAge()
+    {
+        string age = "???";
+        switch (_ageCertainty)
```
JuiceGrape: For extendability it is better to leave the default as an error and then handle all known cases in the switch. Also handling the default in the switch is better practice for this reason, so when a new enum gets added and they forget to extend this switch, it is instantly noticeable

Assets/Scripts/Evidence/EvidenceDictionary.cs
```diff
@@ -1,157 +1,22 @@
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using UnityEngine;
-
-public class EvidenceDictionary : MonoBehaviour
+public class EvidenceDictionary : ObjectDictionaryInterface<Evidence, EvidenceList>, ICourtRecordObjectDictionary
```
JuiceGrape: Kudos: New implementation looks much cleaner

JuiceGrape (dismissed):
Great work! Feel free to close the kudos comments when read

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuMouseTests.cs
```diff
@@ -0,0 +1,92 @@
...
+                ICourtRecordObject evidence = menu.SelectedButton.GetComponent<EvidenceMenuItem>().CourtRecordObject;
+                Assert.AreEqual(menuItem.GetComponent<EvidenceMenuItem>().CourtRecordObject.DisplayName, evidence.DisplayName);
+            }
+        }
+        
+        /// <summary>
+        /// Presses the navigation buttons multiple times then checks if
+        /// the first menu item is the same as the one that the menu thinks is selected.
+        /// </summary>
+        // [UnityTest, Order(1)]
```
ViMaSter: This test appears to be commented out and doesn't show up on the Test Runner. I'd suggest either removing the code for it or integrating it somehow!

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuMouseTests.cs
```diff
@@ -0,0 +1,92 @@
+using System.Collections;
+using System.Linq;
+using NUnit.Framework;
+using Tests.PlayModeTests.Tools;
+using TMPro;
```
ViMaSter: Nit: This namespace seems to be unused in this file. I'd suggest removing it.

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuMouseTests.cs
```diff
@@ -0,0 +1,92 @@
...
+            MenuItem[] menuItems = GameObject.Find("EvidenceContainer").GetComponentsInChildren<MenuItem>();
+            EvidenceMenuItem firstMenuItem = menuItems.First(menuItem => menuItem.gameObject.name == "EvidenceMenuItem").GetComponent<EvidenceMenuItem>();
+            yield return _inputTestTools.SetMousePosition(firstMenuItem.transform.position);
+            Menu menu = _evidenceMenu.GetComponent<Menu>();
+            Assert.AreEqual(firstMenuItem.CourtRecordObject.DisplayName, menu.SelectedButton.GetComponent<EvidenceMenuItem>().CourtRecordObject.DisplayName);
+            
+            // Loop through the menu items and check if they highlight correctly
+            for (int i = 1; i < menuItems.Length; i++)
+            {
+                MenuItem menuItem = menuItems.First(menuItem => menuItem.gameObject.name == $"EvidenceMenuItem ({i})");
```
ViMaSter: Nit: Having the outer variable (`MenuItem menuItem`) **and** parameter of the `First()`-predicate (.First(menuItem`) is a bit confusing to read for me and also prevents the variable from being used inside the scope. I'd suggest renaming either the inner or outer variable.

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuKeyboardTests.cs
```diff
@@ -0,0 +1,172 @@
...
+
+            // Spam navigation button
+            yield return _inputTestTools.PressForFrame(_inputTestTools.Keyboard.leftArrowKey);
+            yield return _inputTestTools.RepeatPressForFrames(_inputTestTools.Keyboard.enterKey, 50);
+
+            // Get the evidence menu text boxes to check if they are updated correctly
+            Menu menu = _evidenceMenu.GetComponent<Menu>();
+            TextMeshProUGUI[] evidenceTextBoxes = _evidenceMenu.GetComponentsInChildren<TextMeshProUGUI>();
+            TextMeshProUGUI evidenceName =
+                evidenceTextBoxes.First(evidenceTextBoxe => evidenceTextBoxe.gameObject.name == "EvidenceName");
```
ViMaSter: Nit: Typo `evidenceTextBoxe` -> `evidenceTextBox`.

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuKeyboardTests.cs
```diff
@@ -0,0 +1,172 @@
...
+        /// Closes profile menu at the end and checks something is selected.
+        /// </summary>
+        [UnityTest, Order(5)]
+        public IEnumerator ProfileMenuCanBeAccessed()
+        {
+            // Get the evidence menu text boxes to check if they are updated correctly
+            Menu menu = _evidenceMenu.GetComponent<Menu>();
+            TextMeshProUGUI[] evidenceTextBoxes = _evidenceMenu.GetComponentsInChildren<TextMeshProUGUI>();
+            TextMeshProUGUI evidenceName =
+                evidenceTextBoxes.First(evidenceTextBoxe => evidenceTextBoxe.gameObject.name == "EvidenceName");
```
ViMaSter: Nit: Typo `evidenceTextBoxe` -> `evidenceTextBox`.

Assets/Tests/PlayModeTests/Tools/InputTestTools.cs
```diff
@@ -0,0 +1,102 @@
+using System.Collections;
+using UnityEngine;
+using UnityEngine.InputSystem;
+using UnityEngine.InputSystem.Controls;
+using UnityEngine.UI;
```
ViMaSter: Nit: This namespace seems to be unused in this file. I'd suggest removing it.

Assets/Tests/PlayModeTests/Tools/InputTestTools.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        /// <summary>
+        /// Start this coroutine to press a specified key every repeatedly,
+        /// holding for a specified number of seconds each time the key is pressed.
+        /// </summary>
+        /// <param name="control">The key to press.</param>
+        /// <param name="seconds">The number of seconds to hold the key for each press.</param>
+        /// <param name="repeats">The number of times to press the key.</param>
+        /// <returns></returns>
+        public IEnumerator RepeatPressForSeconds(ButtonControl control, float seconds, int repeats)
```
ViMaSter: Nit: `RepeatPressForSeconds()` seems to be unused. I'd personally suggest either removing it (if we end up needing it, we can add it later) or merging it with `PressForSeconds()` by adding `repeats` as an optional parameter (`public IEnumerator PressForSeconds(ButtonControl control, float seconds, int repeats = 1)`).  
I'm always a big fan of removing unused code, that way there's less of it to keep in mind when working with it!

Assets/Tests/PlayModeTests/Tools/InputTestTools.cs
```diff
@@ -0,0 +1,102 @@
...
+using UnityEngine.InputSystem.Controls;
+using UnityEngine.UI;
+
+namespace Tests.PlayModeTests.Tools
+{
+    /// <summary>
+    /// Contains useful methods used when testing features that use Unity Input System.
+    /// Also contains input devices that should be used to pass ButtonControls to the methods.
+    /// </summary>
+    public class InputTestTools : InputTestFixture
```
ViMaSter, line 12: Wrapping this functionality in a helper-class is a really solid addition! Always a big fan of the [rule of 3](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))! 👍

Assets/Tests/EditModeTests/ObjectDictionaryTests.cs
```diff
@@ -0,0 +1,173 @@
...
+        private List<GameObject> CreateGameObjectList(int count)
+        {
+            var gameObjectList = CreateObjectList<GameObject>(count);
+            for (int i = 0; i < gameObjectList.Count; i++)
+            {
+                gameObjectList[i].name = i.ToString();
+            }
+            return gameObjectList;
+        }
+    }
```
ViMaSter: Nit: These methods can be made `static` and return `IEnumerable<>` instead of `List<>`.

Assets/Tests/EditModeTests/ObjectDictionaryTests.cs
```diff
@@ -0,0 +1,173 @@
...
+        [Test]
+        public void GetObjectFromDictionaryAtIndex()
+        {
+            int objectDictionaryCount = 1000;
+            var objectDictionary = new ObjectDictionary<GameObject>(CreateGameObjectList(objectDictionaryCount), CreateGameObjectList(objectDictionaryCount));
+            for (int i = 0; i < objectDictionaryCount; i++)
+            {
+                Assert.AreEqual(i.ToString(), objectDictionary[i].name);
+            }
+        }
```
ViMaSter: Nit: The `*Count` variables and `numberOfObjectsToRemove` used during setup of each of these cases can be made `const`, to prevent them from being accidentally changed during runtime.

Assets/Scripts/TextDecoder/DialogueController.cs
```diff
@@ -166,30 +166,12 @@ public void SetMenuOpen(bool isMenuOpen)
...
-        HandlePresenting(actor.name);
-    }
-
     /// <summary>
     /// Actually handles the evidence being presented by seeing if it is one of the current choices and then progressing the story appriopriately. Makes sure we're in cross examination mode before continuing.
     /// May spawn a random failure sub story.
     /// </summary>
     /// <param name="presentedObject">Name of the object you want to present to the court</param>
-    private void HandlePresenting(string presentedObject)
+    public void HandlePresenting(ICourtRecordObject presentedObject)
```
ViMaSter, line 174: Really intuitive way of simplifying the public interface of this class while making it even less error-prone! 👍

Assets/Scripts/ObjectDictionary/ObjectDictionaryInterface.cs
```diff
@@ -0,0 +1,65 @@
+using UnityEngine;
+using Object = UnityEngine.Object;
+
+/// <summary>
+/// This class allows interaction between unity objects and ObjectDictionary objects.
+/// Allows assigning of an object with an IObjectList interface to convert to a dictionary.
+/// </summary>
+/// <typeparam name="T">The type of object to store in the dictionary.</typeparam>
```
ViMaSter: Nit: I personally find specifying `The type of values in the dictionary.` to make it easier to read, as only the value type is being set.

Assets/Scripts/ObjectDictionary/ObjectDictionaryInterface.cs
```diff
@@ -0,0 +1,65 @@
...
+/// <typeparam name="S">The type that the assignable IObjectList should use.</typeparam>
+public abstract class ObjectDictionaryInterface<T, S> : MonoBehaviour where T : Object where S : IObjectList<T>
+{
+    [SerializeReference, Tooltip("Drag an Actor Dictionary here containing all the actors available in the scene.")]
+    private S _masterObjectList;
+    
+    [field: SerializeField, Tooltip("Add any evidence that the player should have at the start of the scene here.")]
+    protected T[] CurrentObjectList { get; private set; }
```
ViMaSter: Nit: The tooltips speak of `actors` and `evidence` explicitly, while the fields are technically being abstract. I'd change this to `ICourtRecordObject` or `Object` instead.

Assets/Scripts/ObjectDictionary/ObjectDictionaryInterface.cs
```diff
@@ -0,0 +1,65 @@
...
+/// <summary>
+/// This class allows interaction between unity objects and ObjectDictionary objects.
+/// Allows assigning of an object with an IObjectList interface to convert to a dictionary.
+/// </summary>
+/// <typeparam name="T">The type of object to store in the dictionary.</typeparam>
+/// <typeparam name="S">The type that the assignable IObjectList should use.</typeparam>
+public abstract class ObjectDictionaryInterface<T, S> : MonoBehaviour where T : Object where S : IObjectList<T>
+{
+    [SerializeReference, Tooltip("Drag an Actor Dictionary here containing all the actors available in the scene.")]
+    private S _masterObjectList;
```
ViMaSter: Nit: The term `master` is immediately associated with a `master/slave`-relationship (think SATA, IDE, databases), which doesn't really apply here. Similarly to the tooltip-text, I'd prefer `availableObjectList`, as it contains a list of all potential objects that can be added to the `CurrentObjectList`.

Assets/Scripts/ObjectDictionary/ObjectDictionaryInterface.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        ObjectsDictionary.RemoveObject(objectName);
+    }
+
+    /// <summary>
+    /// Call this method to get an object at a specific index of the dictionary.
+    /// </summary>
+    /// <param name="index">The index of the object to get.</param>
+    /// <returns>The object at the specified index.</returns>
+    public T GetObjectAtIndex(int index)
```
ViMaSter: Nit: This method is never used. `GetObjectAtIndex()` is only ever used on `ICourtRecordObjectDictionary`-objects, which - as it's an interface - require an explicit implementation and - as it exists as a shallower type - can't fall back to this base implementation. I'd personally suggest removing it.

Assets/Scripts/ObjectDictionary/IObjectList.cs
```diff
@@ -0,0 +1,6 @@
+using UnityEngine;
+
+public interface IObjectList<T> where T : Object
```
ViMaSter: Nit: I'd suggest using `<out T>` instead of `<T>` here, which would make this [a covariant](https://docs.microsoft.com/en-us/dotnet/standard/generics/covariance-and-contravariance#generic-interfaces-with-covariant-type-parameters), allowing it to be used more broadly.

Assets/Scripts/ObjectDictionary/ICourtRecordObject.cs
```diff
@@ -0,0 +1,10 @@
+using UnityEngine;
+
+public interface ICourtRecordObject
+{ 
+    public string name { get; }
```
ViMaSter: Nit: This might be something Unity-specific related to `name` I'm unaware of, but all other properties in this interface start with an uppercase letter, so I'd personally prefer aligning them to all use UpperCamelCase.

Assets/Scripts/Evidence/EvidenceMenu.cs
```diff
@@ -179,4 +220,29 @@ private bool HasEvidenceDictionary()
...
+        {
+            _profileMenuActive = true;
+            _activeDictionary = _actorDictionary;
+            _menuLabel.sprite = _profilesMenuLabel;
+        }
+        
+        UpdateEvidenceMenu();
+        _menu.SelectInitialButton();
+    }
 }
```
ViMaSter, line 250: Nit: Strictly not part of this PR, but the `System` namespace in this file seems to be unused. I'd suggest removing it.

Assets/Scripts/Actor/ActorData.cs
ViMaSter: Agreed! `AgeCertainty.Unknown` is a valid state and should be handled appropriately. The `default` should be treated as an exception or error.

---

I'd personally also suggest converting it to a [switch expression](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression) like so:
```cs
string age = _ageCertainty switch {
    AgeCertainty.Known => Age.ToString(),
    AgeCertainty.Guess => $"Maybe {Age}?",
    AgeCertainty.Unknown => "???",
    _ => throw new ArgumentOutOfRangeException()
};
```

Assets/Scripts/ObjectDictionary/ObjectDictionary.cs
```diff
@@ -0,0 +1,185 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Converts a list of objects into a dictionary of objects
+    /// so they can be accessed by name
+    /// </summary>
+    /// <param name="array">The list to convert to a dictionary.</param>
+    /// <returns>The dictionary that the list has been converted to.</returns>
+    private Dictionary<string, T> ArrayToDictionary(IEnumerable<T> array)
```
ViMaSter: Nit: I'd recommend making this method `static` as it doesn't appear to be using any instance members.

ViMaSter (changes requested):
Really solid implementation! Always great to see a bunch of tests being added as well! <3

My only required changes (anything that's not prefixed with `Nit:` or got a 👍) would be:
- the typos ([1](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/52/files#r706865549), [2](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/52/files#r706865542)) and
- [the test case that's commented out](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/52/files#r706865061)

---

A sidenote:

I noticed, that the class `ObjectDictionary` does not implement `IDictionary` and the `ObjectDictionaryInterface` is actually an `abstract class` instead of an `interface`.  
This made this change a bit difficult to read for me initially.  

Looking through the source it definitely is possible to get an understanding of what these classes actually do, but as the terms `interface` and `dictionary` are already being used for definitions inside the language-standard, my mind immediately jumped to a set of behaviors and available methods that aren't actually present inside these classes when initially looking at these classes, which might cause confusion for people working on it the first time or after a long period of time has passed.

If I can offer two suggestions:
1. From what I can see, `ObjectDictionary` is only ever used inside `ObjectDictionaryInterface` and it's test suite. I'd recommend the following approach:
   * Integrate the logic of `ObjectDictionary` into `ObjectDictionaryInterface` making it one class
   * Rename this new `ObjectDictionaryInterface` class to `ObjectDictionary`
   This will result in less nesting of classes, while also reducing potential confusion caused by the word `Interface`
2. Instead of creating and implementing the custom interface `IObjectList`, implement `IDictionary<>` directly
   This will result in associations with the word `Dictionary` actually being fulfilled
   _Note: The same also applies to `ICourtRecordObjectDictionary`._

I'd be happy to hear your thoughts on that; as it's a bigger topic I'm also open to have a quick voice chat about it! However, as the functionallity operates just fine, this is just a suggestion and not blocking approval!

---

I really appreciate you adhering to the boyscout principle of "leaving the campsite cleaner than you found it" and taking the time to create the `InputTestTools` class! This should keep flakey tests to an absolute minimum!

Assets/Scripts/Actor/ActorData.cs
PhoebeMitchell: Changed to use ViMaSter's suggestion.

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuMouseTests.cs
PhoebeMitchell: Good catch! Commented out that method for debugging purposes and must have missed the attributes when uncommenting. Fixed.

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuMouseTests.cs
PhoebeMitchell: Done

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuMouseTests.cs
PhoebeMitchell: Changed First() parameter to "item".

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuKeyboardTests.cs
PhoebeMitchell: Fixed

Assets/Tests/PlayModeTests/Tools/InputTestTools.cs
PhoebeMitchell: Done

Assets/Scripts/ObjectDictionary/ObjectDictionaryInterface.cs
PhoebeMitchell: Changed

Assets/Scripts/ObjectDictionary/ObjectDictionaryInterface.cs
PhoebeMitchell: Changed to _availableObjectList

Assets/Scripts/ObjectDictionary/IObjectList.cs
PhoebeMitchell: Done

Assets/Scripts/Evidence/EvidenceMenu.cs
PhoebeMitchell: Done

Assets/Scripts/ObjectDictionary/ObjectDictionary.cs
PhoebeMitchell: Done

Assets/Scripts/ObjectDictionary/ICourtRecordObject.cs
PhoebeMitchell: This is to allow the name property of Object to be accessed. Would you suggest changing it to Name and then doing `string Name => name;` on classes that implement this instead?

Assets/Scripts/ObjectDictionary/ObjectDictionaryInterface.cs
PhoebeMitchell: Remnants of old implementations. Updated.

Assets/Tests/EditModeTests/ObjectDictionaryTests.cs
PhoebeMitchell: Done

Assets/Tests/PlayModeTests/Tools/InputTestTools.cs
PhoebeMitchell: Removed "Repeat" methods. Modified PressForFrame and PressForSeconds to allow repeats.

Assets/Tests/EditModeTests/ObjectDictionaryTests.cs
PhoebeMitchell: Done

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuKeyboardTests.cs
PhoebeMitchell: Fixed

Assets/Scripts/ObjectDictionary/ICourtRecordObject.cs
ViMaSter: That actually sounds pretty good!  
The [official docs of `UnityEngine.Object.name`](https://docs.unity3d.com/ScriptReference/Object.html) state that `Components share the same name with the game object and all attached components`. Given that, I'd just go a step further and change it to `DisplayName`. That way this property is distinctly tied to our game and - if we choose to do so - we can use the value from the engine by using `=> name` on any class that derived from `UnityEngine.Object`.

PhoebeMitchell: Changed to InstanceName

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuKeyboardTests.cs
ViMaSter: Just FYI: The latest version of this file still contains the typo. 👀
https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/0bea8fa254d3b558d48947255e17a399e51e4f56/Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuKeyboardTests.cs#L131

Assets/Tests/EditModeTests/ObjectDictionaryTests.cs
```diff
@@ -0,0 +1,102 @@
...
+using NUnit.Framework;
+using UnityEngine;
+
+namespace Tests.EditModeTests
+{
+    public class ObjectDictionaryTests
+    {
+        private EvidenceDictionary _evidenceDictionary;
+
+        private string[] _evidenceNames =
```
ViMaSter: Nit: This field can be made `readonly`.

Assets/Tests/EditModeTests/ObjectDictionaryTests.cs
```diff
@@ -0,0 +1,102 @@
+using System.Collections.Generic;
+using System.Linq;
```
ViMaSter: Nit: No symbols from these namespaces are used in the this file, so the first two lines of this file can safely be removed.

Assets/Scripts/ObjectDictionary/ObjectDictionary.cs
```diff
@@ -0,0 +1,197 @@
...
+    }
+    
+    /// <summary>
+    /// Method used by GetObject, AddObject, and RemoveObject methods to check
+    /// if an object is in the dictionary before attempting to access it.
+    /// </summary>
+    /// <param name="objectName">The name of the object to look for in the dictionary.</param>
+    /// <param name="dictionaryToCheck">The dictionary to look for the object in.</param>
+    /// <returns>True if the object is in the dictionary, and false if it is not.</returns>
+    private bool IsObjectInDictionary(string objectName, Dictionary<string, T> dictionaryToCheck)
```
ViMaSter: Nit: This method can be made static.

Assets/Scripts/Evidence/EvidenceDictionary.cs
```diff
@@ -1,157 +1,24 @@
-using System;
-using System.Collections.Generic;
-using System.Linq;
 using UnityEngine;
```
ViMaSter: Nit: No symbols from this namespace are used in the this file, so this `using`-directive can safely be removed too.

Assets/Scripts/Actor/ActorList.cs
```diff
@@ -0,0 +1,10 @@
+using System;
+using System.Collections.Generic;
```
ViMaSter: Nit: No symbols from this namespace are used in the this file, so this `using`-directive can safely be removed.

Assets/Scripts/Actor/ActorData.cs
```diff
@@ -1,9 +1,16 @@
+using System;
 using System.Collections.Generic;
```
ViMaSter: Nit: No symbols from this namespace are used in the this file, so this using-directive can safely be removed.

JuiceGrape (dismissed):

Assets/Scripts/ObjectDictionary/ObjectDictionaryInterface.cs
PhoebeMitchell: Removed.

Assets/Scripts/Actor/ActorController.cs
```diff
@@ -3,14 +3,16 @@
 using System.Collections.Generic;
```
ViMaSter, line 3: Nit: No symbols are used from `using System;` and `using System.Collections;`, so the first two lines of this file can safely be removed.

Assets/Scripts/Actor/ActorData.cs
```diff
@@ -13,6 +20,12 @@ public class ActorData : ScriptableObject
     [field: SerializeField, Tooltip("Name displayed when referring to this actor or when they are speaking.")]
```
ViMaSter, line 16: Nit: The property `DefaultSprite` isn't used (anymore?), could this be removed until functionality that uses it exists or would this result in a lot of Prefabs having to be updated?

Assets/Scripts/Evidence/Evidence.cs
```diff
@@ -3,11 +3,11 @@
 using UnityEngine;
```
ViMaSter, line 1: Nit: No symbols are used from `using System;` and `using System.Collections;`, so the first two lines of this file can safely be removed.

Assets/Scripts/Evidence/EvidenceController.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
```
ViMaSter, line 2: Nit: No symbols are used from `using System;` and `using System.Collections;`, so the first two lines of this file can safely be removed.

Assets/Scripts/ObjectDictionary/ObjectInventory.cs
```diff
@@ -0,0 +1,68 @@
+using System.Collections;
```
ViMaSter: Nit: No symbols from this namespace are used in the this file, so this using-directive can safely be removed.

ViMaSter (dismissed):
Looks good to me! Just a few nits, but nothing that's worth blocking in my opinion! Oh and thanks for integrating the typo fix for the build artifacts too.

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuKeyboardTests.cs
PhoebeMitchell: Fixed. Must have come back when reverting back to humble object pattern.

Assets/Tests/EditModeTests/ObjectDictionaryTests.cs
PhoebeMitchell: File no longer exists.

Assets/Tests/EditModeTests/ObjectDictionaryTests.cs
PhoebeMitchell: File no longer exists.

Assets/Scripts/ObjectDictionary/ObjectDictionary.cs
PhoebeMitchell: File no longer exists. The version of this method in latest version is public. I'm told static methods are bad in OOP. Thoughts?

Assets/Scripts/Evidence/EvidenceDictionary.cs
PhoebeMitchell: File no longer exists.

Assets/Scripts/Actor/ActorList.cs
PhoebeMitchell: Outdated.

Assets/Scripts/Actor/ActorData.cs
PhoebeMitchell: Removed.

Assets/Scripts/Actor/ActorController.cs
PhoebeMitchell: Will leave them to avoid changing too much of a file that isn't explicitly introduced in this PR.

Assets/Scripts/Actor/ActorData.cs
PhoebeMitchell: Removed as its intended functionality did not work it is and no longer used.

Assets/Scripts/Evidence/Evidence.cs
PhoebeMitchell: Removed

Assets/Scripts/Evidence/EvidenceController.cs
PhoebeMitchell: Will leave them to avoid changing too much of a file that isn't explicitly introduced in this PR.

Assets/Scripts/ObjectDictionary/ObjectInventory.cs
PhoebeMitchell: Removed

Assets/Scripts/Actor/ActorData.cs
```diff
@@ -21,4 +30,21 @@ public class ActorData : ScriptableObject
 
     [field: SerializeField, Tooltip("The animator controllers that this actor uses.")]
     public RuntimeAnimatorController AnimatorController { get; private set; }
+
+    public string InstanceName => name;
+    public string CourtRecordName => GenerateNameWithAge();
+    public Sprite Icon => Profile;
+    public string Description => Bio;
+
+    private string GenerateNameWithAge()
```
lyndsiWilliams, line 44: This method is missing a summary ✨

lyndsiWilliams (dismissed):
Found just one thing! It looks like you've got some failing tests as well, but otherwise this looks great!

Assets/Scripts/ObjectDictionary/ObjectDictionary.cs
JuiceGrape: Static methods are fine in OOP if you have a reason for making them. In this case, making this method static is a good idea if we want to use it in other classes as well, to avoid duplicate code. If we don't, leave it as is

Assets/Scripts/Actor/ActorData.cs
PhoebeMitchell: Added

Assets/Scripts/ObjectDictionary/ObjectDictionary.cs
PhoebeMitchell: Made IsObjectInDictionary and EnumerableToDictionary methods static as their output is always the same if the input doesn't change.

JuiceGrape (dismissed):

JuiceGrape (approved):
Solved bug looks good, ready to merge!

lyndsiWilliams (approved):
Looks great! ✨

PR #78, opened 2021-09-23
Feature 56 Optional Target Actor for SetPose and SetEmotion

### LINKED ISSUE #56: Add target actor value to SET_POSE and PLAY_EMOTION
## Description

The actions SET_POSE and PLAY_EMOTION should be able to take an actor value as well, as an optional second parameter. When this is added, the action should set the emotion or pose of every actor in the unity scene corresponding to the provided actor

## Acceptance Criteria

This story will be finished when the functionality is added to the standard inclusion prefab and is testable in a unity scene

notexplosive:
## Summary
You can now call `&SET_POSE:<pose_name>,<target_actor>` to set the pose of a target actor. By default you'll generally want `&SET_POSE:<pose_name>` because that uses the current actor which is what you want 99% of the time. We support the same functionality for SetEmotion.

To do this I created a lookup table on ActorController that corresponds `ActorData`s to `Actor`s. This assumes that this relationship is 1 ActorData to 1 Actor, which it technically might not be but Bas said we will probably keep them 1:1.

The unfortunate way this works is we have to store a "pending animation" on the target actor because the animator does not work while the actor is disabled (ie: the actor's scene is not visible)

(old gif is in accurate removed so it's not misleading)

## Testing instructions (Updated!!)
Load `OffScreenPose - TestScene`
This entire scene just shows Ross the whole time, but if you switch to the scene view you can see things happen o f f   s c r e e n
When the dialogue says Arin is "big now" switch to the Scene view and pan over to see Arin is big.
Similarly when the dialogue says tutorial boy is angry, switch to scene view to see his expression is now angry.

## Additional information

- `[ ]` Changes UI
closes #56


Assets/Scripts/Actor/Actor.cs
```diff
@@ -90,4 +102,9 @@ public bool MatchesActorData(ActorData actor)
     {
         return _actorData == actor;
     }
+
+    public void SetPendingAnimation(string animationName)
```
lyndsiWilliams: This method is missing a summary ✨

lyndsiWilliams (dismissed):
Just one missing summary, looks great otherwise! ✨

Assets/Scripts/Actor/Actor.cs
```diff
@@ -51,7 +63,7 @@ public void PlayAnimation(string animation)
 
         _animator.Play(animationHash);
     }
-    
+
     /// <summary>
     /// This method is called by animations when they are completed and require the next line to be to be read.
     /// </summary>
```
ViMaSter: **Nit:** `OnAnimationComplete()` appears to be never called directly or via any UnityEvent. (At least I couldn't find any. If there is, feel free to point it out and disregard this. 👀)
The only thing with bindings somewhat related to it, is the `_onAnimationComplete`-event of `ActorController`:
image


If it is no longer used, I'd suggest removing it.  
If it's still required, it should be re-implemented at the appropriate site.

Assets/Scripts/Actor/Actor.cs
```diff
@@ -8,6 +7,7 @@ public class Actor : MonoBehaviour
     private Animator _animator;
```
ViMaSter: **Nit:** `_spriteRenderer` appears to not be used (it's only set up inside `Awake()`). If this is the case, I'd suggest removing it.

Assets/Scripts/Actor/Actor.cs
```diff
@@ -36,7 +48,7 @@ public void SetActor(ActorData actorData)
     public void PlayAnimation(string animation)
```
ViMaSter: Mega **nit:** This is a **huge** edge-case, so feel free to disregard this, but I still wanna point it out:  

Unity previously allowed access to commonly used components via fields like `.transform`, `.rigidbody` and also `.animation`. Usage is discouraged and `GetComponent<Transform>()`/`GetComponent<Rigidbody>()`/`GetComponent<Animation>()`, but the field is still available.

With this parameter being named `animation`, it hides the `animation`-field of the deriving class `MonoBehaviour`. I'd suggest renaming it to `animationName`; to resolve this override and also make it more verbose (as we're technically passing the name of an animation not an instance of `Animation` directly at this point).

Assets/Scripts/Actor/Actor.cs
```diff
@@ -80,7 +92,7 @@ public void SetTalking(bool isTalking)
     {
         _animator.SetBool("Talking", isTalking);
```
ViMaSter: **Nit:** String-based property lookups on animations can get expensive.
Adding
```cs
private static readonly int Talking = Animator.StringToHash("Talking");
```
and replacing this line with 
```cs
_animator.SetBool(Talking, isTalking);
```
would make this a one-time operation at startup, instead of looking up the property each time `SetTalking()` gets called.

Assets/Scripts/Actor/ActorController.cs
```diff
@@ -22,6 +20,7 @@ public class ActorController : MonoBehaviour, IActorController
 
     public bool Animating { get; set; }
 
+    private Dictionary<ActorData, Actor> _actorDataToActor = new Dictionary<ActorData, Actor>();
```
ViMaSter: **Nit:** This field can be made `readonly`. (This only prevents this field from being re-assigned, calls to methods like `.Add()` or `.Remove()` are still possible.)

Assets/Scripts/Actor/ActorController.cs
```diff
@@ -59,49 +60,107 @@ public void SetActiveActorObject(Actor actor)
...
+    private Actor FindActor(string actorName)
+    {
+        var actorData = FindActorData(actorName);
+        if (actorData != null && _actorDataToActor.ContainsKey(actorData))
+        {
+            return _actorDataToActor[actorData];
+        }
+
+        return null;
+    }
```
ViMaSter: **Nit:** While we do utilize comments, I'm personally a big fan of having verbose method names. This allows for methods to be legible, even when seeing them inside another context (i.e. when they're called or referenced in another file). I'd suggest the following renamings:
* `FindActorData` ->`FindActorDataInInventory`
* `FindActor` -> `FindActorInScene`

Assets/Scripts/Actor/ActorController.cs
```diff
@@ -59,49 +60,107 @@ public void SetActiveActorObject(Actor actor)
...
     /// <summary>
     /// Sets the pose of the active actor.
     /// In working this is mostly the same as PlayEmotion without calling OnAnimationStarted so the system can continue without waiting for the animation to end.
     /// </summary>
     /// <param name="pose"></param>
-    public void SetPose(string pose)
+    public void SetPose(string pose, string actorName = null)
     {
-        if (_activeActor == null)
+        if (actorName == null || FindActor(actorName) == _activeActor)
```
ViMaSter: **Nit:** I personally prefer using `string.IsNullOrEmpty` or even `string.IsNullOrWhiteSpace` instead of direct `null` comparison as treating "an empty string or string consisting entirely of whitespace" differently to `null` usually leads to confusion and unexpected behavior.

Assets/Scripts/Actor/ActorController.cs
```diff
@@ -59,49 +60,107 @@ public void SetActiveActorObject(Actor actor)
...
+            }
+            _activeActor.PlayAnimation(pose);
+        }
+        else
+        {
+            var actor = FindActor(actorName);
+
+            if (actor == null)
+            {
+                Debug.LogError(string.Format("Actor not found: {0}", actorName));
```
ViMaSter: **Nit:** I personally prefer string interpolation syntax (`$"Actor not found: {actorName}"`) over `string.Format()`.

Assets/Scripts/Actor/ActorController.cs
```diff
@@ -166,7 +225,7 @@ public void OnAnimationDone()
             Animating = false;
             _onAnimationComplete.Invoke();
         }
-        
+
     }
 
     /// <summary>
```
ViMaSter, line 250: **Nit:** Typo: `appropiately` -> `appropriately`. (Housekeeping item; not part of this PR's changes.)

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -144,26 +141,55 @@ private void SetSpeaker(string actor, SpeakingType speakingType)
...
+        }
+        else if (parameters.Length == 2)
+        {
+            _actorController.SetPose(parameters[0], parameters[1]);
+            _onActionDone.Invoke();
+        }
+        else
+        {
+            Debug.LogError("Invalid number of paramaters for function PLAY_EMOTION");
+        }
```
ViMaSter: **Nit 1:** Not really a nit, just a good point to show off new language features: This could be simplified using [deconstruction](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/deconstruct) and [switch expressions](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression):
```cs
string[] parameters = poseAndActorName.Split(ACTION_PARAMETER_SEPARATOR);

var (pose, actorName) = parameters.Length switch {
    1 => (parameters[0], null),
    2 => (parameters[0], parameters[1]),
    _ => throw new NotSupportedException($"'${parameters.Length}' is an invalid number of parameters for function PLAY_EMOTION")
};
_actorController.SetPose(pose, actorName);
_onActionDone.Invoke();
```

---

**Nit 2:** Typo: `Invalid number of paramaters` -> `Invalid number of parameters`

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -144,26 +141,55 @@ private void SetSpeaker(string actor, SpeakingType speakingType)
...
+            _actorController.PlayEmotion(parameters[0]);
+        }
+        else if (parameters.Length == 2)
+        {
+            _actorController.PlayEmotion(parameters[0], parameters[1]);
+        }
+        else
+        {
+            Debug.LogError("Invalid number of paramaters for function PLAY_EMOTION");
+        }
```
ViMaSter: **Nit:** [See comment for Line 165](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/78/files#diff-1fad38fd8a052f9a7753f963a1c6d48c6a6a590f3421e68668dbb765a02ace8aR165).

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -1,9 +1,6 @@
-using System.Collections;
-using System.Collections.Generic;
-using System;
 using System.Globalization;
-using UnityEngine.Events;
 using UnityEngine;
```
ViMaSter, line 2: Removing unnecessary `using`-directives! Great catch! 👍

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -110,7 +107,7 @@ private void SetActorVisibility(string showActor)
         bool shouldShow;
         if (bool.TryParse(showActor, out shouldShow))
```
ViMaSter, line 113: This `out` variable declaration can be inlined:
```cs
if (bool.TryParse(showActor, out bool shouldShow))
```

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -144,26 +141,55 @@ private void SetSpeaker(string actor, SpeakingType speakingType)
     /// <summary>
     /// Set the pose of the current actor
     /// </summary>
-    /// <param name="pose">Pose to display for the current actor</param>
-    private void SetPose(string pose)
+    /// <param name="poseAndActorName">Pose to display for the current actor</param>
```
ViMaSter: I'd update this parameter comment (and the one for `animationAndActorName` of `PlayEmotion()`) to reflect the ability to also override the actor and thereby queue the pose or animation.
Proposal:
### poseAndActorName
`[pose name]` to set pose for current actor OR `[pose name]$ACTION_PARAMETER_SEPARATOR[actor name]` to set pose for another actor
### animationAndActorName
`[animation name]` to set pose for current actor OR `[animation name]$ACTION_PARAMETER_SEPARATOR[actor name]` to queue animation for another actor (gets played as soon as actor is visible)

ViMaSter (dismissed):
Simple and clean change! Test-scene works, added a few nits, but absolutely nothing worth blocking. Solid work!

Assets/Scripts/Actor/Actor.cs
JuiceGrape: It is called in animations

Assets/Scripts/Actor/Actor.cs
JuiceGrape: This PR didn't touch this. Please keep comments to things this pr did touch

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
JuiceGrape: Not part of this pr

Assets/Scripts/Actor/ActorController.cs
JuiceGrape: Not part of this pr

Assets/Scripts/Actor/Actor.cs
JuiceGrape: It needs one for the animator to work

Assets/Scripts/Actor/Actor.cs
JuiceGrape: Not part of th is pr, although I do agree. If you're bothered by it, you can change it yourself.

Assets/Scripts/Actor/Actor.cs
JuiceGrape: Also this method isn't necessary with the fix I proposed in the programmer chat

JuiceGrape: Speaking of: Please include the fix

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
notexplosive: Formatter did this for me ;)

Assets/Scripts/Actor/Actor.cs
notexplosive: Future Wyatt is on it!

Hey Future Wyatt, the fix is to use: `_animator.keepAnimatorControllerStateOnDisable = true;` in Awake.

TooManyMoths: Isn't the problem that you cant Play animations on disabled animators though, rather than animators not keeping their state on disable?

notexplosive: Dammit Past Wyatt! That fix didn't work. But apparently once the new courtroom changes are in, it should.

JuiceGrape: It won't future wyatt! But, as we discussed, you get to choose exactly how it will work and we'll put that in the wiki.
Also update the wiki

JuiceGrape: Hey future wyatt, I am here to solve your woes: Just forego this functionality and make it an error state.

lyndsiWilliams: This was an interesting read 😹 
So is this being removed or kept?

ViMaSter (dismissed):

Assets/Scripts/Actor/ActorController.cs
```diff
@@ -59,49 +60,107 @@ public void SetActiveActorObject(Actor actor)
...
+        }
+        else
+        {
+            var actor = FindActorInInventory(actorName);
+
+            if (actor == null)
+            {
+                Debug.LogError($"Actor not found: {actorName}");
+            }
+            actor.SetPendingAnimation(pose);
```
PhoebeMitchell: Should there be a return in the if statement to avoid a null reference exception?

Assets/Scripts/Actor/Actor.cs
PhoebeMitchell: I think we're waiting for #80 to be merged? As this should set poses of actors within one BGScene, thus avoiding having to set poses on inactive game objects.

Assets/Scripts/Actor/ActorController.cs
notexplosive: Good catch, thanks.

Assets/Scripts/Actor/ActorController.cs
```diff
@@ -73,10 +73,20 @@ public void SetActiveActor(string actor)
...
+            AddActorToLookupTable(targetActorData, _activeActor);
         }
     }
 
+    /// <summary>
+    /// Adds actor to the lookup table so we can go from ActorData -> Actor
+    /// </summary>
+    /// <param name="actorData">ActorData that corresponds to the actor</param>
+    /// <param name="actor">Actor that corresponds to the actorData</param>
+    private void AddActorToLookupTable(ActorData actorData, Actor actor)
```
ViMaSter: Minor **nit**: [`Add`ing to a dictionary in C# usually means that an exception is thrown, if the same key is already present.](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.add?view=net-5.0#:~:text=//%20The%20Add%20method%20throws%20an%20exception%20if%20the%20new%20key%20is%0A//%20already%20in%20the%20dictionary.%0Atry%0A%7B%0A%20%20%20%20openWith.Add(%22txt%22%2C%20%22winword.exe%22)%3B%0A%7D%0Acatch%20(ArgumentException)%0A%7B%0A%20%20%20%20Console.WriteLine(%22An%20element%20with%20Key%20%3D%20%5C%22txt%5C%22%20already%20exists.%22)%3B%0A%7D)  
However, this code `set`s an entry instead; either creating a new one or overwriting an existing one. If we want to keep it, I think renaming this to `SetActorInsideLookupTable` and changing the names of parameter `actorData` to `key` and `actor` to `value` makes this code easier to read.

That said, I personally don't think this method is worth extracting to begin with, unless it prevent additional code (like checks or tracing) from being duplicated too.

ViMaSter (dismissed):
LGTM! Less code for identical behavior is always good!

Assets/Scripts/Actor/ActorController.cs
notexplosive: I'll take your renames but I'm keeping the extracted method. I think the method is worth extracting because I think the concept is meaningful enough to have a name/docstring to describe the behavior.

lyndsiWilliams (approved):
Code looks great! I pulled down the branch locally and the test scene works just as described. Nice work!

ViMaSter (approved):
Looked over the few changes since my last review; looks good to me once more! And sorry for the delay, especially for such small of a change.

Explicitly want to say thanks for taking the time to [elaborate on your choice on the `AddActorToLookupTable`-method discussion](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/78#discussion_r723358156)! I can definitely see the logic behind the wrapper to make things more legible; always a big fan of writing code for humans first and the compiler second!

Issue #68, opened 2021-09-17
[BUG] Able to present evidence outside of presenting moments


JuiceGrape:
## Describe the bug

When trying to present evidence, you can do it during any moment in a cross examination, not just during the part where the witness is reciting their testimony.

## Steps To Reproduce

- Open inky test scene
- Play cross examination
- Press the witness
- During the press witness dialogue, press z to open the evidence menu
- Press enter to present the attorney's badge

## Expected behavior

Presenting evidence should be fully disabled when it isn't applicable, so it should only be enabled during cross examination when the witness is reciting their testimony. Every other time, it should simply do nothing (and keep the evidence menu open)

JuiceGrape:
You should also be able to present if you were prompted using the PRESENT_EVIDENCE command

ViMaSter:
The fix we've discussed works and it's covered [by a test](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/c256e7a5416c0004fbf05032a1e65d7cf5f00361/Assets/Tests/PlayModeTests/Scenes/CrossExamination/ViaKeyboard.cs#L60-L103) on the branch (`bugfix/68-present-evidence-outside-of-presenting-moments`) as well.

However, regarding preventing the evidence menu from closing entirely:  
`UnityAction` doesn't seem to support `out` parameters. The only other option would be [System.Func<>](https://docs.microsoft.com/en-us/dotnet/api/system.func-2?view=net-5.0) which would support a return value, but has no support for exposing it via the Unity Editor Inspector-tab.  
That said, even were we to use a `System.Func<>`, we still need support for the ability to hide and show the "[ENTER] -> Present"-part of the controls inside the EvidenceMenu once we add that part of the UI. With a `Systen.Func<>`, we'd only have the ability to see if some event has handled presenting evidence **on enter press**, but no way to check if pressing enter/presenting evidence is an option to begin with.

Instead of only checking for `!_isAtChoice` inside the DialogueController when attempting to present, I'd argue that preventing that event from ever being sent is even better.
I would propose...
- **Adding a `SetCanPresentEvidence(bool canPresent)`-method to `EvidenceMenu`**  
  This would work similar to `SetNavigationButtonsActive()`, except that the value of `canPresent` will determine whether or not presenting evidence is an option right now, instead of `EvidenceMenu` determining it on its own
[...]

PR #79, opened 2021-09-23
feature/68 present evidence outside of presenting moments

ViMaSter:
## Summary

Resolves the bug outlined inside https://github.com/Studio-Lovelies/GG-JointJustice-Unity/issues/68 (presenting evidence outside of presenting moments) and introduces automated tests for it, so this bug is automatically caught in the future.  
Also contains general housekeeping for all files affected by this fix (typos, unused `using`s, etc.).

---

## Testing instructions

**Either...**
1. Follow the steps inside the related issue (https://github.com/Studio-Lovelies/GG-JointJustice-Unity/issues/68)

**...or...**

1. Open the Unity Editor
2. Run the following test cases:
    * `CrossExamination.ViaKeyboard.CanPresentEvidenceOnExamination`
    * `CrossExamination.ViaKeyboard.CantPresentEvidenceDuringExaminationDialogue`
3. Notice both tests succeeding

## Additional information

- `[ ]` Changes UI
  - https://github.com/Studio-Lovelies/GG-JointJustice-Unity/issues/68


lyndsiWilliams (approved):
Looks good to me!

Assets/Scripts/TextDecoder/DialogueController.cs
```diff
@@ -167,12 +164,17 @@ public void SetMenuOpen(bool isMenuOpen)
...
 
     /// <summary>
-    /// Actually handles the evidence being presented by seeing if it is one of the current choices and then progressing the story appriopriately. Makes sure we're in cross examination mode before continuing.
+    /// Actually handles the evidence being presented by seeing if it is one of the current choices and then progressing the story appropriately. Makes sure we're in cross examination mode before continuing.
     /// May spawn a random failure sub story.
     /// </summary>
     /// <param name="presentedObject">Name of the object you want to present to the court</param>
     public void HandlePresenting(ICourtRecordObject presentedObject)
     {
+        if (_subStory != null)
```
JuiceGrape, line 173: Not every sub-story is going to be a failure state, and sub-stories can have their own sub-stories as well, so this doesn't seem like the best way to handle this

JuiceGrape (changes requested):

Assets/Scripts/TextDecoder/DialogueController.cs
JuiceGrape: Maybe propegate this check upwards to the sub-story if there is one. It will realize it's not at a point where it can present evidence, and just ignore the request.

Assets/Scripts/TextDecoder/DialogueController.cs
```diff
@@ -167,12 +164,17 @@ public void SetMenuOpen(bool isMenuOpen)
...
+        {
+            return;
```
JuiceGrape, line 175: I'm pretty sure the evidence menu closes when doing this, which shouldn't be the case.

Assets/Scripts/TextDecoder/DialogueController.cs
JuiceGrape: Also also: this doesn't even fix the bug. I can still present evidence at all times, except during a failure state now, and that was never the issue. Simply add a check to see if we're at a point where presenting evidence is an option, and use that

JuiceGrape:
I'm actually going to close this, because it doesn't fix the bug presented in the bug report

PR #80, opened 2021-09-23
Feature/60 courthouse scene controls

### LINKED ISSUE #60: Courthouse scene controls
blocked by #29 and #56 

## Description

During courthouse scenes, the background scene will contain 3 active actors instead of 1. This is done because the background for the for the courthouse scene also contains some motion blur lines which portray quick movement.
To move around this 3 person background, 3 new actions must be added:
1. JUMP_TO_POSITION:int position - This jumps to a preset position set in the background scene
2. PAN_TO_POSITION:int position, float time - This pans to a preset position set in the background scene, taking time amount of seconds
3. SET_ACTOR_POSITION:int position, actor - This sets an actor to a preset position in the background scene.

## Dev guide

Probably best to pick up #58 first and re-use some of the code, or at least steal some of it

## Acceptance Criteria

This story will be finished when the aforementioned functionality is implemented and testable in a unity scene.

JuiceGrape:
## Summary
Added support for sub positions in bg-scenes

OBJECTIE

## Testing instructions
Do the inky-test scene thing like normal!

## Additional information

- `[ ]` Changes UI
closes #60


Assets/Scripts/BGScenes/BGScene.cs
```diff
@@ -4,15 +4,81 @@
...
+    /// Gets the actor object at a certain position without activating the system
+    /// </summary>
+    /// <param name="position">Target position, 1 based</param>
+    /// <returns>Actor object at target position</returns>
+    public Actor GetActorAtSubposition(int position)
+    {
+        int actualPosition = position - 1;
+        if (actualPosition >= _subPositions.Count)
+        {
+            Debug.LogError($"No such subposition in scene {gameObject.name}");
```
notexplosive: It would be helpful if this logged the entered `position` value as well, gives a slightly better hint to the writer where the error was.

Assets/Scripts/Interfaces/ISceneController.cs
```diff
@@ -15,6 +15,8 @@ public interface ISceneController
     void SetScene(string background);
     void SetCameraPos(Vector2Int position);
     void PanCamera(float seconds, Vector2Int finalPosition);
+    void PanToSubPosition(int position, float seconds);
+    public void SetToSubPosition(int position);
```
notexplosive: nit `public` is not needed on an interface

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -165,6 +168,29 @@ private void PlayEmotion(string animation)
...
+            return;
+
+        string[] parameters = positionAndActor.Split(ACTION_PARAMETER_SEPARATOR);
+
+        if (parameters.Length != 2)
+        {
+            Debug.LogError("Invalid amount of parameters for function SET_ACTOR_POSITION");
+            return;
+        }
+        if (int.TryParse(parameters[0], out int pos))
```
notexplosive: consider: Add error case for if I provide invalid (ie: non-integer) input.

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -372,6 +398,45 @@ void Wait(string seconds)
...
+    /// <summary>
+    /// Jump-cuts the camera to the target sub position if the bg-scene has sub positions.
+    /// </summary>
+    /// <param name="position">String containing an integer referring to the target sub position, 1 based.</param>
+    void JumpToSubPosition(string position)
+    {
+        if (!HasSceneController())
+            return;
+
+        if (int.TryParse(position, out int pos))
```
notexplosive: Same as other comment;  Consider adding an error if this parse fails.

Assets/Scripts/BGScenes/BGScene.cs
```diff
@@ -4,15 +4,81 @@
...
+    /// Sets the active sub position so the actor and position can be used.
+    /// </summary>
+    /// <param name="position">target position, 1 based.</param>
+    public void SetSubPosition(int position)
+    {
+        int actualPosition = position - 1;
+        
+        if (actualPosition >= _subPositions.Count)
+        {
+            Debug.LogError($"No such subposition in scene {gameObject.name}");
```
notexplosive: It would be helpful if this logged the `position` that was provided, gives a hint as to what went wrong.

notexplosive (dismissed):
LGTM, few thoughts about error logging but that's about it.

Assets/Scripts/Actor/ActorController.cs
```diff
@@ -177,4 +187,31 @@ public void SetSpeakingType(SpeakingType speakingType)
...
+    /// <summary>
+    /// Sets an actor to a sub position in the scene, if the active bg-scene has sub positions.
+    /// </summary>
+    /// <param name="actor">Target actor</param>
+    /// <param name="subPosition">Target sub position, 1 based</param>
+    public void AssignActorToSubPosition(string actor, int subPosition)
+    {
+        if (_activeScene == null || !_activeScene.HasSubPositions())
+        {
+            Debug.LogError("No active scene or active scene doesn't have sub positions");
```
PhoebeMitchell: Split these into two separate checks

Assets/Scripts/BGScenes/BGScene.cs
```diff
@@ -4,15 +4,81 @@
 
 public class BGScene : MonoBehaviour
 {
+    [SerializeField] private List<SubPosition> _subPositions;
```
PhoebeMitchell: Add tooltip

Assets/Scripts/BGScenes/BGSceneList.cs
```diff
@@ -29,16 +29,15 @@ void Start()
...
-        }
-
         try
         {
-            _activeScene = _scenesInChildren[sceneName];
-            _activeScene.gameObject.SetActive(true);
+            BGScene targetScene = _scenesInChildren[sceneName];
+            if (_activeScene == null || targetScene != _activeScene)
+            {
+                _activeScene?.gameObject.SetActive(false);
```
PhoebeMitchell: Change to `if (targetScene != _activeScene)
            {
                if (_activeScene != null)
                {
                    _activeScene.gameObject.SetActive(false);
                }` to avoid bypassing object lifetime with ? and checking if _activeScene is null twice.

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -165,6 +168,29 @@ private void PlayEmotion(string animation)
...
+    private void SetActorPosition(string positionAndActor)
+    {
+        if (!HasActorController())
+            return;
+
+        string[] parameters = positionAndActor.Split(ACTION_PARAMETER_SEPARATOR);
+
+        if (parameters.Length != 2)
+        {
+            Debug.LogError("Invalid amount of parameters for function SET_ACTOR_POSITION");
```
PhoebeMitchell: Major nit: "Invalid number" instead of "invalid amount" because parameters are countable.

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -372,6 +398,45 @@ void Wait(string seconds)
...
+    void PanToSubPosition(string positionAndTime)
+    {
+        if (!HasSceneController())
+            return;
+
+        string[] parameters = positionAndTime.Split(ACTION_PARAMETER_SEPARATOR);
+
+        if (parameters.Length != 2)
+        {
+            Debug.LogError("Invalid amount of parameters for function PAN_TO_POSITION");
```
PhoebeMitchell: Major nit: Number instead of amount.

Assets/Resources/InkDialogueScripts/Ross_Cool_X.ink
```diff
@@ -0,0 +1,210 @@
+//Do these before fading in, sets up the scene
+
+&SCENE:TMPH_Judge
+&ACTOR:Brent_Judge
+
+&SCENE:TMPH_Assistant
+&ACTOR:Dan
+
+&SCENE:TMPH_CourtPan
```
PhoebeMitchell: The prefab for this BGScene seems to be called TMPH_Court so this gives errors when trying to play (unless it's just my unity being weird and not updating properly, ignore if that is the case). After changing all "CourtPan"s to "Court" everything worked fine.

PhoebeMitchell (dismissed):
Functions well. Couple of nits.

Assets/Scripts/BGScenes/BGSceneList.cs
JuiceGrape: Couldn't I techincally only do target scene != active scene and then use the inline null check (the ?)

Assets/Resources/InkDialogueScripts/Ross_Cool_X.ink
JuiceGrape: Ah, renamed it later, will change back.

Assets/Scripts/BGScenes/BGSceneList.cs
PhoebeMitchell: Yeah, though I was under the impression we agreed == null was preferred in a previous PR. Could not say which one though. Personally I dont mind. "?" will think an object has not been destroyed if it's destroyed on the same frame as the check, but I don't suppose _activeScene will ever be destroyed?

notexplosive (dismissed):

Assets/Scripts/BGScenes/BGScene.cs
```diff
@@ -4,15 +4,81 @@
...
+            return null;
+        }
+        return _subPositions[actualPosition].AttachedActor;
     }
 
     
 }
+
+[System.Serializable]
+public struct SubPosition
```
ViMaSter: Nit: I find the term "SubPosition" to be confusing for two reasons:
1. "Sub" implies the existence of something like a Superposition this SubPosition relates to (like an offset), however these positions are absolute inside a scene and have no other point of reference.
2. The class `SubPosition` actually contains a property `position` which makes it difficult to refer to them individually and infer the meaning of the word `position` in either context

Suggestion: Renaming `SubPosition` to `ActorLocationSlot` (or just `ActorSlot`) and making it part of the `BGScene` class. That way, it will always be referred to as `BGScene.Actor(Location)Slot` anywhere else, which describes its use (a list of possible positions that actors can take inside a scene) more accurately and clears up the `global::`-namespace.

Assets/Scripts/Actor/ActorController.cs
```diff
@@ -177,4 +187,31 @@ public void SetSpeakingType(SpeakingType speakingType)
...
+
+        try
+        {
+            tempActor.SetActor(_actorInventory[actor]);
+        }
+        catch (KeyNotFoundException exception)
+        {
+            tempActor.SetActor(null);
+            Debug.Log($"{exception.GetType().Name}: Actor {actor} was not found in actor dictionary");
+            return;
```
ViMaSter: Nit: `return` is superfluous.

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -372,6 +398,45 @@ void Wait(string seconds)
...
+            return;
+        }
+        if (int.TryParse(parameters[0], out int pos) && float.TryParse(parameters[1], NumberStyles.Any, CultureInfo.InvariantCulture, out float seconds))
```
ViMaSter: Nit: This can silently fail if 2 parameters exist but can't be parsed to an integer.  
Suggestion: Invert and split these `if`s into individual conditions and handle failed parses with a call to `Debug.LogError()`.

Assets/Scripts/BGScenes/BGSceneList.cs
ViMaSter: Null Coalescing and Null Conditional operators inside Unity also have the added downside of [hiding their intent](https://github.com/JetBrains/resharper-unity/wiki/Possible-unintended-bypass-of-lifetime-check-of-underlying-Unity-engine-object#null-conditional-operator).  
A reference can be "not null", although the underlying UnityEngine Object has been destroyed. Both can be checked for and it's unclear which one is taking place written like that.

There's also [an official blog post](https://blog.unity.com/technology/custom-operator-should-we-keep-it) about this topic.

Assets/Scripts/Interfaces/ISceneController.cs
ViMaSter: Nit: `Jump` is used instead of `Set` inside the Ink scripts and feels more appropriate in this context too. Preferably we'd use `Jump` all the way down for logic related to to jump to a specific camera position without any transition.

ViMaSter (dismissed):
Looks great generally! Nothing majorly blocking, just a couple of nits.

---

Looking at the Ink script, I was wondering if it would be more intuitive, to name the SubPositions instead of accessing them by their index. This would make the scripts more human-readable, plus reduce the room for error with 1-based indexing inside .ink-files and 0-based indexing everywhere else.

Right now the .ink-files look like this:
```
&SCENE:TMPH_Court
&SET_ACTOR_POSITION:1,Arin
&SET_ACTOR_POSITION:2,Ross
&SET_ACTOR_POSITION:3,Tutorial_Boy

...

&SCENE:TMPH_Court
&JUMP_TO_POSITION:2
```
which maps to SubPositions like so:
image
A possible alternative would be 
```
&SCENE:TMPH_Court
&SET_ACTOR_POSITION:Witness,Ross
&SET_ACTOR_POSITION:Prosecution,Tutorial_Boy

&SCENE:TMPH_Court
&JUMP_TO_POSITION:Witness
```
which could be easily achieved by using the name of the gameObject of the `Actor`-component of a specific SubPosition.

@notexplosive @TooManyMoths Since I'll be taking over the PR from @JuiceGrape and you've already added a few comments, do you feel this would be worth implementing?
Quick 👍 (implement it in a later PR) 👎 (no need for it) works too.

In the meantime I'll resolve the conversations and either push a few fixes if needed or merge this in to unblock other stories.

Assets/Scripts/Interfaces/ISceneController.cs
bas-mentjot: Setting the camera position is more logical in my opinion, but jump is more of a "movie" term, which is why we use it on the top level

notexplosive (dismissed):

Assets/Scripts/Actor/ActorController.cs
```diff
@@ -117,7 +113,8 @@ public void PlayEmotion(string emotion)
     /// <summary>
     /// Sets the active speaker in the scene, changing the name shown.
     /// </summary>
-    /// <param name="actor">Target actor. This gets the correct name and colour from the list of existing actors.</param>
+    /// <param name="actor">Target actor. This gets the correct name and
```
PhoebeMitchell: Missing word "colo(u)r"

Assets/Scripts/BGScenes/BGScene.cs
```diff
@@ -1,84 +1,90 @@
...
 using System.Collections.Generic;
 using UnityEngine;
+using UnityEngine.Serialization;
 
 public class BGScene : MonoBehaviour
 {
-    [SerializeField] private List<SubPosition> _subPositions;
-    public Actor ActiveActor { get; private set; }
+    [Tooltip("List of potential actor slots in the scene that can be filled with ActorData and an associated camera position")]
+    [FormerlySerializedAs("_ActorSlots")]
```
PhoebeMitchell: Can be safely removed now Unity has compiled once.

Assets/Scripts/BGScenes/BGSceneList.cs
```diff
@@ -26,23 +23,27 @@ void Start()
...
+        if (!_scenesInChildren.ContainsKey(sceneName))
         {
-            BGScene targetScene = _scenesInChildren[sceneName];
-            if (_activeScene == null || targetScene != _activeScene)
-            {
-                _activeScene?.gameObject.SetActive(false);
-                _activeScene = targetScene;
-                targetScene.gameObject.SetActive(true);
-            }
+            Debug.Log($"BGScene '{sceneName}' was not found in bg-scenes dictionary");
```
PhoebeMitchell: Nit: Make this a LogError or a LogWarning.

PhoebeMitchell (dismissed):
Looks good. In regard to the question about referring to subpositions by name, I would argue it would be more intuitive for authors to use the SCENE command instead of the JUMP_TO_POSITION command, and having the system work out if the scene is actually a position, as to authors changing scene and jumping to a position look indentical. Unsure how feasible this is, so in the meantime I like the suggestion of referring to positions (or slots now?) by name if we wanted to do that in future (or referring to them by index _or_ name 👀 )

PhoebeMitchell (approved):

notexplosive (approved):
✔️ ✔️ ✔️ ✔️ ✔️ ✔️

PR #81, opened 2021-09-25
Bugfix/68 present evidence outside of cross examination dialogue options

### LINKED ISSUE #68: [BUG] Able to present evidence outside of presenting moments

ViMaSter:
## Summary

Resolves the bug outlined inside https://github.com/Studio-Lovelies/GG-JointJustice-Unity/issues/68 (presenting evidence outside of presenting moments) and introduces automated tests for it, so this bug is automatically caught in the future.  
Also contains general housekeeping for all files affected by this fix (typos, unused `using`s, etc.), will move those out to separate PRs in the future.

---

## Testing instructions

**Either...**
1. Follow the steps inside the related issue (https://github.com/Studio-Lovelies/GG-JointJustice-Unity/issues/68)

**...or...**

1. Open the Unity Editor
2. Run the following test cases:
    * `CrossExamination.ViaKeyboard.CanPresentEvidenceDuringExamination`
    * `CrossExamination.ViaKeyboard.CantPresentEvidenceDuringExaminationDialogue`
    * `CrossExamination.ViaKeyboard.CantPresentEvidenceDuringPressingDialogue`
3. Notice all tests succeeding

## Additional information

- `[ ]` Changes UI
  closes #68


Assets/Scripts/TextDecoder/DialogueController.cs
```diff
@@ -343,7 +348,7 @@ private bool IsAction(string line)
     /// <param name="subStory">Inky dialogue script to be set as the sub story</param>
     public void StartSubStory(TextAsset subStory)
     {
-        _subStory = GameObject.Instantiate(_dialogueControllerPrefab); //Returns the DialogueController component attached to the instantiated gameobject
+        _subStory = Instantiate(_dialogueControllerPrefab); //Returns the DialogueController component attached to the instantiated gameobject
```
JuiceGrape, line 363: Nit: Referring to static methods like this makes them seem non static. The GameObject. was there for a reason

Assets/Scripts/TextDecoder/DialogueController.cs
```diff
@@ -330,7 +335,7 @@ public void SetBusy(bool busy)
     /// </summary>
     /// <param name="line">Line to check</param>
     /// <returns>Whether the line is an action or not</returns>
-    private bool IsAction(string line)
+    private static bool IsAction(string line)
```
JuiceGrape, line 350: Nit: Private and Static are polar opposites of each other. Doesn't make sense for something to be both. Furthermore, this function isn't being used anywhere else, so it shouldn't be static

JuiceGrape (dismissed):
Two nits, looks great otherwise

Assets/Scripts/TextDecoder/DialogueController.cs
PhoebeMitchell: Though very negligable in our case, marking members as static is a dotnet performance rule https://docs.microsoft.com/en-gb/dotnet/fundamentals/code-analysis/quality-rules/ca1822?view=vs-2019&redirectedfrom=MSDN&viewFallbackFrom=vs-2015
I would argue there's no reason for this _not_ to be static.

Assets/Scripts/TextDecoder/DialogueController.cs
PhoebeMitchell: This should be added to our code style doc if we are doing this, though personally I do not think it is necessary. Unity docs and source code references all just use Instantiate, favouring Object.Instantiate when not in a MonoBehaviour.

lyndsiWilliams (dismissed):
Pulled it down and tested it locally, everything looks/works great!

Assets/Scripts/TextDecoder/DialogueController.cs
ViMaSter: I'm actually agreeing with moths' statement here. `const`, `static` and `readonly` encapsulate things (by either making them a compiler constant, prevent access to member values or prevent them from being reassigned, respectively) which not only introduces (minor) performance enhancements, but enhances legibility too. A static method only has access to what's provided as arguments, pre-existing constants or other `static` members.

I tend to use `private static` methods for utility functions that are only needed in that specific class. A method should be `private` if it only exists in that classes scope and `static` if it's not accessing or modifying any instance members. Both of which apply here.
That being said, I can definitely see that sentiment applying for `private static` fields or properties, which - in almost all cases - are better suited to be marked `const` or part of each individual instance.

Assets/Scripts/TextDecoder/DialogueController.cs
ViMaSter: `This should be added to our code style doc if we are doing this`
Great idea! I personally also don't see the need for explicitly marking static members in any way (hence my change : D), but similarly to using explicit types when initializing variables, I'm happy to follow whatever most people prefer. I'll resolve this specific conversation for now, start a vote on Discord and make a note inside the Code Style Document and adhere to that moving forward.

ViMaSter:
This PR now contains the changes necessary to prevent the user from closing the `EvidenceMenu` using the 'present evidence'-key, unless the game allows the user to present evidence at this point. The UI shouldn't close and the story will no longer be affected by it.

Quick note:
I've renamed `EvidenceController::OpenEvidenceMenu()` to `EvidenceController::RequirePresentEvidence()`:
This method currently not only opens the menu (like a press on `Z`) but also **forces the user to present evidence to continue**; which is different to the user **just opening the evidence menu to check on existing evidence or when (optionally) presenting evidence during cross examinations**.  
This way, the term `open` is reserved for when only the visibility is changed and `require present evidence` is used, when evidence has to be presented to continue with the story and close the window.

Assets/Scripts/Evidence/EvidenceMenu.cs
```diff
@@ -40,14 +40,18 @@ public class EvidenceMenu : MonoBehaviour
     [SerializeField, Tooltip("This event is called when a piece of evidence has been clicked.")]
     private UnityEvent _onEvidenceClicked;
 
-    private bool _profileMenuActive;
+     private bool _profileMenuActive;
```
notexplosive: I think you accidentally added a spa ce

notexplosive (dismissed):
There's a lot of changes here so I might have missed something, but it sounds like this was already gone over with a few other people so here's one little nit just to ruin your day :p

Assets/Scripts/Evidence/EvidenceMenu.cs
ViMaSter: Fixed in https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/81/commits/59168ca4300bc3f7474c8c19e9fff848f401c38a, thereby invalidating your approval, so now who's the one whose day is ruined?! :P



Thanks for spotting it though, not sure how I missed that. : S

notexplosive (dismissed):

Assets/Scripts/Evidence/EvidenceController.cs
```diff
@@ -19,6 +20,16 @@ public class EvidenceController : MonoBehaviour, IEvidenceController
     [Tooltip("Drag an EvidenceDictionary component here.")]
     [SerializeField] public EvidenceInventory _evidenceInventory;
 
+    [Tooltip("Drag an EvidenceMenu component here, which will updated when the game state (i.e. ability to present evidence) changes.")]
+    [SerializeField] public EvidenceMenu _evidenceMenu;
+
+    /// Called either when invoking <see cref="DialogueController._onCrossExaminationLoopActive" />
```
lyndsiWilliams, line 27: Can you make this comment a summary? ✨

Assets/Scripts/TextDecoder/DialogueController.cs
```diff
@@ -186,79 +200,91 @@ public void HandlePresenting(ICourtRecordObject presentedObject)
...
+            HandleChoice(evidenceFoundAt);
+            return;
+        }
+
+        //Deal with bad consequences, spawn sub story and continue that
+        StartSubStory(_failureList.GetRandomFailurestate());
+    }
+
+    /// <summary>
+    ///
```
lyndsiWilliams: This summary is missing ✨

lyndsiWilliams (dismissed):
Pulled down your branch and tested it, it works and the tests pass! Just a couple of nits here, this works great!

JuiceGrape (dismissed):

Assets/Scripts/TextDecoder/DialogueController.cs
```diff
@@ -167,17 +168,30 @@ public void SetMenuOpen(bool isMenuOpen)
...
+        if (_subStory)
+        {
+            _subStory.HandlePresenting(presentedObject);
+            return;
+        }
+
+        if (!_isAtChoice)
+        {
+            return;
+        }
```
PhoebeMitchell, line 186: A question for @JuiceGrape as much as @ViMaSter but is presenting evidence _supposed_ to be handled as a choice? It works in EvidenceAddingTestScript if I do the following:

```
There, I added coins.

&PRESENT_EVIDENCE
+ [Choice1] -> Line1

=== Line1 ===
I'm going to add more
```

I was under the impression there would be one correct evidence to present and if you got it right you would continue and if you got it wrong it would play a sub-story. Seems redundant to require the extra choice lines.

ViMaSter: I personally can live with either solution and was just building based on the existing implementation.

As long as only limiting it to one correct choice works, I'd go with whatever works best for the writers.

JuiceGrape: So the way it works is that presenting wrong evidence gives you a failure state, but then makes you continue normally. This is simply how the game works. In cross examination the first line MUST be continue and the second line MUST be press, that's just how its handled. These can loop back upon themselves if so required. 3+ is going to be any evidence

JuiceGrape: So in your case moths, your dialogue script is wrong, not the code

PhoebeMitchell: Yeah, I suspected I was just misunderstanding how to handle presenting evidence.

PhoebeMitchell (dismissed):
Appears to function. Would be nice if you could apply your changes to the standard inclusion prefab. EvidenceController is missing its EvidenceMenu reference and DialogueController is missing listeners for OnCrossExaminationLoopActive.

lyndsiWilliams (dismissed):
Code looks great! Pulled down your branch locally and it's still working great and all your tests pass. Nice!

notexplosive (dismissed):
I'm gonna assume with how much iteration this has had that it's done.

notexplosive (approved):
🚀

PhoebeMitchell (approved):
Bug appears to be fixed and tests pass.

PR #82, opened 2021-09-25
Feature/70 story handler

### LINKED ISSUE #70: Implement story handler
## Description

Each unity scene can consist of multiple inky dialogue scripts. To make sure this is possible, the story controller should be implemented. It will hold a list of inky dialogue scripts to be played sequentially, with transitions between them (like fading to black or showing a zoomed out view, or no transition at all).

To make sure cross examination parts work properly, each of these scripts should also be flagged with a  DialogueControllerMode value, so the DialogueController can be set into the correct mode when loading the new inky script.

This script should also handle transitioning to the next unity scene when done.

## Dev Guide

- Make sure to use _onDialogueFinished event to continue when the current inky dialogue script is done
- SetNarrativeScript currently only takes a new inky dialogue script, so this should be extended to also take a DialogueControllerMode
- The first action in a script should be automatically called, which it currently isn't.

## Acceptance Criteria

This story will be finished when the aforementioned functionality is implemented and a first pre-trial unity scene is set up to test it all and set up some of the actual working behaviour going forward.

JuiceGrape:
## Summary
Stories are now handled by the story handler, which can transition between different inky dialogue scripts. It also loads the next unity scene when reaching the end

## Testing instructions
Check out the inky test scene

## Additional information

- `[ ]` Changes UI
  
closes #70


Assets/Scripts/TextDecoder/StoryController.cs
```diff
@@ -1,18 +1,68 @@
 using System.Collections;
 using System.Collections.Generic;
 using UnityEngine;
+using UnityEngine.Events;
 
+[RequireComponent(typeof(SceneLoader))]
 public class StoryController : MonoBehaviour
 {
-    // Start is called before the first frame update
+    [SerializeField] private List<Dialogue> _dialogueList;
```
PhoebeMitchell, line 9: Nit: Add tooltip

Assets/Scripts/TextDecoder/StoryController.cs
```diff
@@ -1,18 +1,68 @@
...
+
+    [Header("Events")]
+
+    [SerializeField] private UnityEvent<Dialogue> _onNextDialogueScript;
```
PhoebeMitchell, line 13: Nit: Add tooltip

Assets/Scripts/TextDecoder/StoryController.cs
```diff
@@ -1,18 +1,68 @@
...
+    /// </summary>
     void Start()
     {
-        
+        if (_dialogueList.Count == 0)
+        {
+            Debug.LogError("This unity scene does not have any dialogue scripts");
+            return;
+        }
+        _sceneLoader = GetComponent<SceneLoader>();
```
PhoebeMitchell: Nit: Component getting should be done on awake to avoid potential script execution order issues.

Assets/Scripts/TextDecoder/StoryController.cs
```diff
@@ -1,18 +1,68 @@
...
-    // Start is called before the first frame update
+    [SerializeField] private List<Dialogue> _dialogueList;
+
+    [Header("Events")]
+
+    [SerializeField] private UnityEvent<Dialogue> _onNextDialogueScript;
+
+
+    private SceneLoader _sceneLoader;
+
```
PhoebeMitchell, line 16: Nit: Put private fields together, without space.

Assets/Scripts/TextDecoder/StoryController.cs
```diff
@@ -1,18 +1,68 @@
...
-    void Update()
+}
+
+[System.Serializable]
+public struct Dialogue
+{
+    public Dialogue(TextAsset script, DialogueControllerMode type)
     {
-        
+        DialogueScript = script;
```
PhoebeMitchell, line 71: Nit: Rename to narrativeScript

PhoebeMitchell (dismissed):
Looks good and functions properly. Found a bug where you can loads lots of scenes at once, but technically its an oversight from one of my PRs as opposed to a fault with this PR, so approving.

Assets/Scripts/SceneLoading/SceneLoader.cs
```diff
@@ -82,6 +84,7 @@ public void LoadSceneCallback()
     /// </summary>
     private IEnumerator LoadSceneCoroutine()
     {
+        Busy = true;
```
PhoebeMitchell: This needs to be moved to the beginning of the Transition method, and potentially it should be checked in the ChangeSceneBy methods (instead of StoryController) as its possible to spam enter on the main menu and spam x in the inky test scene (when using a fade to image transition) to start loading lots of scenes.

Assets/Scripts/TextDecoder/StoryController.cs
JuiceGrape: Turns out tooltips on events don't actually do anything

Assets/Scripts/SceneLoading/SceneLoader.cs
JuiceGrape: Oh I totally forgot I was fixing this! I'll do it properly

Assets/Scripts/TextDecoder/StoryController.cs
```diff
@@ -1,18 +1,68 @@
 using System.Collections;
 using System.Collections.Generic;
 using UnityEngine;
+using UnityEngine.Events;
```
ViMaSter, line 3: Nit: `using System.Collections;` seems to be no longer required after these changes.

Assets/Scripts/TextDecoder/StoryController.cs
```diff
@@ -1,18 +1,68 @@
...
+
+    private SceneLoader _sceneLoader;
+
+    private int _currentStory = -1;
+
+    /// <summary>
+    /// Initializes the script and starts the first dialogue script.
+    /// This script should always be last in the run order in order for this to work properly.
+    /// </summary>
     void Start()
```
ViMaSter: Nit: Missing explicit (`private`) visibility modifier.

Assets/Scripts/TextDecoder/StoryController.cs
```diff
@@ -1,18 +1,68 @@
...
+{
+    public Dialogue(TextAsset script, DialogueControllerMode type)
     {
-        
+        DialogueScript = script;
+        ScriptType = type;
     }
+
+    public TextAsset DialogueScript;
+    public DialogueControllerMode ScriptType;
```
ViMaSter, line 76: Nit: Current Code Style requires all member variables to be `_lowerCamelCase` (i.e. `_dialogueScript`, `_scriptType`).

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuKeyboardTests.cs
```diff
@@ -273,7 +273,7 @@ public IEnumerator ProfileMenuCanBeAccessed()
         public IEnumerator EvidenceCanBeSubstitutedWithAltEvidence()
         {
             SceneManager.LoadScene("EvidenceMenu - Test Scene", LoadSceneMode.Single);
-            yield return null;
+            yield return new WaitForSeconds(2);
```
ViMaSter: I personally think that adding arbitrary delays to tests (especially all cases) is not a good practice. Is there an underlying reason for this change?

Wild guess: If you're waiting for an object a reach a certain state, I'd suggest adding logic to explicitly wait for that state, as a simple delay doesn't guarantee that your system under test actually reaches the desired state under different conditions (i.e. on a slower machine) and the test can end up failing for the exact same reason.

If that's the case, my suggestion would be to add a static method to the `Tests.PlayModeTests.Tools.InputTestTools` which takes a function (that evaluates to a boolean) and timeout as parameters and returns an `IEnumerator` that can be yielded. It waits for either `X` seconds to pass (and fails the test) or continues operation as soon as the condition is satisfied.  
The timeout can be set fairly firm and loosened, if it ends up resolving the issue. If it doesn't. there's clear information available on what expected condition is not fulfilled.

Something along the lines of:
```cs
public static IEnumerator WaitForState(Func<bool> hasReachedState, double timeoutInSeconds)
{
    DateTime timeoutAt = DateTime.Now.AddSeconds(timeoutInSeconds); // we cannot rely on UnityEngine.Time inside tests
    while (DateTime.Now < timeoutAt)
    {
        yield return null;
        if (hasReachedState())
        {
            break;
        }
    }
    Assert.Fail($"Waiting for an expected state timed out (took longer than {timeoutInSeconds} seconds){Environment.NewLine}If increasing the timeout does not resolve this and the associated expression worked before making any changes to this project, a precondition required for this test has changed and needs to be updated accordingly");
}
```
which then could be used using `yield return WaitForState(() => readyWhenIsFour == 4, 5);` )where `readyWhenIsFour == 4` would be whatever expression needs to be true for the tests to execute reliably) to replace the timeout.

Assets/Scripts/SceneLoading/SceneLoader.cs
```diff
@@ -35,7 +37,7 @@ private void Awake()
     /// <param name="menuNavigator">The menu navigator used to call this method. It is passed in case it needs to be disabled</param>
```
ViMaSter: Nit: Pointing it out, as the contents of this method changed; this parameter comment appears to be superfluous.

Assets/Scripts/SceneLoading/SceneLoader.cs
```diff
@@ -45,7 +47,7 @@ public void ChangeSceneBySceneIndex()
...
```
ViMaSter: Nit: Pointing it out, as the contents of this method changed; this parameter comment appears to be superfluous.

Assets/Scripts/SceneLoading/SceneLoader.cs
```diff
@@ -35,7 +37,7 @@ private void Awake()
...
     public void ChangeSceneBySceneIndex()
     {
-        _sceneLoadOperation = SceneManager.LoadSceneAsync(_sceneIndex, LoadSceneMode.Additive);
+        _sceneLoadOperation = SceneManager.LoadSceneAsync(_sceneIndex, LoadSceneMode.Single);
```
ViMaSter: Nit: It appears this method (and thereby `_sceneIndex` too) is unused (unless it's part of an Animation-script). Rather than updating it, I would suggest deleting it until it's actually required.

Assets/Scripts/SceneLoading/SceneLoader.cs
ViMaSter: Note: Unless it's part of an Animation somewhere, this property is also never set back to `false`.

Assets/Scripts/TextDecoder/DialogueController.cs
```diff
@@ -25,7 +25,7 @@ public class DialogueController : MonoBehaviour
     private const char ACTION_TOKEN = '&';
 
 
-    [SerializeField] private TextAsset _narrativeScript;
+    private TextAsset _narrativeScript;
```
ViMaSter: Nit: This field is unused now and should be safe to delete.

ViMaSter (changes requested):
This looks solid! Features that are getting rid of `TODO`/placeholder code are always great to see!

That said, [I don't agree with introducing arbitrary timeouts to the tests](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/82#discussion_r717702127). From my point of view, they don't introduce stability and are probably a symptom of a deeper issue. I've added a suggestion on how I would go about it and it would be great if we could find a stable way around adding 14 seconds of idle time (which is around 50% of my average local run-time for all) to the tests.

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuKeyboardTests.cs
JuiceGrape: @TooManyMoths this was your doing, please check this out :)

PhoebeMitchell: This was added to wait for the first line of dialogue to finish playing so the menu could be opened. Possible changes could be using AppearingTextController's OnDialogueEnd event, or waiting for the DialogueController's busy property to be false.

Assets/Scripts/SceneLoading/SceneLoader.cs
PhoebeMitchell: Assuming this method always leads to a scene load (maybe not if there is an error?), it will be false in the instance of this class in the next scene.

Assets/Scripts/TextDecoder/StoryController.cs
PhoebeMitchell: I think we never accounted for using public fields in the code style doc, assuming we'd always use properties with PascalCase. On that note should these be properties?

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuKeyboardTests.cs
JuiceGrape: I'll let you guys fight this one out and open a new pr when the time comes. For now, it's functional

Assets/Scripts/TextDecoder/StoryController.cs
JuiceGrape: I'll only do this if I mean to group them, and in this case I didn't want to group them

Assets/Scripts/TextDecoder/StoryController.cs
JuiceGrape: I think DialogueScript works well enough

Assets/Scripts/SceneLoading/SceneLoader.cs
JuiceGrape: Fixed it

Assets/Scripts/TextDecoder/StoryController.cs
JuiceGrape: As it's a struct purely for data storing purposes, it's mostly important how any instance of this is handled, as opposed to the struct itself. Seeing as they are public variables, I handled them as we have handled public variables (or properties) so far.

Assets/Scripts/SceneLoading/SceneLoader.cs
JuiceGrape: For now. I'd rather keep this as it's logical functionality to have in a class like this

PhoebeMitchell (dismissed):
Looks good and functional. I made a quick commit to fix the tests.

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuKeyboardTests.cs
ViMaSter: Just had a quick chat with Moths on it and we're not disagreeing; we're both in agreement actually! Moths simply offered guidance on the exact state change to wait for inside the new method I proposed and we would like that to be integrated.

The reason for the test failing here, is that the `AppearingTextController` is still animating the text box, not that 2 seconds less have passed (which is only a side effect on lower-end machines), so this is what we should be asserting inside the tests.

Also please don't resolve conversations of other people. Especially when they are explicitly marked as a required change.  
The concept of reviews inside PRs is to make sure things aren't just functional, but adhere to a technical design we all agree on and understand. Proposed alternative code is always up for debate, but the current implementation is something I can not approve: Arbitrary timeouts are a sign of flakey test, which are worse than not having them at all and will result to them being abandoned in the long run, if they don't provide a clear value. Tests are implemented to add stability and allow clear pinpointing of misbehavior.

Assets/Scripts/SceneLoading/SceneLoader.cs
ViMaSter: Gonna put this up for debate on Discord. In my mind, app-specific code that isn't run, should generally be removed. (The only exception being a library, where its entire purpose is presenting a set of logic while not being opinionated; even then however, methods should be covered with tests and thereby are used at least once.)  
For people using less sophisticated IDEs and/or trying to familiarize themselves with a new section of code, only having code that reflects current use cases is easier to understand too.

That said, this is nothing worth blocking for me.

ViMaSter (changes requested):
The [test changes](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/82#discussion_r720745487) are not implemented; making changes only to game logic is fine, but removing tests or changing their level of stability to me is not, as we've agreed on creating these tests for a reason.

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuKeyboardTests.cs
JuiceGrape: Added these changes, please make sure they're okay :)
Feel free to resolve this convo / merge if everything is a-o-fine.

Assets/Scripts/TextDecoder/StoryController.cs
```diff
@@ -1,18 +1,77 @@
...
+            _onNextDialogueScript.Invoke(_dialogueList[_currentStory]);
+        }
+    }
+
+}
+
+[System.Serializable]
+public struct Dialogue
+{
+    public Dialogue(TextAsset script, DialogueControllerMode type)
```
lyndsiWilliams, line 69: This method is missing a summary 🎃

lyndsiWilliams (approved):
Code looks great, just one nit. I pulled down your branch locally and played around with it, it works great! Nice job on this 😁

Assets/Scripts/TextDecoder/StoryController.cs
JuiceGrape: It's a pretty default constructor in a struct. We can probably leave it un-summarized?

lyndsiWilliams: I'd like a summary just to be sure, I'd rather have an explicit definition instead of assuming/having to read through code to find out.

JuiceGrape: Oki, will do :)

PR #83, opened 2021-09-26
Feature/57 add support for full screen animations

### LINKED ISSUE #57: Add support for full screen animations
## Description

The PLAY_ANIMATION action should be added which plays a full screen animation. These animations should exist as assets and all be gettable from there like the actorController does with actors. 
The system is considered busy until the animation finishes playing (or whenever we decide within that animation to unbusy it (with an OnAnimationComplete)).

Current list of animations:
- Ross's galaxy brain scene
- Gavel hit

## Acceptance Criteria

This story will be finished when the PLAY_ANIMATION action is implemented and added to the standardInclusion prefab.

PhoebeMitchell:
## Summary

Adds support for full screen animations e.g. Gavel hit, Ross galaxy brain, witness testimony intro. Reuses code for playing actor animations by extracting the animation code into AnimatableObject class, which Actor inherits from.

galaxybrain
gavelhit

## Testing instructions

Open "Inky-Test Scene" and run it with the InkTest and GavelHitDemo narrative scripts.

## Additional information

- `[ ]` Changes UI
  
closes #57


lyndsiWilliams (dismissed):
Pulled your branch down locally to test it out, everything looks/works great!

Assets/Scripts/AnimatableObject.cs
```diff
@@ -0,0 +1,55 @@
+using System.Collections;
+using UnityEngine;
+using UnityEngine.Events;
+
+[RequireComponent(typeof(SpriteRenderer)), RequireComponent(typeof(Animator))]
+public class AnimatableObject : MonoBehaviour
```
notexplosive: Consider class name: `Animatable`. "Object" is implied.

Assets/Scripts/AnimatableObject.cs
```diff
@@ -0,0 +1,55 @@
...
+        Animator = GetComponent<Animator>();
+    }
+    
+    /// <summary>
+    /// Checks if an animation is on the animator controller (if one is present) and plays it.
+    /// </summary>
+    /// <param name="animation"></param>
+    public void PlayAnimation(string animation)
+    {
+        int animationHash = Animator.StringToHash(animation); // Required for HasState method
```
notexplosive: What's the comment on the far right trying to tell me? I can see that animationHash is used for HasState, it's right there in the code 😛 
edit: It looks like you didn't write this comment (at least not in this PR), but I don't think we need to keep it.

notexplosive (dismissed):

Assets/Scripts/AnimatableObject.cs
```diff
@@ -0,0 +1,55 @@
+using System.Collections;
```
ViMaSter: Nit: It seems like no symbols from that namespace are being used in this file. I'd recommend removing it.

Assets/Scripts/AnimatableObject.cs
```diff
@@ -0,0 +1,55 @@
...
+using UnityEngine;
+using UnityEngine.Events;
+
+[RequireComponent(typeof(SpriteRenderer)), RequireComponent(typeof(Animator))]
+public class AnimatableObject : MonoBehaviour
+{
+    [SerializeField, Tooltip("This event is called when an animation begins.")]
+    private UnityEvent _onAnimationBegin;
```
ViMaSter: Nit: I'd personally prefer `_onAnimationStart` to have vocabulary consistent with Unity.

Assets/Scripts/AnimatableObject.cs
```diff
@@ -0,0 +1,55 @@
...
+    protected virtual void Awake()
+    {
+        Animator = GetComponent<Animator>();
+    }
+    
+    /// <summary>
+    /// Checks if an animation is on the animator controller (if one is present) and plays it.
+    /// </summary>
+    /// <param name="animation"></param>
+    public void PlayAnimation(string animation)
```
ViMaSter: Gonna reference [something from another PR](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/78#discussion_r714966396) since it applies here and we're actually modifying (or rather moving) this part of code. Plus, it's changed to `animationName` inside other places of this change already.

---

Mega nit: This is a **huge** edge-case, so feel free to disregard this, but I still wanna point it out:  

Unity previously allowed access to commonly used components via fields like `.transform`, `.rigidbody` and also `.animation`. Usage is discouraged and `GetComponent<Transform>()`/`GetComponent<Rigidbody>()`/`GetComponent<Animation>()`, but the field is still available.

With this parameter being named `animation`, it hides the `animation`-field of the deriving class `MonoBehaviour`. I'd suggest renaming it to `animationName`; to resolve this override and also make it more verbose (as we're technically passing the name of an animation not an instance of `Animation` directly at this point).

Assets/Scripts/Actor/Actor.cs
```diff
@@ -1,21 +1,22 @@
 using UnityEngine;
 using UnityEngine.Events;
```
ViMaSter: Nit: Due to the changes in this PR it seems like no symbols from those namespaces are being used in this file anymore. I'd recommend removing them.

Assets/Scripts/AnimatableObject.cs
ViMaSter: I think I have a technical design question for this change: What's been the main reason for splitting `Actor` and `AnimatableObject` into separate classes? Inside the current revision of this PR only the class which previously contained this code (`Actor`) derives from `AnimatableObject`, which didn't seem like worryingly large file or complex class to me.

Even if it's something we want to do in the future, I personally would advocate for keeping class hierarchy flat and only start abstracting logic, once doing so actively reduces the amount of duplicate code.

Assets/Scripts/BGScenes/SceneController.cs
```diff
@@ -177,7 +187,33 @@ public void HideItem()
...
-    
+
+    /// <summary>
+    /// Plays a fullscreen animation e.g. Ross' galaxy brain or the gavel hit animations.
+    /// </summary>
+    /// <param name="animationName">The name of the animation to play.</param>
+    public void PlayAnimation(string animationName)
+    {
+        if (!HasFullScreenAnimationPlayer())
+            return;
```
ViMaSter, line 199: (Not a direct action item here, this is a general technical design question, but something that might impact this PR.)
I'm wondering if creating errors about the fact, that a reference object (like a fullscreen animation player) is missing, but then continuing execution as normal otherwise is a good idea. (This also applies to how the `DirectorActionDecoder` runs checks with methods like `HasAppearingDialogController()`/`HasAudioController()` which only create errors, if it's unable to actually act on an action requested from an Ink script.)

This current design would imply that a SceneController without a full-screen animation player (or similarly a `DirectorActionDecoder` without `AppearingDialogController` or `AudioController`) is an odd, but expected scenario.  
I would make the argument, that we don't ever want only a subset of functionality to be handled:
- If a scene has no `AudioController` we couldn't play any sound
- If a scene has no `AppearingDialogController` the scene couldn't be read by a user
- If a scene has no `EvidenceController` we can't add or present new items
- If a scene has no full-screen animation player, a cutscene that's requested couldn't play

These situation - to me - read as bugs that should not allow the game to continue (for example: being unable to add evidence but allowing to continue to play the game will cause unexpected behavior in later parts of the scene and errors aren't actively shown to the user outside of development builds).
Instead of writing code to handle null-references and allow further progression, code should assert their existence and having Unity throw appropriate fatal exceptions that halt execution seem more appropriate to me.

The only case, where having parts of the system would be beneficial, is tests. That said, I can see testing the `DirectorActionDecoder` and controllers **on their own** as useful, but running tests against a partially set up `DirectorActionDecoder` with only a few controllers present doesn't seem like an ideal way to go about it.

ViMaSter (dismissed):
Such a small change of actual code compared to the impact feature-wise, love to see it!
A few nits and two (somewhat big) technical questions from my end, but nothing worth blocking; very nicely done!

Assets/Scripts/AnimatableObject.cs
PhoebeMitchell: Basically, to remind me why I converted it to a hash this one time. Because HasState doesn't support strings.

Assets/Scripts/AnimatableObject.cs
PhoebeMitchell: Personally I think it's confusing to have a component called Actor on something that isn't an actor, and renaming Actor to Animatable would imply that all Animatables have functionality like talking and extracting information from actor data.

Assets/Scripts/AnimatableObject.cs
PhoebeMitchell: Done

Assets/Scripts/AnimatableObject.cs
PhoebeMitchell: Done

Assets/Scripts/BGScenes/SceneController.cs
PhoebeMitchell: Stuff like this is mostly to help me work out what component I forgot to assign. Maybe make a discord thread on this, and make some kind of outline on how we should handle errors in future.

Assets/Scripts/AnimatableObject.cs
ViMaSter: That's actually a solid point. I've noticed it being used on Prefabs besides Actor too (only checked the code on my first pass) so this becomes a non-issue. Thanks for elaborating on it!

Assets/Scripts/BGScenes/SceneController.cs
ViMaSter: Just for the record: Feel free to mark this as resolved, if you personally don't want to wait on whatever comes out of the thread on Discord; to me it's too big of a change, that it is worth blocking the PR over (as in: if we do address it, I'd prefer addressing it in it's own, separate PR).

Assets/Scripts/AnimatableObject.cs
PhoebeMitchell: Changed to Animatable

Assets/Scripts/AnimatableObject.cs
PhoebeMitchell: Removed in latest commit

Assets/Scripts/AnimatableObject.cs
PhoebeMitchell: Changed to animationName

ViMaSter (approved):
Thanks for addressing all the nits again! LGTM now!

lyndsiWilliams (approved):
Pulled your branch down locally for a re-test, everything still looks/works great!

Issue #43, opened 2021-09-02
Writer should be able to "stop" a music track


notexplosive:
## Description

I haven't run this by anyone but I think this is a feature we want. I thought about adding it to my existing audio controller work but it seemed too feature-y to add in a bugfix.

Currently you can do `PLAYSONG:aBoyAndHisTrial` to play a song, but say we want to transition into silence. How do we do that? This could be as simple as `PLAYSONG:silence` where "silence" is a magic string that the AudioController knows how to deal with. Or perhaps a `STOPSONG` command?

## Acceptance Criteria

This story will be finished when there is some command you can write in the NarrativeScript that stops the current background song from playing.

JuiceGrape:
STOP_SONG is the chosen command, please use that

PR #84, opened 2021-09-26
Feature/43 Writer can stop music

lyndsiWilliams:
## Summary

Implements the `STOP_SONG` command, which allows writers to stop a song within an ink script.

## Testing instructions

- Run `Inky-TestScene` with the new `AudioTest` ink script in the DialogueController

## Additional information

- `[ ]` Changes UI
- [x] Introduces new feature
- [x] Has associated resource: https://github.com/Studio-Lovelies/GG-JointJustice-Unity/issues/43

### DIFF
--- a/Assets/Scripts/AudioController/AudioController.cs
+++ b/Assets/Scripts/AudioController/AudioController.cs
@@ -120,13 +120,10 @@ public class AudioController : MonoBehaviour, IAudioController
     /// </summary>
     public void StopSong()
     {
-        _isTransitioningMusicTracks = true;
         if (_musicAudioSource != null)
         {
             _musicAudioSource.Stop();
         }
-
-        _isTransitioningMusicTracks = false;
     }
 
     /// <summary>
--- a/Assets/Scripts/AudioController/AudioController.cs
+++ b/Assets/Scripts/AudioController/AudioController.cs
@@ -116,6 +116,20 @@ public class AudioController : MonoBehaviour, IAudioController
     }
 
     /// <summary>
+    /// If music is currently playing, stop it!
+    /// </summary>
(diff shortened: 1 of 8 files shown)


Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -115,6 +115,20 @@ public IEnumerator FadeToNewSong(string songName)
...
+    /// </summary>
+    public void StopSong()
+    {
+        _isTransitioningMusicTracks = true;
+        if (_musicAudioSource != null)
+        {
+            _musicAudioSource.Stop();
+        }
+
+        _isTransitioningMusicTracks = false;
```
PhoebeMitchell: No need to set _isTransitioningMusicTracks as it's never used anyway, and theoretically the FadeToSong coroutine could still be running.

PhoebeMitchell (dismissed):
Looks good! Small nit, but either way functionality is the same as far as I can tell so up to you whether you address it 👍

Assets/Scripts/AudioController/AudioController.cs
lyndsiWilliams: I was wondering! It looked like some sort of "audio is currently playing" check so I put it in just in case. Removed the check in [`this commit`](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/84/commits/fd30f446856c83d4da5c73b474d430d9fa9e208e).

PhoebeMitchell (approved):

ViMaSter:
LGTM! Ran this locally and everything seems to work fine!

ViMaSter (approved):

PR #85, opened 2021-09-27
Feature/69 add choice handling

### LINKED ISSUE #69: Add choice handling
## Description

When a choice is encountered during normal dialogue, we should be able to show the choices on screen and choose one, continuing the story from there.

## Dev Guide

Use OnChoicePresented in the DialogueController to subscribe to any choice presented. Then call the HandleChoice function (preferably through an event) in the DialogueController. This handles everything it needs to by itself

## Acceptance Criteria

This story will be finished when the aforementioned functionality is implemented and testable in a unity scene

PhoebeMitchell:
## Summary

Adds functionality for displaying choices on screen that can be clicked. This is handled using the ChoiceMenu script which uses Menu and MenuOpener scripts to create a choice menu, listens for the OnChoicePresent event on dialogue controller, and calls DialogueController's HandleChoice method via ChoiceMenu's OnChoiceClicked event.

Also unticks "deselect on background click" on EventSystems and removes code related to keeping buttons selected which is now handled by the EventSystems.

Edit: ~~need to fix a bug where 2 buttons can be selected at once~~ Should be fixed now.

choicehandling

## Testing instructions

Open "Ink-Test Scene" and run the ChoiceTest narrative script.

## Additional information

- `[x]` Changes UI
  
closes #69


Assets/Prefabs/StandardInclusion.prefab
```diff
@@ -384,7 +384,19 @@ MonoBehaviour:
       m_Calls: []
   _onChoicePresented:
     m_PersistentCalls:
-      m_Calls: []
+      m_Calls:
```
JuiceGrape, line 585: Single comment to add to this: There seems to be an issue where only 1 choice sequence can happen. If you input the ross_x_examination inky script as a regular dialogue script, you'll see that only the first choice is ever presented. The second time a choice is hit, the system doesn't show anything

JuiceGrape (dismissed):
With the bug fixed, everything looks good :)

Assets/Prefabs/StandardInclusion.prefab
JuiceGrape: Fixed

Assets/Scripts/Menu/MenuItem.cs
```diff
@@ -11,45 +13,90 @@
...
 public class MenuItem : MonoBehaviour, ISelectHandler, IDeselectHandler, IPointerEnterHandler
 {
-    private Selectable _selectable;
+    [Tooltip("Drag the menu item's text component here.")]
+    [SerializeField] private TextMeshProUGUI _textMeshProUGUI;
+    
     private Menu _menu;
     private IHighlight _highlight;
     
+    public Selectable Selectable { get; set; }
```
notexplosive: Should this be a `private set`?

Assets/Scripts/Menu/ChoiceMenu.cs
```diff
@@ -0,0 +1,87 @@
...
+    /// <returns>Whether the required components have been assigned (true) or not (false).</returns>
+    private bool HasRequiredComponents()
+    {
+        if (_menu != null && _menuOpener != null)
+        {
+            return true;
+        }
+        
+        string missingComponent = _menuOpener == null ? "MenuOpener" : "Menu";
+        Debug.LogError($"{missingComponent} has not been assigned to {name} on {gameObject.name}.", this);
```
notexplosive: nit-ish: What if both components are missing? What error do we log then?
(You don't actually need to fix this, it's fine as it is)

Assets/Scripts/Menu/Menu.cs
```diff
@@ -47,15 +43,20 @@ public void SelectInitialButton()
...
             Selectable selectable = GetComponentInChildren<Selectable>();
             if (selectable.interactable)
             {
                 selectable.Select();
             }
             return;
         }
-
-        EventSystem.current.SetSelectedGameObject(null); // Select event is not called if selecting the game object already selected
+
```
notexplosive, line 58: n   i   t:    w h i t e s p a c e

notexplosive (dismissed):
One little thing and some nits

Assets/Scripts/Menu/Menu.cs
PhoebeMitchell: Not sure I'm understanding. A line of white space is pretty standard for separating logical blocks of code, no?

Assets/Scripts/Menu/ChoiceMenu.cs
PhoebeMitchell: Separated into two error messages. `name` and `gameObject.name` output the same thing anyway so needed to change `name` to `this`

Assets/Scripts/Menu/MenuItem.cs
PhoebeMitchell: Changed to private set.

Assets/Scripts/Menu/Menu.cs
JuiceGrape: Agreed, don't have to fix this if you feel it isn't necessary. I personally like a lot of whitespaces in my code, and only group relevant things to increase readability

JuiceGrape (dismissed):
Re-approve!

lyndsiWilliams (dismissed):
Code looks great, pulled down your branch and tested it - all the choices are handled! Nice work!

lyndsiWilliams (approved):
Pulled down the branch to test it after your changes, everything works great! Code looks great too, nice job!

JuiceGrape (approved):

PR #88, opened 2021-10-03
Refactor/simplified object inventory

PhoebeMitchell:
## Summary

A refactor which aims to simplify the previous object inventory system by using a list to store items currently in the court record, instead of an ordered dictionary. Functionality should remain the same.
Fixes actors and evidence sometimes being pulled from OrderedObjectStorage instead of the dictionary of available objects (hence why an ordered dictionary was no longer required).

OrderedObjectStorage is replaced with ObjectLookup which is used for getting objects from the dictionary of available objects and storing them in a list so they can be displayed on the evidence/profiles menu.
Makes ObjectInventory class a lot more "humble", moving logic to ObjectLookup for testing.
Replaces OrderedObjectStorage tests with ObjectLookup tests.
Renames some folders and game objects to remove the word "Dictionary".

No visual changes. Tests pass.
Screenshot 2021-10-03 at 23 28 58

## Testing instructions

Run the PlayMode and EditMode tests. Play ShowItemDemo and EvidenceAddingTestScript narrative scripts and see if they function as expected. For ShowItemDemo the backgrounds are broken, but the items appearing should work.

## Additional information

- `[ ]` Changes UI


lyndsiWilliams (dismissed):
Code looks great! Pulled down your branch and ran the tests, all pass. Nice! ✅

Assets/Scripts/BGScenes/SceneController.cs
```diff
@@ -172,7 +172,7 @@ public void ShowItem(string item, itemDisplayPosition position)
             Debug.LogError($"Cannot show item, no ItemDisplay component assigned to {name}.", gameObject);
         }
 
-        Evidence evidence = _evidenceInventory.GetObjectFromAvailableObjects(item);
+        Evidence evidence = _evidenceInventory[item];
```
ViMaSter, line 175: Replacing that function call with an index operator... Just 👀👌...

Assets/Scripts/Evidence/EvidenceInventory.cs
```diff
@@ -1,22 +1,26 @@
+using System.Linq;
+using UnityEngine;
```
ViMaSter: Nit: No symbols from those namespaces are being used, so they are safe to delete.

Assets/Scripts/ObjectLookup/ObjectLookup.cs
```diff
@@ -0,0 +1,131 @@
...
+    /// <summary>
+    /// Adds an object to the list of current objects.
+    /// </summary>
+    /// <param name="objectName">The name of the object to add.</param>
+    public void AddObject(string objectName)
+    {
+        if (!IsObjectInDictionary(objectName))
+        {
+            return;
+        }
```
ViMaSter: Nit: Similarly [to the check of whether or not the object is part of the allowed items list](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/88/files#diff-017fc3eb1ef0cba834dd5cea8f07d2d03c1cad66db7bd8223de3e916435bc56aR44-R48), I think adding a `Debug.LogWarning` here would be a good idea. That way, attempts to re-add an item do not fail silently.

Assets/Scripts/ObjectLookup/ObjectLookup.cs
```diff
@@ -0,0 +1,131 @@
...
+    /// <summary>
+    /// Removes an object from the list of current objects.
+    /// </summary>
+    /// <param name="objectName">The name of the object to remove.</param>
+    public void RemoveObject(string objectName)
+    {
+        if (!IsObjectInDictionary(objectName))
+        {
+            return;
+        }
```
ViMaSter: Nit: Similarly [to the check of whether or not the object is part of the allowed items list](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/88/files#diff-017fc3eb1ef0cba834dd5cea8f07d2d03c1cad66db7bd8223de3e916435bc56aR44-R48), I think adding a `Debug.LogWarning` here would be a good idea. That way, attempts to remove an item that's not even present do not fail silently.

Assets/Scripts/ObjectLookup/ObjectLookup.cs
```diff
@@ -0,0 +1,131 @@
...
+    /// <param name="objectName">The name of the object to check for.</param>
+    /// <returns>Whether the object was in the dictionary (true) or not (false).</returns>
+    private bool IsObjectInDictionary(string objectName)
+    {
+        if (_availableObjectDictionary.ContainsKey(objectName))
+        {
+            return true;
+        }
+        
+        throw new ArgumentException($"Object {objectName} could not be found in dictionary of available objects.");
```
ViMaSter: Nit: I'd personally flip this conditional. For most of our methods, we first check and handle all unexpected or erroneous scenarios and only after we've verified an expected state, execute the main logic. That way, it's consistent across all our scripts and I personally think makes scripts more readable in general too.
Also, if we want to extend this method by introducing another check later on, the only change in Git will be an additonal `if(...) { ... }` block in future Git patches, instead of having to also review the flip of the conditional then.

Assets/Scripts/ObjectLookup/ObjectLookup.cs
```diff
@@ -0,0 +1,131 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using UnityEngine;
+using Object = UnityEngine.Object;
+
+/// <summary>
+/// Testable class that retrieves objects by name from a dictionary and stores them in a list
+/// that is used for displaying them in order.
```
ViMaSter: Nit: I think this summary could be extended:
- From my understanding, this classes main purpose is to maintain a mapping of `string`s to `T` and a related `List<T>` which entries **can only consist of items present as values in the aforementioned dictionary**. 
- It also is used to manage **all ever-available pieces of evidence and characters in a scene** and the lists represents the values of that dictionary, **that have already been introduced to the player at any given time**.

In my opinion, adding these facts to the comments, would make this class more legible.

Assets/Scripts/ObjectLookup/ObjectLookup.cs
```diff
@@ -0,0 +1,131 @@
...
+        if (newObject == null)
+        {
+            Debug.LogWarning($"Could not substitute {originalObjectName}. Provided alternate object was null.");
+            return;
+        }
+        
+        if (!IsObjectInDictionary(originalObjectName) || !IsObjectInList(_availableObjectDictionary[originalObjectName]))
+        {
+            return;
+        }
```
ViMaSter: Nit: I think these methods and the way they are called here, mixes two possible ways of handling validation at this point:
- Option 1:
  - Keep their current names
  - Instead of having them throw an exception, **only** have them return `true` or `false`
  - Split them into two separate `if`-conditionals
    - If they return `false`, print something via `Debug.LogWarning()` and return from the call to `SubstituteObject()` here
    - If they return `true`, continue execution
- Option 2:
  - Changing them to ` VerifyObjectIsInDictionary(...)` and `VerifyObjectIsInList(...)` respectively
  - Changing their return value from `bool` to `void`
  - Make them throw an exception (which contains a description on why it's being thrown), if they're unable to successfully verify
  - Instead of calling them inside a conditional (`if(Verify(...))`), simply call them normally inside the method body (`VerifyA(...);VerifyB(...);`)
  - That way, the individual `Verify(...)`-methods are responsible of interrupting the execution flow by throwing an exception, instead of `SubstituteObject()` handling it (like in Option 1)

Ideally, both methods **either** return a status **or** throw an exception if their asserted state has failed. Mixing both behaviors would only make sense to me, if not meeting one of the conditions is "worse" than not meeting the other, which doesn't appear to be the case here. (But feel free to correct me on that, if I've misunderstood something!)

Personally I could live with either option. I personally always prefer exceptions for fatal errors (like `Option 2`), but so far in our codebase, we've mostly followed the approach in `Option 1`.  
Nothing worth blocking over, but wanted to point it out still. 👀 Would be cool to hear your opinion on it!

Assets/Scripts/ObjectLookup/ObjectLookup.cs
```diff
@@ -0,0 +1,131 @@
...
+    public int AvailableObjectCount => _availableObjectDictionary.Count;
+    public int CurrentObjectCount => _currentObjectList.Count;
+
+    /// <summary>
+    /// Creates an ObjectDictionary by converting the list of available objects to the dictionary
+    /// and creating a new current object list based on the list of current objects inputted.
+    /// </summary>
+    /// <param name="availableOjectList">List to be converted to a dictionary of available objects.</param>
+    /// <param name="currentObjectList">List to be added to list of current objects on creation.</param>
+    public ObjectLookup(IEnumerable<T> availableOjectList, IEnumerable<T> currentObjectList)
```
ViMaSter: Typo: `availableOjectList` -> `availableO>b<jectList`

ViMaSter (dismissed):
image
Less code for the same functionality! Loooove to see it! 🙏

---

Couple of nits, nothing worth blocking though. Curious about your opinion on some validation code though, if you feel like looking into it!

Assets/Scripts/ObjectLookup/ObjectLookup.cs
notexplosive: 🤓 👉  OJECTION!

notexplosive (dismissed):
I love refactors!!
I love that you have tests, and they pass!!

What I would have liked to see is each commit to be it's own tiny refactor (meaning you're back to a unbroken green happy state at the end of each commit). That way instead of reviewing your whole change all at once I can just look at each refactor commit and check them off. If I can confidently say each commit is safe then I know your whole change is safe.

That being said, looks good! I agree with Vim's feedback so once you've responded to that you can get a 👍 from me

Assets/Scripts/ObjectLookup/ObjectLookup.cs
PhoebeMitchell: Updated. Let me know what you think of the new description.

Assets/Scripts/ObjectLookup/ObjectLookup.cs
PhoebeMitchell: Good catch 😆 Fixed

Assets/Scripts/ObjectLookup/ObjectLookup.cs
PhoebeMitchell: The error is thrown in the IsObjectInDictionary method itself. This isn't done for checking the list as the error messages are different for the list errors.

Assets/Scripts/ObjectLookup/ObjectLookup.cs
PhoebeMitchell: The error is thrown in the IsObjectInDictionary method itself. This isn't done for checking the list as the error messages are different for the list errors.

Assets/Scripts/ObjectLookup/ObjectLookup.cs
PhoebeMitchell: Implemented option 2.

Assets/Scripts/ObjectLookup/ObjectLookup.cs
PhoebeMitchell: No longer relevant with updated exception handling.

Assets/Scripts/ObjectLookup/ObjectLookup.cs
PhoebeMitchell: No longer relevant with updated exception handling.

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuKeyboardTests.cs
```diff
@@ -357,7 +357,7 @@ public IEnumerator EvidenceCanBeSubstitutedWithAltEvidence()
             }
 
             yield return _inputTestTools.PressForFrame(_inputTestTools.Keyboard.rightArrowKey);
-            Assert.AreEqual("Attorney's Badge", _evidenceMenu.GetComponent<Menu>().SelectedButton.GetComponent<EvidenceMenuItem>().CourtRecordObject.DisplayName);
+            Assert.AreEqual("Switch", _evidenceMenu.GetComponent<Menu>().SelectedButton.GetComponent<EvidenceMenuItem>().CourtRecordObject.DisplayName);
             SceneManager.UnloadScene("EvidenceMenu - Test Scene");
         }
     }
```
TooManyMoths, line 363: tests need fixing

Assets/Scripts/ObjectLookup/ObjectLookup.cs
ViMaSter: Way more verbose, I definitely prefer it over the original!

Tiny word repetition on the comment for `ObjectLookup<T>` though:  
`Stores all >available< objects that are >available< [...]`  
If that first `available` goes, I think it's perfect 👀👌

notexplosive (dismissed):
dew it

Assets/Scripts/ObjectLookup/ObjectLookup.cs
```diff
@@ -36,18 +39,14 @@ public ObjectLookup(IEnumerable<T> availableOjectList, IEnumerable<T> currentObj
     /// <param name="objectName">The name of the object to add.</param>
     public void AddObject(string objectName)
     {
-        if (!IsObjectInDictionary(objectName))
+        VerifyObjectInDictionary(objectName);
+        T obj = this[objectName];
+        if (IsObjectInList(obj))
```
ViMaSter: Nit: Could this be replaced with a call to `VerifyObjectInList()` as well? I see two benefits with that:
- Error handling would be consistent across the whole class (everywhere else the `Verify()`-methods are being used; thanks for implementing that, btw.!)
- Plus, any type of `.Add()`-method inside .NET `Dictionary<>`s usually throws an exception, if an item is already present. With that change, that would make us be in line with general .NET behavior here too which is always good.  
  The less we're "special" about the behavior of our code, the easier it gets to get (back) into our codebase coming from other C# projects (and thereby also applies the "write code for humans first, and compilers second"-stance I really like).

Assets/Scripts/ObjectLookup/ObjectLookup.cs
```diff
@@ -5,8 +5,11 @@
 using Object = UnityEngine.Object;
 
 /// <summary>
-/// Testable class that retrieves objects by name from a dictionary and stores them in a list
-/// that is used for displaying them in order.
+/// Stores all available objects that are available in a scene in a dictionary.
```
ViMaSter: Accidentally left my comment [on the old thread](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/88#discussion_r725558254):

---

Way more verbose, I definitely prefer it over the original!

Tiny word repetition on the comment for ObjectLookup<T> though:
Stores all >available< objects that are >available< [...]
If that first available goes, I think it's perfect 👀👌

ViMaSter (dismissed):
One "consistency"-nitpick and a tiny grammatical issue. That aside, it looks super solid now!
I can't reiterate enough how much I've enjoyed seeing this change. : D

notexplosive (dismissed):

ViMaSter (dismissed):
Ship it! 💿

ViMaSter (approved):

notexplosive (approved):

PR #93, opened 2021-10-10
Refactor Director Action Decoder to be easier to extend

notexplosive:
## Summary
This is a PROPOSAL. It might get merged if people like the changes, and I will adjust as I get feedback, but if we decide as a team we don't want this change, I'll close the PR.

### Design Goals
- Make the decoding step testable
- Adding new commands is easy
    - No more writing your own parsing code for each command
- Automated, Correct, and Clear error logging if you get a command wrong
    - If you use an enum type, the error should output all possible enum values
    - If you enter 0 when the argument wanted a one-based integer, that should be an error
- Command parameters are easily discoverable
- (Experimental!) Docs write themselves! `ActionDecoder.WriteDocsForAllCommands()` returns a markdown doc with all the commands and the args they take.

### Changes
The switch case cases in DirectorActionDecoder (now moved to ActionDecoder) looks like the following:
`ActionName.PLAY_EMOTION: PlayEmotion(actionLine.NextString("emotion name"), actionLine.NextOptionalString("target actor"));`

This means `PlayEmotion` has a very clean implementation:
```
private void PlayEmotion(string poseName, string optional_targetActor)
{
    if (!HasActorController())
        return;

    if (optional_targetActor == null)
    {
        ActorController.PlayEmotion(poseName);
    }
    else
    {
        ActorController.PlayEmotion(poseName, optional_targetActor);
    }
}
```

### Notable Findings
- SetActorPosition used to log an error explaining it's usage that's actually incorrect (the args were in the wrong order). This would have been really hard to notice and is exactly the type of problem I'm trying to fix.
- To test a command in isolation (eg: `"&SET_ACTOR_POSITION:3,Arin"` you actually need to end with a space: `"&SET_ACTOR_POSITION:3,Arin "`). This is probably a bug but I didn't fix it because this is a Refactor PR.

### Drawbacks to this approach
Although using the API (aka: creating a new command) is much easier, the underlying logic is significantly more complicated. This won't matter to most people on the team, but if you want an `OptionaBool` for instance you'll have to figure out how to add it (it's not hard to figure out, but it's learning you didn't need to do before).

If you provide _too many_ arguments there is no longer a mechanism to detect that. `&SET_ACTOR_POSTIION:3,Arin,butts,360noscope` will not error. The extra args will be ignored.

## Testing instructions
- Test all the features you know about to make sure I didn't break anything (93% confident it's safe)
- Try to enter commands that don't exist or have improper arguments and see if you get a helpful error message

## Additional information

- `[ ]` Changes UI


Assets/Tests/EditModeTests/ActionDecoderTests.cs
```diff
@@ -0,0 +1,179 @@
+using NUnit.Framework;
```
ViMaSter, line 1: # Technical design: Tests
I mean... I'm the guy who's still trying to make tests work inside GitHub Actions; this is **great work**!  
Only executed code can be truly stable code. I love it, end of story. : D

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -673,21 +82,7 @@ private bool HasAudioController()
...
-    /// <returns>Whether an evidence controller is connected</returns>
-    private bool HasEvidenceController()
-    {
-        if (_evidenceController == null)
-        {
-            Debug.LogError("No evidence controller attached to the action decoder");
-            return false;
-        }
-        return true;
+        _decoder.EvidenceController = newController;
```
ViMaSter: # [`DirectorActionDecoder.Set*Controller()` methods](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/f60e31bd58eafb2ac81f7d3e9ca2eea48178bfd8/Assets/Scripts/TextDecoder/DirectorActionDecoder.cs#L52-L95)
By making `DirectorActionDecoder._decoder` public, all `DirectorActionDecoder.Set*Controller()` calls could be replaced with direct `DirectorActionDecoder._decoder.*Controller = value;` assignments. All `Set*Controller()` methods could be removed for - what looks to me like - the same result. (Unless there's something Unity-specific we'd lose that I'm missing right now.)

If we need some validation logic when setting a new controller, we can always add it by replacing [the default property `set;`er](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/f60e31bd58eafb2ac81f7d3e9ca2eea48178bfd8/Assets/Scripts/TextDecoder/ActionDecoder.cs#L76-L80) with an expression containing whatever validation we need.

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -1,624 +1,61 @@
...
+    /// <param name="line">The full line in the script containing the action and parameters</param>
+    public void OnNewActionLine(string line)
     {
-        if (!HasSceneController())
-            return;
-
-        bool shouldShow;
-        if (bool.TryParse(showActor, out shouldShow))
+        var actionLine = new ActionLine(line);
+        try
```
ViMaSter, line 25: # Multiple `catch` es
This could be simplified:  
Instead of having `catch` blocks for each exceptions, the generated message could be generated by overloading the `ToString()` method of the exception. Then only one type of exception needs to be handled here:
### Original:
```cs
///// Definition /////
// for each exception-type...
public class UnknownCommandException : Exception
{
    public string CommandName { get; private set; }

    public UnknownCommandException(string commandName)
    {
        CommandName = commandName;
    }
}

///// Usage /////
try
{
    _decoder.OnNewActionLine(actionLine);
}
// for each exception-type...
catch (UnknownCommandException e)
{
    Debug.Log($"Unknown action: {e.CommandName}");
}
```

### Suggestion
```cs
///// Definition /////
// for each exception-type...
public class UnknownCommandException : Exception
{
    public string CommandName { get; private set; }

    public UnknownCommandException(string commandName)
    {
        CommandName = commandName;
    }

    // ADD THIS
    public override string ToString()
    {
        return $"Unknown action: {e.CommandName}"; // move the human readable string here
    }
    // ADD THIS [END]
}

///// Usage /////
var actionLine = new ActionLine(line);
try
{
    _decoder.OnNewActionLine(actionLine);
}
// This now handles all exceptions in one go:
catch (Exception e)
{
    Debug.Log(e.ToString());
}
```

**Note:** If we only want to handle custom exceptions, create a base Exception and handle that instead:
### Only-custom-exception handler suggestion
```cs
///// Definition /////
// ADD THIS
public abstract class CommandException : Exception // abstract to prevent direct initialization
{
    public string CommandName { get; private set; }

    public CommandException(string commandName)
    {
        CommandName = commandName;
    }
}
// ADD THIS [END]

// for each exception-type...
public class UnknownCommandException : CommandException
{
    public string CommandName { get; private set; }

    public UnknownCommandException(string commandName)
    {
        CommandName = commandName;
    }

    // ADD THIS
    public override string ToString()
    {
        return base.ToString();
    }
    // ADD THIS [END]
}

///// Usage /////
var actionLine = new ActionLine(line);
try
{
    _decoder.OnNewActionLine(actionLine);
}
// This now handles all our custom exceptions:
catch (CommandException e)
{
    Debug.Log(e.ToString());
}
```

---

Also a super tiny nit: `Debug.Log()` should probably be a `Debug.LogError()`

Assets/Tests/EditModeTests/ActionDecoderTests.cs
```diff
@@ -0,0 +1,179 @@
...
+            public void StartTalking()
+            {
+                this.talking = true;
+            }
+
+            public void StopTalking()
+            {
+                this.talking = false;
+            }
+        }
```
ViMaSter: This is a great use of our interfaces! One thing to keep in mind: If we end up with more than 3 "mocks" for our tests, I would recommend looking into a [library like Moq](https://stackoverflow.com/a/49740489/1088653) to automate creating mocks.  

Moq can create mocks of interfaces with just using `new Mock<IActorController>()` instead of having to implement them ourself. These `Mock<>` instances can then be used to fake return values or keep track of how often they have been called.

Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -0,0 +1,672 @@
...
+            case ActionName.HIDE_TEXTBOX: HideTextbox(); break;
+            //Do nothing
+            case ActionName.WAIT_FOR_INPUT: break;
+            //Default
+            // If we got here then the action exists in the ActionName enum but doesn't have a case in the switch hooked up to it.
+            default: throw new UnknownCommandException(actionLine.Action.ToString());
+        }
+    }
+
+    #region DialogStuff
```
ViMaSter: ## `Has*Controller()` methods
To me, all `Has*Controller()` methods seem superfluous, as all they do, is return an exception that informs the user a controller is missing.  
If we simply use the references to our controllers (i.e. `SceneController.FadeIn(...)`) and `SceneController` turns out to be set to `null`, C# generates a `NullReferenceException` which is identical, requires no custom exception and behaves completely identical.  

This actually came up a few PRs ago too. [Here's the conversation on GitHub](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/83#discussion_r719074650) with a few more arguments related to that as well as [a thread on Discord](https://discord.com/channels/274342839041916928/622212006820315136/893290005844353094).

This PR seemed like a good place to bring this up again, as my stance on it hasn't really changed. 🙈   
I really hope this is something we can simplify.

Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -0,0 +1,672 @@
...
+    OVERALL_SPEED,
+    PUNCTUATION_SPEED,
+    CLEAR_SPEED,
+    DISABLE_SKIPPING,
+    AUTOSKIP,
+    CONTINUE_DIALOG,
+    APPEAR_INSTANTLY,
+    HIDE_TEXTBOX,
+    WAIT_FOR_INPUT
+}
```
ViMaSter: # Technical design: Redundant action names and parameters
The fact that this is being addressed is **a-mazing**! As you mention in your PR, with the design currently on `develop`, there's a lot of room for error [which we've already fell into](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/93#:~:text=SetActorPosition%20used%20to%20log%20an%20error%20explaining%20it%27s%20usage%20that%27s%20actually%20incorrect%20(the%20args%20were%20in%20the%20wrong%20order).%20This%20would%20have%20been%20really%20hard%20to%20notice%20and%20is%20exactly%20the%20type%20of%20problem%20I%27m%20trying%20to%20fix.) (and let's be honest: probably won't be the only and/or last time).
In my mind adjusting this class to reduce code while achieving the same results as we do right now is an important part of this refactor, as it will make this class a lot easier to read and extend.
  
I've noticed, that, even with the changes in this PR, for each action we still have...
- the action name in...
  - a value inside the `ActionName` enum (`ActionName.>SET_ACTOR_POSITION<`)
  - a switch case (`case ActionName.>SET_ACTOR_POSITION<: ...`)
  - a wrapper method (`>SetActorPosition<`)
- and for each parameter:
  - the type and name in the switch case (`actionLine.NextOneBased>Int<(>"slot index"<)`)
  - the type and name of the parameter inside the wrapper method (`SetActorPosition(>int< >oneBasedSlotIndex<)`)

Proposal:
If we have a method with the following syntax for each action:
```cs
/// <summary>
/// Sets an actor inside a slot in the scene, if the active bg-scene has support for slots.
/// </summary>
/// <param name="oneBasedSlotIndex">Target slot, 1 based</param>
/// <param name="actorName">Target actor</param>
private void SET_ACTOR_POSITION(int oneBasedSlotIndex, string? actorName)
{ 
    ActorController.AssignActorToSlot(actorName, oneBasedSlotIndex); // the method we actually want to expose via Inky
    OnActionDone?.Invoke(); // invoking the "on action done" event
}
```
[everything else (from parsing the parameters, generating error messages, to executing the wrapped method) can be generated during runtime](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/compile-time-constant-actions/Assets/Scripts/TextDecoder/DirectorActionDecoder.cs#L126-L180).  
This would be super useful, as it will fully eliminate the need for both the `switch` statement and `enum`; reducing a lot of lines of code and (more importantly) room for error when updating or adding actions and their parameters. (As all that would be needed to generate an action, is a method of this syntax and nothing else.)
 
**Note:** Optional parameter handling is missing from that linked code example. That can easily be added however, by checking if `methodParameter` in line 145 is of type `System.Nullable<>`.

ViMaSter (dismissed):
So uh... this is pretty sweet. 👀

To start off with my simple stance on whether or not we should follow through with this proposal: Yes. Definitely. Thank you for starting this. 🙏  
I personally really think that a refactor of `DirectorActionDecoder` is needed. A few more tweaks are needed in my opinion, but redesigning the decoder is a good thing.

As is evident by the fact that I already tried to create a refactor myself, I'm not too happy with the current state of this class. It's reassuring to see I'm not alone on it. I also **really** appreciate, that this is an actionable PR with suggestions and not just an abstract design discussion pointing out problems.  
Talking about code, feels like dancing about architecture. 😅


To keep the discussions somewhat organized, I'll comment **on code, even if it's an opinion on a broad design design**.  
I'll mark them accordingly so it's apparent whether or not I'm talking about general design decisions or that very specific block of code.

(Just marking this review as `requesting changes`, just so it doesn't accidentally get merged in. If we do talk about all points and end up on agreeing to not implement any further changes, I'll approve of course.)

Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -0,0 +1,672 @@
...
+            OnNewActionLine(docAction);
+        }
+        catch (DecoderMissingComponentException)
+        {
+        }
+
+        return docAction.Description();
+    }
+
+    public string WriteDocsForAllCommands()
```
ViMaSter: # Automated doc generation
This is actually a really great idea too. Two suggestions for the process you put in place:
1. **Generally, I wouldn't let classes generate info about themselves. Instead, I would use an external class to extract it**  
  Obtaining documentation and generating a human readable format isn't exactly game code and could run in-editor only; hence I'd move that work to an editor-only assembly.  
  Additionally, I think making it part of an editor assembly has a few additional advantages:
    - it simply can't be run in-game: a good interface is easy to use but also **hard to use wrong**. If we don't need generation in-game, just disallow access to it
    - [it can be ran via GitHub Actions](https://docs.unity3d.com/Manual/CommandLineArguments.html#:~:text=-executeMethod%20%3CClassName.MethodName,defined%20as%20static.): this way, we could automatically update our wiki during the build process (the biggest plus imo)
    - it's not part of the compiled game's executable and saves a few bytes when downloading the game (though this is negligible)

  **BUT, more importantly:**

2. **We already have human-readable documentation: [the xmldoc-comments of methods](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/ecf527994d555c091b150cef2de359c233c3ada7/Assets/Scripts/Evidence/EvidenceController.cs#L54-L58)**  
  Instead of generating documentation based on executed code, I would recommend looking into code generation. Scan the code for comments and just expose these instead.

Assets/Scripts/TextDecoder/ActionLine.cs
```diff
@@ -0,0 +1,232 @@
...
+}
+
+public class ActionLine : IActionLine
+{
+    private const char ACTION_SIDE_SEPARATOR = ':';
+    private const char ACTION_PARAMETER_SEPARATOR = ',';
+
+    private readonly string fullParametersString;
+    private readonly string[] splitParameters;
+    private int parameterIndex;
```
PhoebeMitchell: Private fields should start with _

Assets/Scripts/TextDecoder/ActionLine.cs
```diff
@@ -0,0 +1,232 @@
...
+    string NextString(string input);
+    string NextOptionalString(string input);
+    int NextInt(string input);
+    float NextFloat(string input);
+    int NextOneBasedInt(string input);
+    bool NextBool(string input);
+    T NextEnumValue<T>(string input) where T : struct, IConvertible;
+}
+
+public class DocActionLine : IActionLine
```
PhoebeMitchell: Methods are perhaps self explanatory, but summaries for classes would be nice.

Assets/Scripts/TextDecoder/ActionLine.cs
```diff
@@ -0,0 +1,232 @@
...
+{
+    private List<string> args = new List<string>();
```
PhoebeMitchell: Private field should start with _

Assets/Tests/EditModeTests/ActionDecoderTests.cs
```diff
@@ -0,0 +1,179 @@
...
+        /// <summary>
+        /// Clumsy implementation of IActorController for tests
+        /// </summary>
+        private class FakeActorController : IActorController
+        {
+            private bool talking;
+            private SpeakingType speakingType;
+            private string pose = "";
+            private string activeSpeaker = "";
+            private string activeActor = "";
```
PhoebeMitchell: Private fields should start with _

Assets/Tests/EditModeTests/ActionDecoderTests.cs
```diff
@@ -0,0 +1,179 @@
...
+            public string emotion = "";
+            public string emotionActiveActor;
+            public readonly Dictionary<int, string> actorSlots = new Dictionary<int, string>();
+
+            public Action onAnimationDone;
```
PhoebeMitchell: Public fields should be properties and start with a capital.

Assets/Scripts/Interfaces/ISceneController.cs
```diff
@@ -1,6 +1,6 @@
 using UnityEngine;
 
-public enum itemDisplayPosition
+public enum ItemDisplayPosition
```
PhoebeMitchell, line 3: Finally 😆 👍

Assets/Tests/EditModeTests/ActionDecoderTests.cs
notexplosive: I've never written a C# mock before (I actually hate mocks normally but this seemed like an OK use for them). Thanks for the suggestion! I'll look into it.

Assets/Tests/EditModeTests/ActionDecoderTests.cs
PhoebeMitchell: Could this be a unity event to make it consistent with the rest of the project?
Otherwise could it be an "event Action" to make it thread safe?
Also the event on ActorController says "complete". It should probably be "done" to be consistent with other events like OnActionDone on DirectorActionDecoder.

Assets/Tests/EditModeTests/ActionDecoderTests.cs
```diff
@@ -0,0 +1,179 @@
...
+
+            public void AssignActorToSlot(string actor, int oneBasedSlotIndex)
+            {
+                this.actorSlots[oneBasedSlotIndex] = actor;
```
PhoebeMitchell: For all of these methods, is explicitly stating "this" necessary?

Assets/Scripts/TextDecoder/ActionDecoder.cs
notexplosive: You're totally right that there's still a ton of duplication. Your suggestion uses some C# Dark Magic that I'm not yet familiar with. I'll have to dig into it more.

Also as an aside, `Nullable` only works on structs as far as I know, you can't make a `Nullable<string>` unless I misunderstood my error messages.

Assets/Scripts/TextDecoder/ActionDecoder.cs
notexplosive: This sounds reasonable to me, the automated doc generation was a last minute like "hey I bet I could do this cool trick now that I have all these other pieces," so I'm happy to replace it with something better.

Assets/Tests/EditModeTests/ActionDecoderTests.cs
```diff
@@ -0,0 +1,179 @@
...
+            {
+                decoder.OnNewActionLine(new ActionLine("&SPEAK:Arin:Dan:Ross"));
+            });
+        }
+
+        [Test]
+        public void UseValidSyntax()
+        {
+            var decoder = new ActionDecoder();
+            decoder.ActorController = new FakeActorController();
```
PhoebeMitchell: These two lines are repeated a lot so maybe put them in a [SetUp] method and have them set private fields.

Assets/Scripts/TextDecoder/ActionLine.cs
```diff
@@ -0,0 +1,232 @@
...
+
+public class UnableToParseException : System.Exception
+{
+    public UnableToParseException(string typeName, string parameterName, string token)
+        : base($"Failed to parse {parameterName}: cannot parse `{token}` as {typeName}")
+    {
+    }
+}
+
+public class NotEnoughParametersException : System.Exception
```
PhoebeMitchell: System in System.Exception is redundant as we already have using System.

Assets/Scripts/TextDecoder/ActionLine.cs
```diff
@@ -0,0 +1,232 @@
...
+        args.Add(argName + ":bool");
+        return true;
+    }
+
+    public T NextEnumValue<T>(string argName) where T : struct, IConvertible
+    {
+        args.Add($"{argName}:[{string.Join(",", typeof(T).GetEnumNames())}]");
+
+        // I apologize for this line
+        return (typeof(T).GetEnumValues() as T[])[0];
```
PhoebeMitchell: You use "as" implying it's possible for this to be null if the conversion fails, but then use [0] which requires that it
 must not be null.

Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -0,0 +1,672 @@
...
+            return;
+
+        AppearingDialogueController.ClearAllWaiters();
+    }
+
+    ///<summary>
+    ///Toggles skipping on or off
+    ///</summary>
+    ///<param name = "disable">Should the text skipping be disabled or not</param>
+    private void DisableTextSkipping(bool value)
```
PhoebeMitchell: `Disable` should be updated to `value`

Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -0,0 +1,672 @@
...
+
+    public IActorController ActorController { get; set; }
+    public ISceneController SceneController { get; set; }
+    public IAudioController AudioController { get; set; }
+    public IEvidenceController EvidenceController { get; set; }
+    public IAppearingDialogueController AppearingDialogueController { get; set; } = null;
+
+    public ActionDecoder()
+    {
+    }
```
PhoebeMitchell: Empty constructor can be removed

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -1,624 +1,61 @@
...
-
-    #region ControllerStuff
     /// <summary>
     /// Attach a new IActorController to the decoder
     /// </summary>
     /// <param name="newController">New action controller to be added</param>
     public void SetActorController(IActorController newController)
     {
-        _actorController = newController;
-    }
```
PhoebeMitchell: As this class is being "humblified" is this method of setting the controllers still necessary? Could this class not just have serialized fields for each controller, which it passes to the decoder??

Assets/Scripts/TextDecoder/ActionLine.cs
```diff
@@ -0,0 +1,232 @@
...
+            throw new UnknownCommandException(actionNameAsString);
+        }
+        fullParametersString = (actionAndParam.Length == 2) ? actionAndParam[1] : "";
+        this.splitParameters = fullParametersString.Split(ACTION_PARAMETER_SEPARATOR);
+    }
+
+    public string[] Parameters()
+    {
+        return this.splitParameters;
+    }
```
PhoebeMitchell: I prefer using properties for things that just return a value without any hidden logic.

PhoebeMitchell (commented):
Simplifying DirectorActionDecoder and making actions easier to add is definitely something I approve of. I've got a lot of nits about code style which I've commented on, but I understand this is just a draft.

One problem I have is although DirectorActionDecoder has been shortened, a lot of its logic has just been extracted into its own 600 lines monolith file, ActionDecoder (albeit with tests! 👍 ). A lot of it is just methods that call other methods so maybe the correct methods could just be directly called in the switch statement.

Great work! Look forward to seeing something like this implemented in the project.

PhoebeMitchell (commented):
I forgot to mention, the first line of text when playing appears instantly now which it probably shouldn't do.
textappear

Assets/Tests/EditModeTests/ActionDecoderTests.cs
notexplosive: I deliberately didn't use a UnityAction because I have this pipe dream that ActionDecoder can be extracted into a separate parser application completely disconnected from Unity (and therefore won't use any unity namespaces)... Not sure if this is actually ever gonna happen.

Will address the rest of your feedback though

Assets/Scripts/TextDecoder/ActionLine.cs
notexplosive: I got a compiler warning about Properties returning static arrays. Apparently Microsoft thinks that's a bad practice? Otherwise yeah I agree.

Assets/Tests/EditModeTests/ActionDecoderTests.cs
ViMaSter: > [...] I actually hate mocks normally but this seemed like an OK use for them [...]

Now I'm kinda curious...👀  
Any particular experience or reason for that stance? : D

Assets/Scripts/TextDecoder/ActionDecoder.cs
ViMaSter: > You're totally right that there's still a ton of duplication. Your suggestion uses some C# Dark Magic that I'm not yet familiar with. I'll have to dig into it more.

Fair enough! It is a pretty complex block of code. If you want to, we can hop on a VC if you don't feel like staring at it for hours. : D

The gist: C# has a feature called [Reflection](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/reflection). It's essentially [run-time type information](https://en.wikipedia.org/wiki/Run-time_type_information), meaning that any game or app written in C#, has information about all it's types available **while it's running** instead of just while it's compiling. We can use this to see if we have a method with parameters, that match what we have in the action line of the Ink script.

What I'm doing in that code block, is pretty straight forward:
1. based on the character string of the action name we have in the Ink script (in this example `SET_ACTOR_POSITION`), look for a method inside the `DirectorActionDecoder` with that exact name
2. If it doesn't exist, `GetMethod()` is null and I abort
3. With the `MethodInfo` returned by `GetMethod()`, I can also check which parameters it needs and then iterate over every parameter from the Ink script and see if there's a `Parser<>` instance of that type. Let's say we need an `int` for `oneBasedSlotIndex`:
  3.1.  I grab the `Parser<int>` class.
  3.2.  Then I take the string-representation of the first parameter from inside the Ink script, pass it to `Parser<int>.Parse()` and it either...
    - returns a string with an error message which I pass to the console and abort
    - returns an empty string and the `out` parameter is set with an actual converted `int` I can use
  
    3.3. Repeat this for every parameter until the entire Ink line is parsed
4. Then I just invoke the method with all the parameters we've extracted

Again, feel free to let me know, if you want to go over this on Discord or if anything's unclear!

---

> Also as an aside, Nullable only works on structs as far as I know, you can't make a Nullable<string> unless I misunderstood my error messages.

You're absolutely right. I was mixing up `Nullable` and "optional parameters".  
The better syntax would be something along the lines of
```cs
private void SET_ACTOR_POSITION(int oneBasedSlotIndex, string actorName = null) { ... }
private void SET_ACTOR_POSITION(int oneBasedSlotIndex, string actorName = "") { ... }
private void SET_ACTOR_POSITION(int oneBasedSlotIndex, string actorName = default(string)) { ... }
```
This has multiple advantages over my original `Nullable` idea:
1. it works with every type
2. we'd have to explicitly define the fallback value to use if supplied by Ink script
  (because like you mentioned, for things like `struct Foo { }` or `int`, a value of `null` wouldn't be acceptable, unless we'd change it to `Foo?` or `int?`)
3. it actually makes `actorName` an >optional parameter<, instead of just >accepting nulls<; which is a different requirement we don't even need here

This can also easily be checked for, by checking if `methodParameter.HasDefaultValue` is set to `true` or `false` respectively.

Good catch!

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
ViMaSter: It was probably invisible thanks to my sea of words, [but I opened a thread for a change along those lines as well](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/93#discussion_r725621747). Just linking it so we can mark one of these conversations as resolved and only have one thread for it.

Always great to see an independent reviewer pointing out similar ideas though. : D

Assets/Scripts/TextDecoder/ActionLine.cs
ViMaSter: Wild guess: I think you're talking about [CA1819](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1819?view=vs-2019)? If so, I personally think it's actually a valid point from Microsoft. 👀 

Basically, a "get"-parameter in C# usually implies that the underlying value is **copied** and **the underlying field is never modified just by `get`ting it**. 
I.e.: `public CoolClass Foo { get => _foo; set => _foo = value; }` would mean, that we get a new copy of that `CoolClass`-object and no matter what we do to the returned value, `_foo` never gets modified UNLESS we reassign it using `Foo = new CoolClass()`.

The only way to guarantee that with an array, is by **copying the entire array**. This can get expensive, if it's done per frame for example.  

Marking it as `readonly` doesn't help here either, as `readonly` only prevents **re-assignments** (for example, `.Parameters = new string[0];` wouldn't compile), but an array could simply be `.Clear()`ed and new values be `.Add()`ed to **achieve the same result** which can be done **even when the collection is marked as `readonly`**.

MS suggests one of two ways:
1. If we need a copy of the array we can modify further (thereby leaving the underlying static array intact), return `this.splitParameters.Clone()` inside a method
2. If we only ever need to read values, we can keep it a property, but make it be of type `ReadOnlyCollection<string>` and then return `this.splitParameters`

Hope that makes sense? The MSDN usually has pretty solid examples for their compiler warnings (check out the link at the top), but they can get pretty technobabbly. : S

bas-mentjot:
Won't comment on the code as I'm supposed to not be a part of the project, but I wanted to say that this:
image
is because in my implementation the function that separates the arguments from the action expects a newline at the end. 
IMO this is bad practice, as we can't expect the function to always be called from the same place, and thus we can't expect the provided string to always have that \n at the end. I'd recommend moving that to the dialogueController. Makes it easier to not mess up testing as well :)

JuiceGrape:
Won't comment on the code as I'm supposed to not be a part of the project, but I wanted to say that this:
image
is because in my implementation the function that separates the arguments from the action expects a newline at the end. 
IMO this is bad practice, as we can't expect the function to always be called from the same place, and thus we can't expect the provided string to always have that \n at the end. I'd recommend moving that to the dialogueController. Makes it easier to not mess up testing as well :)
If you've already caught it, feel free to disregard this message.

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
notexplosive: If I'm reading this right I don't think I can do this with `ToString()` because most of the log statements compose a string based on local variables here. I could potentially create a parent class with a function that takes a `String line` and an `ActionName`, does that align with your suggestion?

Assets/Tests/EditModeTests/ActionDecoderTests.cs
notexplosive: Sure, I'll do my best to articulate it.

Consider the following code (super contrived example):

```cs
public class SomeClass {
  // We have some private field that holds some complicated object Foo
  private Foo _foo;
  
  // We do some stuff involving Foo
  public void myFunction() {
     if(_foo.SomeComplicatedCondition()){
        var animal = _foo.GetAnimal();
        /* some other code (A) */
     }
     /* some more code (B) */
  }
}
```

Now consider I write a test for myFunction where I mock a Foo. I then make sure that when I call `_foo.GetAnimal` I'll get a Cat, and when I call `SomeComplicatedCondition` it always returns true. I also write into my test that I assert that `GetAnimal` will be called, the test will fail otherwise. These are (as far as I understand) the two main use cases for mocks: (1) fake the return for this thing, and (2) assert this function will be called. It also has bonus use case (3) fake this object so I can write a test for it.

(1) Assumes that the system works a certain way, what if `GetAnimal()` never actually returns a Cat when you run this code for real? Now you're testing a scenario that never happens. What if it _used to_ return a Cat but doesn't anymore? The test became less effective over time and you didn't notice.

(2) Asserting a function will be called feels like an implementation detail to me, unless it's firing an event, which you can hook into without using a mock.

(3) This is I think the _real_ reason people use Mocks. You can't have a SomeClass without a Foo, and Foo is too complicated to build in a test case. That's a code smell, you should address the underlying design problem. My first thougt is we should extract code blocks `(A)` and `(B)` into their own methods and test those, and assuming Foo has its own tests then we might be able to get away with not writing a test for `myFunction` because it's a composition of things that area already tested.

Assets/Scripts/TextDecoder/ActionDecoder.cs
notexplosive: I love deleting code, let's do it

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
notexplosive: Done and done

Assets/Scripts/TextDecoder/ActionDecoder.cs
notexplosive: For now I'm just removing my auto-documentation code. We can consider automatic documentation as a future PR.

Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
notexplosive: Feel free to take this on and merge it in.

Assets/Scripts/TextDecoder/ActionDecoder.cs
notexplosive: Feel free to take this on and merge it into this branch

Assets/Tests/EditModeTests/ActionDecoderTests.cs
ViMaSter: Whoa, thanks for the detailed explanation! That's exactly what I was asking about and I totally see where you're coming from. I've definitely simply had the luxury of never having to consider them - so thanks for pointing them out!

Most projects I wrote tests for in the past, I either mock an interface which - when run in production - simply interfaces with library code or an external API (at which point I do make the assumption, that the underlying component I'm replacing with the mock, is covered on their own tests<sup>*1</sup>) at which point I'd expect breaking changes to be indicated by changing version numbers.

I'd argue that there's scenarios where, depending on the size of code block `(A)` and `(B)`, extracting them into another method and testing them isolated might not make sense, but I definitely agree with the general idea of trying to cover components on their own and thereby implicitly making "glue code" as error resilient as possible, instead of testing the glue code itself.

<sup>*1</sup>: As an aside: The amount of "yeah, that's not covered by automated testing here"-emails I've received from professional payment providers trying to integrate edge-cases is terrifying. 👀

ViMaSter:
Just FYI: I've pushed a merge commit from `develop` so that this PR is up-to-date.

ViMaSter:
Assigning this to me, just for visibility as I've taken this feature over after explosive's suggestion.

Also moved all changes from my sub-branch of this to this branch instead and with that **the change is in a reviewable state now**.  
Feel free to check if all your scenes still work with the changes and the code both makes sense to you and is easier to work with.

Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -0,0 +1,367 @@
...
+    public IActorController ActorController { get; set; }
+    public ISceneController SceneController { get; set; }
+    public IAudioController AudioController { get; set; }
+    public IEvidenceController EvidenceController { get; set; }
+    public IAppearingDialogueController AppearingDialogueController { get; set; }
+
+    /// <summary>
+    /// Parse action lines inside .ink-files
+    /// </summary>
+    /// <param name="lines">Action line</param>
```
PhoebeMitchell: Param name doesn't match parameter name actionLine (and the description 'Action Line' is also a bit useless)

Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -0,0 +1,367 @@
...
+    /// <remarks>
+    ///     Writers are able to call methods inside .ink files. This is done by using the following syntax:
+    ///     <code>
+    ///     &amp;{methodName}:{parameter1},{parameter2},...
+    ///     </code>
+    ///     This method is responsible for:
+    ///         1. Finding a method inside this class, matching `methodName`
+    ///         2. Verifying the amount of parameters matches the amount of parameters needed in the method
+    ///         3. Attempting to parse each parameter into the correct type using <see cref="Parser<T>"/> of the type
```
PhoebeMitchell: Use escape characters \&lt; and \&gt; for < and >

Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -0,0 +1,367 @@
...
+    {
+        SceneController.FadeIn(timeInSeconds);
+    }
+
+    private void FADE_OUT(float timeInSeconds)
+    {
+        SceneController.FadeOut(timeInSeconds);
+    }
+
+    private void SHAKE_SCREEN(float intensity, float duration, bool isBlocking)
```
PhoebeMitchell: isBlocking should be optional. Default is false

Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -0,0 +1,367 @@
...
+        }
+
+        // Call the method
+        method.Invoke(this, parsedMethodParameters.ToArray());
+    }
+
+    // ReSharper disable InconsistentNaming
+    // ReSharper disable UnusedMember.Local
+#pragma warning disable IDE0051 // Remove unused private members
+    #region DialogStuff
```
PhoebeMitchell: Change DialogStuff to AppearingDialogueController to be consistent with naming of the other regions

Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -0,0 +1,367 @@
...
+    private void DIALOG_SPEED(float seconds)
+    {
+        AppearingDialogueController.SetTimerValue(WaiterType.Dialog, seconds);
+    }
+    private void OVERALL_SPEED(float seconds)
```
PhoebeMitchell, line 129: Pretty big nit but would be nice to have the same line spacing between every method. Some have a single line, some have no spacing, and some have double spacing

Assets/Scripts/TextDecoder/Parser/Parser.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    /// <summary>
+    ///     Base class responsible for parsing parameters from .ink files
+    /// </summary>
+    /// <see cref="ActionDecoder.OnNewActionLine" />
+    /// <typeparam name="T">The type handled in a parser derivative</typeparam>
+    public abstract class Parser<T>
+    {
+        /// <summary>
+        ///     Converts .ink-file representation of a parameter to a usable object
```
PhoebeMitchell: Naming of .ink file and .ink-file is inconsistent. I think it should be .ink file. Also on line 17 of ActionDecoder.

PhoebeMitchell (dismissed):
I have some lovely nits for you

Assets/Scripts/TextDecoder/ActionDecoder.cs
ViMaSter: Good catch! I've updated it so that regions have one line of margin, but no padding. Let me know if this works for you!
```cs
/// ...
private void FOO()
{
}

#region SceneController
private void FADE_IN(float timeInSeconds)
{
    SceneController.FadeIn(timeInSeconds);
}

private void FADE_OUT(float timeInSeconds)
{
    SceneController.FadeOut(timeInSeconds);
}
#endregion

private void BAR()
{
}
/// ...
```

If this syntax is okay, I'd also add it to the Code Style Guide, but I don't have write access to the Google Doc.

I still like the idea of [having it be part of the Wiki](https://github.com/ViMaSter/GG-JointJustice-Unity/wiki/Code-Style-Guide) and I think I'd also have permissions to add it there. 👀

Assets/Scripts/TextDecoder/Parser/Parser.cs
ViMaSter: Ah, superfluous hyphenation. That's a germanism, sorry for that. : D  
Makes sense and it's fixed in my latest PR.

(On that note, shoutout to "[Donaudampfschifffahrtsgesellschaftskapitänsmütze](https://translate.google.com/?sl=de&tl=en&text=Donaudampfschifffahrtsgesellschaftskapit%C3%A4nsm%C3%BCtze&op=translate)" which - for some reason - doesn't need hyphens though. Real German. Actual word.)

Assets/Scripts/TextDecoder/ActionDecoder.cs
ViMaSter: I... Oh dear... Yup, that's pretty useless. 🙃
Made it a bit more verbose, lemme know if that's enough of an improvement or if there's still something you'd like to see here!

PhoebeMitchell (approved):
Please accept this green tick

Assets/Scripts/TextDecoder/Parser/ItemDisplayPositionParser.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace TextDecoder.Parser
+{
+    public class ItemDisplayPositionParser : Parser<ItemDisplayPosition>
+    {
+        public override string Parse(string input, out ItemDisplayPosition output)
+        {
+            if (!Enum.TryParse(input, out output))
+            {
+                return $"Cannot convert '{input}' into an {typeof(ItemDisplayPosition)} (valid values include: '{string.Join(", ", Enum.GetValues(typeof(ItemDisplayPosition)))}')";
```
notexplosive, line 11: (not blocking comment, this is gold plating) I wonder if we could genericize this into any enum type, and just have an EnumParser<T> that takes an enum type?

notexplosive (commented):
Thanks for carrying this monster PR through. Sorry to the rest of the team for clogging up our progress with this massive change 😅

There are a few variable name nits here and there but rather than holding this PR any longer I'm just gonna say let's get this thing through the pipeline.

Because I technically own this PR I can't give approval. But uh...

**[APPROVED]**

lyndsiWilliams (approved):
Pulled your branch down locally, tests are passing and everything looks good!

Assets/Scripts/TextDecoder/Parser/ItemDisplayPositionParser.cs
ViMaSter: This _is_ actually possible and I was thinking about it, then thought "Meh, probably not worth it.", but since you now pointed it out too, I'll address it! : D

In the interest of this thing getting merged I'll make that change a new, tiny PR instead though. 👀

PR #96, opened 2021-10-19
71 add hold it / objection / take that actions with support for random character specific variations

### LINKED ISSUE #71: Add hold it / objection / take that actions with support for random character specific variations
## Description

During the game, a lot of characters have a "hold it / objection / take that" moment. These moments should be triggered by the actions:
- TAKE_THAT:Actor
- HOLD_IT:Actor
- OBJECTION:Actor

Calling these will cause the corresponding sound to play (specific per actor) and the corresponding image (with animation, shaking a little before standing still, look at lua build for refrence) to show up. The system then waits for a short amount of time (about half the time the sound takes), and continue naturally.

However, each actor should have a list of random variations these shout outs can become. Any shout out can become one of the random variations. The chance a shout becomes a random one should be adjustable per unity scene.

This story only focuses on laying the ground work for this functionality. The actual random shout outs (including new graphics) will be added later

Also, to make sure the writers can call out any of these shouts specifically, the action SHOUTOUT:Actor,string should be added. This will shoutout a specific shout out of the provided actor. 
For example, if arin had "my dick's falling off" as a shoutout, it would look something like:

SHOUTOUT:Arin,Dick_falling_off

These should be based on the name of the provided graphic, as the SFX file will probably have the name of the actor in the title as well.

## Dev Guide

Keep in mind that we will have several image objects that are going to be animated the same (all shoutout animations are the same but with a different image). Maybe we should animate it in code so we can just put in a different image? Or maybe an animation which just changes position?

## Acceptance Criteria

This story will be finished when the aforementioned code is implemented and testable in a unity scene (the random chance should be adjustable to show that it will always pick a random option at 100%)

PhoebeMitchell:
## Summary

Implements OBJECTION, HOLD_IT, and TAKE_THAT actions.

Kapture 2021-10-19 at 02 17 05

## Testing instructions

Run the "ShoutTest" narrative script and check shouts play correctly. Change Arin's random shouts adding audio clips and sprites to the "Shout Variants" array on the Arin scriptable object.

## Additional information

- `[ ]` Changes UI
  
closes #71


Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -95,6 +95,15 @@ public void PlaySFX(string soundEffectName)
...
     }
+    
+    /// <summary>
+    /// Overload for PlaySFX that plays sound using an audio clip rather than a name.
+    /// </summary>
+    /// <param name="soundEffect">The audio clip to play.</param>
+    public void PlaySFX(AudioClip soundEffect)
+    {
+        _sfxAudioSource.PlayOneShot(soundEffect);
+    }
```
ViMaSter: This method is great for a very specific reason: Having the AudioController only accept AudioClips (and not be concerned with where they're fetched, like the `PlaySFX(string)`-overload does) feels **way better** in my opinion.

I'd suggest moving the code inside `PlaySFX(string)` to the `DirectorActionDecoder:PlaySFX()` method (so that the job of resolving the filename inside the Ink to an `AudioClip` is part of the decoder and we can call this new `PlaySFX(AudioClip)`-overload instead and delete the `PlaySFX(string)` one.  
Let me know if you have any opinions on it; I'd definitely add it myself too, but just need to know if you'd be okay with me appending it to this PR or creating a separate one for it! (But if you agree and want to, feel free to do it yourself. : D)

Assets/Scripts/Animation/Shout.cs
```diff
@@ -0,0 +1,107 @@
+using System;
+using System.Linq;
+using UnityEngine;
+using Random = UnityEngine.Random;
+
+[RequireComponent(typeof(SpriteRenderer)), RequireComponent(typeof(ScreenShaker))]
+public class Shout : MonoBehaviour
```
ViMaSter: I think the name of this class could be improved:  
A `Shout` seems like a passive object (with just plain data, similar to a [ScriptableObject](https://docs.unity3d.com/Manual/class-ScriptableObject.html)) to me rather than a component that's responsible for executing logic.  
Something along the lines of `Shouter` or `ShoutPlayer` would make it sound a bit more "active".

Maybe there's something behind your initial decision that I'm missing, so definitely feel free to correct me or suggest other alternatives too. Native speaker advantage and all! : D

Assets/Scripts/Animation/Shout.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// Call this method to play the "take that" animation.
+    /// </summary>
+    /// <param name="actorData">An actor data containing an array of shout variants.</param>
+    public void TakeThat(ActorData actorData)
+    {
+        PlayShout(_takeThatSprite, actorData.TakeThat, actorData.ShoutVariants);
+    }
+
+    /// <summary>
+    /// Call this method to play the "take that" animation.
```
ViMaSter: This method's comment is identical to the `TakeThat()` method. I'd suggest either removing it completely or updating it to reflect this method's actual logic.

Assets/Scripts/Animation/Shout.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        _spriteRenderer = GetComponent<SpriteRenderer>();
+        _screenShaker = GetComponent<ScreenShaker>();
+    }
+    
+    /// <summary>
+    /// Call this method to play the "objection" animation.
+    /// </summary>
+    /// <param name="actorData">An actor data containing an array of shout variants.</param>
+    public void Objection(ActorData actorData)
```
ViMaSter: These 4 method names seem to be nouns; I usually try to either lead with or exclusively use verbs for method names, as methods describe what an instance of a class is doing.
How would you feel about prepending something like `Execute` (`ExecuteObjection(...)`, `ExecuteHoldIt(...)`, `ExecuteTakeThat(...)`, `ExecuteShoutout(...)`) or `Play`?

Assets/Scripts/Screenshake/ScreenShaker.cs
```diff
@@ -7,14 +7,23 @@ public class ScreenShaker : MonoBehaviour
...
     [SerializeField] private float _timeStep = 0.02f;
 
+    [Tooltip("The duration of the shake in seconds.")]
+    [SerializeField] private float _duration = 0.5f;
+    
+    [Tooltip("The frequency of the shake's sine wave in oscillations per second.")]
+    [SerializeField] private float _frequency = 10f;
+    
+    [Tooltip("The maximum distance the object can move from its starting point.")]
+    [SerializeField] private float _amplitude = 0.1f;
```
ViMaSter: It seems like, these three values, the default value assignment in line 26 and the parameterless overload of `Shake()` have been added, to allow for shouts to shake the screen, without having writers specific the "intensity" of screen shakes, parse it from the script and forward it. This is a great idea, but if that is the only reason for the changes on this file, I have an alternative proposal for the architecture:

Instead of having the `ScreenShaker` contain these default values and the new method overload, remove the new method overload (`ScreenShaker.cs:60-64`) completely, move the fields to `Shout` and directly call `ScreenShaker.Shake(float intensity, float duration, bool shouldWaitForShake)` with the values there.

As these values are only tied to how a screen shake caused by a `Shout` should look, `Shout` should define the intensity. This has other benefits too:
- the interface for `ScreenShaker` remains simple
- the code coverage for `ScreenShaker` isn't reduced by this PR
- reading the fields automatically grants a developer more context for their existence (as they are now part of `Shout`) and don't require comments that might be overlooked

What are your thoughts about this?

Assets/Scripts/Actor/ActorData.cs
```diff
@@ -53,3 +65,11 @@ private string GenerateNameWithAge()
         return $"{DisplayName} (Age: {age})";
     }
 }
+
+[Serializable]
+public struct SpriteAudioClipPair
+{
+    public string Name => Sprite.name;
+    [field: SerializeField] public Sprite Sprite { get; set; }
```
ViMaSter: TIL: You can apply attributes to the underlying fields of properties. That's suuuuper clean code! 👍

Assets/Scripts/Animation/Shout.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// </summary>
+    /// <param name="defaultSprite">The sprite that should be displayed.</param>
+    /// <param name="defaultAudio">The audio clip that should be played.</param>
+    /// <param name="shoutVariants">An array of alternate sprites
+    /// that have a chance to randomly appear instead. Set this to null to disable random variations.</param>
+    private void PlayShout(Sprite defaultSprite, AudioClip defaultAudio, SpriteAudioClipPair[] shoutVariants)
+    {
+        _spriteRenderer.enabled = true;
+
+        if (shoutVariants != null && shoutVariants.Length != 0 && Random.Range(0f, 1f) < _randomShoutChance)
```
ViMaSter: This line is pretty long and contains a lot of logic. I'd suggest extracting it into a `ShouldPlayShoutVariant()` method and call that instead.
```cs
private bool ShouldPlayShoutVariant(SpriteAudioClipPair[] shoutVariants)
{
    if (shoutVariants == null)
    {
        return false;
    }

    if (shoutVariants.Length == 0)
    {
        return false;
    }

    if (Random.Range(0f, 1f) < _randomShoutChance)
    {
        return true;
    }

    return false;
}
```

Assets/Scripts/Actor/ActorData.cs
```diff
@@ -31,6 +31,18 @@ private enum AgeCertainty {
...
+
+    [field: SerializeField, Tooltip("Drag this actor's 'HoldIt' sound here")]
+    public AudioClip HoldIt { get; private set; }
+    
+    [field: SerializeField, Tooltip("Drag this actor's 'TakeThat' sound here")]
+    public AudioClip TakeThat { get; private set; }
+
+    [field: SerializeField, Tooltip("Drag sprites for any shout variations this actor has here.")]
+    public SpriteAudioClipPair[] ShoutVariants { get; private set; }
+
```
ViMaSter: This implementation definitely reduces redundancy, but I personally feel like exposing three separate fields, and then only overriding the AudioClip for an actor actually makes this a bit difficult to read. Defining the **audio overrides** and **default sprites** in **two different places** inside the code, makes locating the defaults a bit difficult and also introduces a lot of redundant code.

Instead of having separate fields, I'd only use the `ShoutVariants` array and add all normal variants there too.
I see multiple benefits there:
- [line 36-61 of `Shout.cs`](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/96/files#diff-5d92497b7c8f7258f323a9981e5e886c21f6cb02982e2e8841330f4bcfbfb526R36-R61) and [line 341-366 of `SceneController.cs`](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/96/files#diff-24f1b69f75ec838fa74f3c0f78b671b8df0cd85149610bea2b062b41232ddb54R341-R366) would become completely redundant:  
    instead a call to e.g. `.ShoutOut(actorData, "objection")`, `.ShoutOut(actorData, "takethat")`, etc. would suffice
- right now the methods related to the `SHOUTOUT`-Ink script action fulfil no purpose, but this change would flip that, making it where every type of shout ends up eventually
- the redundance of re-defining the Objection/HoldIt/TakeThat sprite for each `ActorData` instance actually makes all sprites and sounds tied to that character visible at one glance (and batch-changing them, if needed, is easy thanks to Unity Editor's support for it too) image

If we want to play a special clip, we could get them like so:
```cs
private readonly string[] normalShouts = {"objection", "holdit", "takethat"};
...
var specialShouts = actorData.ShoutVariants.Where(variant => !normalShouts.Contains(variant.Name));
```

---

Do you feel this is something worth considering?

ViMaSter (commented):
Ngl., I can't wait for the final assets to come in and randomly be confronted with a special `Objection!` sound. Really looking forward to this feature! : D

I've got a few questions and ideas regarding this implementation and I've added them as comments, and wanted to hold off with my approval until they're at least discussed. 

That said, if two other people feel like this is already in a mergeable state, these topics are nothing I would halt the merge for, so I didn't mark them as "required changes" either! :)

Assets/Scripts/AudioController/AudioController.cs
PhoebeMitchell: Agreed, but as you're making heavy modifications to DirectorActionController maybe it would be better to do this in that PR, or a separate PR

Assets/Scripts/Animation/Shout.cs
PhoebeMitchell: Changed to ShoutPlayer

Assets/Scripts/Animation/Shout.cs
PhoebeMitchell: Fixed

Assets/Scripts/Animation/Shout.cs
PhoebeMitchell: Change to PlayObjection etc

Assets/Scripts/Screenshake/ScreenShaker.cs
PhoebeMitchell: Moved variables to ShoutPlayer. Shake method now has frequency and amplitude parameters. SceneController now converts intensity to frequency and amplitude.

Assets/Scripts/Animation/Shout.cs
PhoebeMitchell: Extracted it into its own method, though personally I don't feel like it needs to be three if statements.

Assets/Scripts/Actor/ActorData.cs
PhoebeMitchell: Objection, Holdit and Takethat are now part of the shout variants array. I am not a huge fan of string dependency in case the names of the sprites change so I devoted the first three slots to the normal shouts. Thoughts?

Assets/Scripts/AudioController/AudioController.cs
ViMaSter: Works for me! :) Whenever you get to merge this PR in, just ping me here or on Discord, so I can get started on that. Cross-integrating PR changes is never fun, so I'd rather do it, as soon as this change is on `develop`.

Assets/Scripts/Actor/ActorData.cs
PhoebeMitchell: Thinking more about it, might it make more sense for the scene controller to communicate with the audio controller, instead of ShoutPlayer? So SceneController would pass a sprite to ShoutPlayer to play, and pass the audio clip to AudioController to play.

lyndsiWilliams (dismissed):
Code looks great and the ShoutTest script runs great! All the tests passed, nice work!

Assets/Scripts/Actor/ActorData.cs
ViMaSter: > Objection, Holdit and Takethat are now part of the shout variants array. I am not a huge fan of string dependency in case the names of the sprites change so I devoted the first three slots to the normal shouts. Thoughts?

Hmm... I find strings more readable than an implicit rule like "the first three slots are different to the rest". Given the option, I personally would prefer having string dependencies over [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)#:~:text=Unique%20values%20with%20unexplained%20meaning%20or%20multiple%20occurrences%20which%20could%20(preferably)%20be%20replaced%20with%20named%20constants), which was the reason for the `normalShouts` and `specialShouts` variables in my suggestion. That being said, you definitely raise a valid point; I also don't really find string constants too pleasing either. : D

I think a middle ground might be having `Objection`, `HoldIt`, `TakeThat` be their own fields again (like your original implementation), but have them be a `SpriteAudioClipPair`, instead of an `AudioClip`. Although the sprites for it are rarely (or maybe even never) changed, [I would still like this to apply](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/96#:~:text=the%20redundance%20of%20re%2Ddefining%20the%20Objection/HoldIt/TakeThat%20sprite%20for%20each%20ActorData%20instance%20actually%20makes%20all%20sprites%20and%20sounds%20tied%20to%20that%20character%20visible%20at%20one%20glance%20(and%20batch%2Dchanging%20them%2C%20if%20needed%2C%20is%20easy%20thanks%20to%20Unity%20Editor%27s%20support%20for%20it%20too)). Even if I think the other two mentioned advantages would be worth it too, they're no must-haves in my book. :)

> Thinking more about it, might it make more sense for the scene controller to communicate with the audio controller, instead of ShoutPlayer? So SceneController would pass a sprite to ShoutPlayer to play, and pass the audio clip to AudioController to play.

I think I actually like the idea of having a dedicated `ShoutPlayer`. Is there any specific reason this change came to mind?

Right now, everything shout-related can be found there:
- Want a shout to appear? Call a method on it.
- Curious which sound or sprite is shown and why? Look inside the code.
- Want to cover shouts with tests? You only need to cover `ShoutPlayer` for it.

Sure, both the `DirectorActionDecoder` and `SceneController` only wrap the calls to it, but I think that's that's a fine scenario and just a result of our architecture and ours usage of "*controller" scripts.

PhoebeMitchell: Ooo, didn't see you'd posted this.

> Thinking more about it, might it make more sense for the scene controller to communicate with the audio controller, instead of ShoutPlayer? So SceneController would pass a sprite to ShoutPlayer to play, and pass the audio clip to AudioController to play.

My reasoning for this was the idea being talked about that the Controller classes should be in charge of communicating with each other, and avoiding the reference spaghetti I find Unity scenes often end up with.

ViMaSter: That's fair. I think we're generally missing a "hierarchy of interaction" between classes, and this is another symptom, so I'm not sure if there's an easy way around it.

ViMaSter: Ah, yeah, that's definitely a fair point. I think we're generally missing a "hierarchy of interaction" between classes, and this is another symptom; not sure if there's an easy way to fixing this without a general review of how classes interact inside our game on a higher level.

Explosive mentioned something along this line a few days ago, but I think but I think for the sake of this feature, I'd be cool with just merging the current state of it. Reworking it is always possible, but maybe only if we actually do run into problems with its current state, as - from a gameplay perspective - the feature works fine inside this branch and that's the main priority.

PhoebeMitchell: Decided to just go with your original suggestion in the end of having everything in one array, and using Linq to select the correct shouts.
Also put the AudioController reference on ShoutPlayer for now until a proper structure is decided upon.

Assets/Scripts/Animation/ShoutPlayer.cs
```diff
@@ -39,11 +43,15 @@ private void Awake()
...
+    /// <param name="shoutName">The name of the shout to play</param>
+    /// <param name="allowRandomShouts">Whether random shouts should be allowed to play (true) or not (false)</param>
+    public void Shout(Pair<Sprite, AudioClip>[] shoutVariants, string shoutName, bool allowRandomShouts)
     {
-        var shoutManager = new ShoutManager(shoutVariants, _randomShoutChance);
-        var shout = shoutManager.GetShout(shoutName);
+        var variants = shoutVariants.Where(pair => NormalShoutNames.All(normalShoutName => normalShoutName  != pair.Item1.name)).ToArray();
+        var shout = allowRandomShouts && variants.Any() && Random.Range(0f, 1f) < _randomShoutChance
+            ? variants[Random.Range(0, variants.Length)]
+            : shoutVariants.First(pair => pair.Item1.name == shoutName);
```
ViMaSter: Nit: I'd personally be a fan if we could we move the qualifier and generator into their own methods, like [in the previous implementation](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/2582f874df7ea0175fa2120b8ef23e571f51d77d/Assets/Scripts/Animation/ShoutManager.cs#L25-L51).
`_shoutVariants.Any(pair => NormalShoutNames.Any(name => name != pair.Item1.name))` could also be moved into something like a `SpecialVariants` property, that way the conditional operator would look super clean!
(Although I generally prefer a method with early outs over conditional operators, specifically for factory methods like this.)

ViMaSter (dismissed):
Looks good to me! Totally feel free to merge it in and address the nit later on, so this PR can come to a close! : D

Assets/Scripts/Animation/ShoutPlayer.cs
PhoebeMitchell: Re-extracted this logic into its own class (Shout.cs), and also wrote some tests for it

Assets/Scripts/Animation/ShoutPlayer.cs
```diff
@@ -0,0 +1,55 @@
+using System.Linq;
+using UnityEngine;
+using Random = UnityEngine.Random;
```
ViMaSter: Nit: `using System.Linq` and `using Random = UnityEngine.Random;` should be redundant now that those symbols reside inside `Shout.cs`

Assets/Scripts/BGScenes/SceneController.cs
```diff
@@ -1,6 +1,8 @@
 using System.Collections;
+using System.Linq;
 using UnityEngine;
 using UnityEngine.Events;
+using Random = UnityEngine.Random;
```
ViMaSter: Nit: `using System.Linq` and `using Random = UnityEngine.Random;` should be redundant now that those symbols reside inside Shout.cs

Assets/Tests/EditModeTests/ShoutTests.cs
```diff
@@ -0,0 +1,71 @@
...
+    public void ShoutsCanBeAccessedByName()
+    {
+        for (int i = 0; i < SpecialVariantsLength; i++)
+        {
+            var shout = new Shout(i.ToString(), _shoutVariants, false, 0);
+            Assert.AreEqual(i.ToString(), shout.Sprite.name);
+            Assert.AreEqual(i.ToString(), shout.AudioClip.name);
+        }
+
+        for (int i = 0; i < Shout.NormalShoutNames.Length; i++)
```
ViMaSter: Nit: Since only the iterated object is accessed (`Shout.NormalShoutNames[i]`) this could be converted into a `foreach` loop, and would remove the need for the compiler to keep track of an index when looping over `Shout.NormalShoutNames`.

Assets/Tests/EditModeTests/ShoutTests.cs
```diff
@@ -0,0 +1,71 @@
...
+using NUnit.Framework;
+using UnityEngine;
+
+public class ShoutTests
+{
+    private const int VARIANTS_LENGTH = 8;
+    
+    private readonly Pair<Sprite, AudioClip>[] _shoutVariants = new Pair<Sprite, AudioClip>[VARIANTS_LENGTH];
+
+    private int SpecialVariantsLength => VARIANTS_LENGTH - Shout.NormalShoutNames.Length;
```
ViMaSter: Nit: This could be made static. That way initialization is baked into the assembly, rather than being reinitialized, every time an instance of this class is created.

Assets/Tests/EditModeTests/ShoutTests.cs
```diff
@@ -0,0 +1,71 @@
+using System.Linq;
+using NUnit.Framework;
+using UnityEngine;
+
+public class ShoutTests
```
ViMaSter, line 5: Covering code with tests is probably the best way to get me to shut up about things looking "too complicated to read". Awesome work! 👍

ViMaSter (dismissed):
Few nits, but nothing worth blocking as usual! Always great to see more test coverage, so the new ShoutTests made me extra happy, thanks for adding those!

Assets/Tests/EditModeTests/ObjectshakeCalculatorTests.cs
```diff
@@ -3,80 +3,80 @@
...
+    /// The amplitude of a Object shake should not be greater than the amplitude specified.
     /// Gets all the positions and makes sure they are all within a range of -AMPLITUDE and AMPLITUDE.
     /// Also makes sure all values are not the same.
     /// </summary>
     [Test]
     public void ShakeIsConstrainedWithinSpecifiedAmplitude()
     {
         var animationKeys = new [] { new Keyframe(0, 1), new Keyframe(1, 1) };
-        var screenshakeCalculator = CreateScreenshakeCalculator(animationCurve: new AnimationCurve(animationKeys));
+        var ObjectshakeCalculator = CreateObjectshakeCalculator(animationCurve: new AnimationCurve(animationKeys));
```
notexplosive: nit: vars should be `lowerCase`

Assets/Tests/EditModeTests/ObjectshakeCalculatorTests.cs
```diff
@@ -108,12 +108,12 @@ public void NoShakeWhenAmplitudeIsZero()
     public void NumberOfOscillationsMatchesFrequency()
     {
         var animationKeys = new [] { new Keyframe(0, 1), new Keyframe(1, 1) };
-        var screenshakeCalculator = CreateScreenshakeCalculator(noiseScaleX: 0, noiseScaleY: 0, animationCurve: new AnimationCurve(animationKeys));
-        int previousSign = Math.Sign(screenshakeCalculator.Calculate(DELTA_TIME).x);
+        var ObjectshakeCalculator = CreateObjectshakeCalculator(noiseScaleX: 0, noiseScaleY: 0, animationCurve: new AnimationCurve(animationKeys));
```
notexplosive: ditto

Assets/Tests/EditModeTests/ObjectshakeCalculatorTests.cs
```diff
@@ -89,11 +89,11 @@ public void ShakeIsConstrainedWithinSpecifiedAmplitude()
     [Test]
     public void NoShakeWhenAmplitudeIsZero()
     {
-        var screenshakeCalculator = CreateScreenshakeCalculator(amplitude: 0);
+        var ObjectshakeCalculator = CreateObjectshakeCalculator(amplitude: 0);
```
notexplosive: ditto

Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -106,6 +106,15 @@ public void PlaySfx(AudioClip soundEffectClip)
     {
         _sfxAudioSource.PlayOneShot(soundEffectClip);
     }
+    
+    /// <summary>
+    /// Overload for PlaySFX that plays sound using an audio clip rather than a name.
+    /// </summary>
+    /// <param name="soundEffect">The audio clip to play.</param>
+    public void PlaySFX(AudioClip soundEffect)
```
notexplosive: I added this exact same function in a previous PR (although I capitalized it different). see line 105 just above this.

notexplosive (dismissed):
(sorry for delaying this PR even more. It's _almost_ there!)

Assets/Scripts/AudioController/AudioController.cs
PhoebeMitchell: Good catch 👍

ViMaSter (dismissed):
Still looking good to me!

ViMaSter (approved):

notexplosive (approved):
😂🙏🙏🙋‍♀️

PR #97, opened 2021-10-30
Implement compile-time constant action mapping inside ActionDirector

ViMaSter:
## Summary
As talked about, this introduces an easier way to add or modify actions to the ActionDirector.

Adding an action literally requires one new method now. ([example](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/b02171c25ef967864e99cb62a52f20f26874cd64/Assets/Scripts/TextDecoder/ActionDecoder.cs#L325-L335))
[All existing methods have been ported.](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/b02171c25ef967864e99cb62a52f20f26874cd64/Assets/Scripts/TextDecoder/ActionDecoder.cs#L103-L341)

Just ensure that...
  - The method is **private**
  - The method's **name only consists of capital letter and underscores** (`[A-Z_]+`) and matches what you want the Ink script to use
  - **Optional parameters** are added in **regular C#-syntax** (i.e. `private void FOO(string bar = "default")`
  - All **parameters** have a **parser for their type**. [If unused types are used, appropriate error messages are thrown, once the decoder encounters such a line.](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/b02171c25ef967864e99cb62a52f20f26874cd64/Assets/Scripts/TextDecoder/ActionDecoder.cs#L62-L77)  
    The currently supported types are:
    - [bool](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/b02171c25ef967864e99cb62a52f20f26874cd64/Assets/Scripts/TextDecoder/Parser/BoolParser.cs)
    - [float](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/b02171c25ef967864e99cb62a52f20f26874cd64/Assets/Scripts/TextDecoder/Parser/FloatParser.cs)
    - [int](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/b02171c25ef967864e99cb62a52f20f26874cd64/Assets/Scripts/TextDecoder/Parser/IntParser.cs)
    - [string](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/b02171c25ef967864e99cb62a52f20f26874cd64/Assets/Scripts/TextDecoder/Parser/StringParser.cs)
    - [ItemDisplayPosition enums](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/b02171c25ef967864e99cb62a52f20f26874cd64/Assets/Scripts/TextDecoder/Parser/ItemDisplayPositionParser.cs)  

---

This change also introduces **auto-generated tests** for parsing **each action**. If a method is added and respects the guidelines above, the following tests are automatically generated for it:
- A **valid** line for any given action
  - For actions with optional parameters, **all possible cases are run** (without, with some and with all optional parameters present)
- A line with **too few** arguments for any given action
  - Actions that support **no parameters** at all, are exempt
- A line with **too many** arguments for any given action
- A line with the correct number, but **incorrect types of arguments** for any given action
  - Actions that **only accept `string`s** are exempt

Tests can be found in the `ActionDecoderTests` category.  
To see the actual Inky line, that's being tested, run the test, click on it and check the line under `Attempting to parse:`:
zSydHHwryk

## Testing instructions
1. Switch to this branch
2. Play the game (I've run the current `Inky-TestScene` with the `Ross_Cool_X` narrative script
3. Ensure all `PlayMode`- and `EditMode`-tests pass

## Additional information

- `[ ]` Changes UI


Assets/Scripts/TextDecoder/Parser/Parser.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace TextDecoder.Parser
+{
+    public abstract class Parser<T> : IParser
+    {
+        public abstract string Parse(string input, out T output);
+
+        public bool Applies(Type type)
```
PhoebeMitchell: Summaries might be nice for these, i.e. whats the purpose of parser, how to use it, and when to use Applies

PhoebeMitchell: Applies also seems to be unused, unless I'm just missing it.

Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -1,233 +1,186 @@
...
+        ChangeDialogSpeed(WaiterType.Overall, seconds);
+    }
+    private void PUNCTUATION_SPEED(float seconds)
+    {
+        ChangeDialogSpeed(WaiterType.Punctuation, seconds);
+    }
     private void ChangeDialogSpeed(WaiterType currentWaiterType, float seconds)
     {
         AppearingDialogueController.SetTimerValue(currentWaiterType, seconds);
     }
```
PhoebeMitchell: Could these methods just call SetTimerValue directly instead of calling ChangeDialogSpeed? I feel AppearingDialogueController should handle changing the dialogue speed anyway so it doesn't make sense to me to have a change dialog speed method on ActionDecoder.

Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -413,23 +302,24 @@ private void SetActorVisibility(bool shouldShow)
...
-    /// <param name="actor">Actor to make the speaker</param>
-    /// <param name="speakingType">Type of speaking to speak the text with</param>
+    private void SPEAK(string actor)
+    {
+        SetSpeaker(actor, SpeakingType.Speaking);
+    }
+
+    private void THINK(string actor)
+    {
+        SetSpeaker(actor, SpeakingType.Speaking);
```
PhoebeMitchell: SpeakingType.Thinking instead?

Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -1,233 +1,186 @@
...
-    HIDE_TEXTBOX,
-    WAIT_FOR_INPUT
-}
-
 public class ActionDecoder
 {
-    /// <summary>
-    /// Forwarded from the DirectorActionDecoder
-    /// </summary>
     public event Action OnActionDone;
```
PhoebeMitchell, line 9: Suggestion: Instead of a separate event, could we just use the UnityEvent in DirectorActionDecoder, and inject it upon construction?

Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -1,233 +1,186 @@
...
+        }
+
+        // If the method supports optional parameters, fill the remaining parameters based on the default value of the method
+        for (int suppliedParameterCount = parameters.Length; suppliedParameterCount < methodParameters.Length; suppliedParameterCount++)
+        {
+            parsedMethodParameters.Add(methodParameters[suppliedParameterCount].DefaultValue);
+        }
+
+        // Call the method
+        method.Invoke(this, parsedMethodParameters.ToArray());
```
PhoebeMitchell, line 117: jfk-clone-high

Assets/Tests/EditModeTests/ActionDecoderTests.cs
```diff
@@ -1,170 +1,172 @@
...
 using System;
 using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text.RegularExpressions;
+using TextDecoder.Parser;
+using UnityEngine;
+using UnityEngine.TestTools;
 
 namespace Tests.EditModeTests
```
PhoebeMitchell: This namespace separates this test from the rest. I'd suggest adding the namespace to the other edit mode tests, or just remove it, as the edit mode tests are in their own assembly anyway and cant be accessed by the game scripts, and this would avoid so many dropdowns called "Test".
Screenshot 2021-11-05 at 22 17 26

PhoebeMitchell (approved):
All looks good! Just some nits. Question about the PR itself, would it not make more sense to merge it straight into develop? As opposed to merging into NotExplosive's branch and then needing to wait for two more approvals to finally merge it properly.

Assets/Scripts/TextDecoder/Parser/Parser.cs
ViMaSter: Just had a look at it: Good catch! This method and the entire underlying interface are superfluous. Removing them in a future commit.

ViMaSter: Makes sense! I've added a few lines to `Parser<T>` but mainly added information about `ActionDecoder` and how it uses derivates of this class.

Assets/Scripts/TextDecoder/ActionDecoder.cs
ViMaSter: I think this was done in the original PR, so that `ActionDecoder` has no dependencies on Unity itself. Only `DirectorActionDecoder` uses Unity-specific classes like `UnityEvent`.

Assets/Scripts/TextDecoder/ActionDecoder.cs
ViMaSter: Added comments, just for clarity. : D  
Strong meme-game nevertheless. 👀

PR #99, opened 2021-11-05
Feature/70 story handler

### LINKED ISSUE #70: Implement story handler

PhoebeMitchell:
## Summary

Implements story controller. Identical to [this PR](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/82) except with some added polish.

"Stories are now handled by the story handler, which can transition between different inky dialogue scripts. It also loads the next unity scene when reaching the end"

## Testing instructions

Run the ink test scene and make sure the narratives scripts transition into each other. Also run the tests.
Note: the screenshake doesn't stop when continuing to the next narrative script. This could be considered a bug (and it will need to be fixed in another PR if so), though I suspect screenshakes wont be that long usually, so it might not matter.

## Additional information

- `[ ]` Changes UI
  
closes #70


lyndsiWilliams (approved):
Code looks great and the test scene runs great! All the tests passed, nice work!

Assets/Scripts/TextDecoder/StoryController.cs
```diff
@@ -9,11 +9,10 @@ public class StoryController : MonoBehaviour
     [SerializeField] private List<Dialogue> _dialogueList;
 
     [Header("Events")]
-
+
```
ViMaSter, line 12: I think this line break is superfluous and should be safe to remove.

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuKeyboardTests.cs
```diff
@@ -21,9 +21,10 @@ public class EvidenceMenuKeyboardTests
         [ReloadScene("Assets/Scenes/EvidenceMenu - Test Scene.unity")]
         public IEnumerator EvidenceMenuOpensAndCloses()
         {
-
             yield return null;
-            global::EvidenceMenu evidenceMenu = InputTestTools.FindInactiveInScene<global::EvidenceMenu>()[0];
+            var dialogueController = Object.FindObjectOfType<global::DialogueController>();
+            yield return TestTools.WaitForState(() => !dialogueController.IsBusy);
```
ViMaSter, line 26: 👀 👌

ViMaSter (approved):
Looks good to me! <sup>(Wouldn't be a review without a nit though. Just a minor formatting thing, feel free to skip.)</sup>

Assets/Scripts/TextDecoder/StoryController.cs
PhoebeMitchell: Agreed that you will handle this in your branch.

PR #100, opened 2021-12-05
Generalize enum parsing inside ActionDecoder and add respective tests

ViMaSter:
## Summary
This change removes the need to add explicit `Parser`s for each enum type that should be supported as parameter type inside the `ActionDecoder`.

Before this change, each instance of an `enum` that was available as value type inside our narrative scripts, required an extra class which derives from `Parser<T>`. For example, [ItemDisplayPositionParser](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/4b6db7a79046064a95806d209c104c1b86cd755e/Assets/Scripts/TextDecoder/Parser/ItemDisplayPositionParser.cs#L1).

After this change, [`enum`s are handled generically](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/4b6db7a79046064a95806d209c104c1b86cd755e/Assets/Scripts/TextDecoder/ActionDecoder.cs#L74-L92): If a parameter type derives from the type `System.Enum`, it's attempted to be cast in a generic way.

## Testing instructions
1. Create a new enum type
2. Attempt to add a new "event" inside the ActionDecoder with that new enum as parameter
3. Create a narrative script that uses a value of that enum type
4. Notice the script being executed successfully
5. Deliberately add a typo to the value inside your narrative script (I.e. change `Left` to `LLeft`)
6. Notice the script failing to execute and a human-readable message including all possible values being displayed inside the `Console` window in Unity

Two test cases have also been added:
* `ActionDecoderEnumTests.ParseActionLineWithValidEnumValue`
* `ActionDecoderEnumTests.ParseActionLineWithInvalidEnumValue`

## Additional information

- `[ ]` Changes UI
  - https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/93#discussion_r762464924

### DIFF
--- a/Assets/Scripts/TextDecoder/ActionDecoder.cs
+++ b/Assets/Scripts/TextDecoder/ActionDecoder.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Text.RegularExpressions;
 using UnityEngine;
 
 public class ActionDecoder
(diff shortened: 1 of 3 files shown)


notexplosive (approved):
good

PhoebeMitchell (approved):
good. The parser classes don't seem to be used anymore as far as I can tell? Can they be removed?

ViMaSter:
Thanks for the review, both of you!
I'll merge this in, if there's still something unclear or needs a change, feel free to hit me up on Discord or create a draft PR. :)

---
Re:
> The parser classes don't seem to be used anymore as far as I can tell?

They are! Not directly by name however, hence a simple `Ctrl+F` search or IntelliSense (and similar software) will show it as "unused".

Every time we encounter an action which uses a parameter (for example a `float` or `bool`) [this line gets called](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/refactor/director-action-decoder-generic-enum-support/Assets/Scripts/TextDecoder/ActionDecoder.cs#L95). It searches for any class, which has a base class, that is a template class, with the parameter type we expect (for example `float` or `bool`, which will match `Parser<float>` or `Parser<bool>` and thereby `FloatParser` and `BoolParser` respectively).

PR #101, opened 2021-12-15
Feature/cross examination penalties

PhoebeMitchell:
## Summary

Adds penalties, gameover dialogue, and associated test.

Kapture 2021-12-15 at 02 51 14

## Testing instructions

Run Ink-TestScene and try to present evidence by pressing z then x. A failure dialogue will play which will remove one life. After losing all lives, the gameover dialogue plays and the scene will restart.

Run the test GameOverPlaysOnNoLivesLeft.

## Additional information

- `[x]` Changes UI


Assets/Scripts/BGScenes/SceneController.cs
```diff
@@ -332,4 +342,25 @@ public Vector2 PixelPositionToUnitPosition(Vector2Int pixelPosition)
...
         return new Vector2((float)(pixelPosition.x * -1) / _pixelsPerUnit, (float)(pixelPosition.y * -1) / _pixelsPerUnit);
     }
+
+    /// <summary>
+    /// Issues a penalty to the player, decrementing their lives by one.
+    /// If their lives are zero the game over event is called.
+    /// </summary>
+    public void IssuePenalty()
+    {
+        if (!_penaltyManager.Decrement())
```
notexplosive: nit: `Decrement` is pulling double duty.

I think it would be more expressive to use two separate functions: `Decrement`, which returns void and `IsOutOfPenalties` which returns a bool.

Assets/Scripts/SceneLoading/SceneLoader.cs
```diff
@@ -35,16 +34,24 @@ private void Awake()
     /// Call this method when wanting to change the scene using a specific scene's index.
     /// </summary>
     public void ChangeSceneBySceneIndex()
```
notexplosive: I know this is (technically) outside your code but why do we have a ChangeSceneByIndex() that takes no args? Is there a better name we can pick to explain that? 

Something like `ChangeToThisScene`? (assuming I'm reading this right)

notexplosive (dismissed):
Haven't downloaded it yet but this looks great!

Assets/Scripts/UI/PenaltyManager.cs
```diff
@@ -0,0 +1,35 @@
+using System.Collections.Generic;
+using UnityEngine;
+
+public class PenaltyManager : MonoBehaviour
+{
+    [Tooltip("Drag the prefab for the penalty UI object here.")]
+    [SerializeField]private GameObject _penaltyObject;
+
+    private Queue<GameObject> _penaltyObjects = new Queue<GameObject>();
```
ViMaSter: Nit: This field can be made `readonly`, as it's only assigned on initialization.

Assets/Scripts/SceneLoading/SceneLoader.cs
ViMaSter: I think this had something to do with the fact, that there was no way to call a method with parameters using UnityEvents a few Unity versions ago.

I.e.: If you want to have a button which on click loads "scene 0", there was no way to call `ChangeSceneBySceneIndex(0)` directly, only other public methods without parameters. Hence, specifying a GameObject, which has a preset `SceneLoader`-component with `_sceneIndex` set to `0` and calling the `ChangeSceneBySceneIndex()` (without parameters) might have been a workaround:
image

Just a pure guess though. And that said, I just double checked and that functionality does exist now:
image
👆results in 👇
image

---

I find that doing it the way we do right now, adds a lot of complexity (due to not having the target scene index or name specified at the same place where the event method actually gets invoked, but in the other game object instead). I can create a clean-up PR for it, if both you and moths agree. (Feel free to just close the convo if you do, or suggest an alternative.)

Assets/Scripts/BGScenes/SceneController.cs
ViMaSter: [One possible alternative would be replacing the `bool` returned with a more descriptive `enum`](https://codereview.stackexchange.com/q/11300), but in the interest of legibility, I think I would also prefer Explosive's approach of splitting both methods (and maybe also add an `assert()` for `Decrement()`, that it's never called, with 0 or less "attempts" remaining).

ViMaSter (dismissed):
Looks good to me! Having the PenaltyManager handle the UI stuff directly is super sleek!

Assets/Scripts/SceneLoading/SceneLoader.cs
PhoebeMitchell: The target scene was set using a field in the inspector originally, as this would allow the target scene to be set depending on the current scene, and the StoryController could just load the next scene as needed. I found that if I wanted to reset a scene I'd need a separate SceneLoader, so I added an overload so just one SceneLoader could be used.

Maybe the target scene could be moved to the StoryController?

PhoebeMitchell: Moved the name of the scene to load from SceneLoader to StoryController so now SceneLoader does not dictate which scenes are loaded. 

> I can create a clean-up PR for it, if both you and moths agree

Go for it 👍

Assets/Scripts/BGScenes/SceneController.cs
PhoebeMitchell: Decrement now only decrements. Checking for GameOver is done by SceneController using PenaltiesLeft property. Added Assert to Decrement method.

ViMaSter (approved):
Pretty happy with these changes, so everything looks good to me now! (Thanks for addressing the parameterless scene loading thing too!)

notexplosive (approved):
merry krimbus

PR #102, opened 2021-12-17
Feature/penalty animations

PhoebeMitchell:
## Summary

Adds the explosion animation for when you take a penalty. Also fixes a bug where the pause menu exit button didn't work working due to having no target scene.

penaltyanimation

## Testing instructions

Run the GameOverPlaysOnNoLivesLeft test and witness the animation.

## Additional information

- `[x]` Changes UI

### DIFF
--- a/Assets/Scripts/UI/DestroyOnFinish.cs
+++ b/Assets/Scripts/UI/DestroyOnFinish.cs
@@ -2,7 +2,7 @@ using UnityEngine;
 
 public class DestroyOnFinish : StateMachineBehaviour
 {
-    // OnStateExit is called when a transition ends and the state machine finishes evaluating this state
+    // OnStateEnter is called when a transition starts and the state machine starts to evaluate this state
     public override void OnStateEnter(Animator animator, AnimatorStateInfo stateInfo, int layerIndex)
     {
         Destroy(animator.gameObject, stateInfo.length);
--- a/Assets/Scripts/UI/DestroyOnFinish.cs
+++ b/Assets/Scripts/UI/DestroyOnFinish.cs
@@ -1,0 +1,10 @@
+using UnityEngine;
+
+public class DestroyOnFinish : StateMachineBehaviour
+{
+    // OnStateExit is called when a transition ends and the state machine finishes evaluating this state
+    public override void OnStateEnter(Animator animator, AnimatorStateInfo stateInfo, int layerIndex)
+    {
+        Destroy(animator.gameObject, stateInfo.length);
+    }
+}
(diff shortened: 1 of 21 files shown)


Assets/Scripts/UI/PenaltyManager.cs
```diff
@@ -29,8 +29,6 @@ public void Initialize(int penaltyCount)
     public void Decrement()
     {
         Debug.Assert(PenaltiesLeft > 0, "Decrement must not be called with 0 or fewer penalty lifelines left");
-
-        GameObject penaltyObject = _penaltyObjects.Dequeue();
-        Destroy(penaltyObject);
+        _penaltyObjects.Dequeue().Play("Explosion");
```
ViMaSter, line 32: I don't know why, but from time to time there's one-liners that I love and this is one of them. This method asserts a state, acts on it and wraps it in a method name that gives it the proper context. Super readable; love it! 👍

Assets/Scripts/UI/DestroyOnFinish.cs
```diff
@@ -0,0 +1,10 @@
+using UnityEngine;
+
+public class DestroyOnFinish : StateMachineBehaviour
+{
+    // OnStateExit is called when a transition ends and the state machine finishes evaluating this state
```
ViMaSter: Nit:  
My understanding of this class is that, because all GameObjects that this behavior is added to, are objects which animation state machines only ever consist of a "play 'A to B' once" tree, using `OnStateEnter` here works, because it **starts** in "state A" (and thereby never **enters** it from another) and then moves to "state B" (which is the only time it properly **enters** a new state).  

If that's the case (and only if : D), the mention of `OnStateExit` in this comment seems a bit misplaced to me. The comment reads to me, like It describes `OnStateExit()` being the place for the game object destruction to take place in (`[...] when a transition ends and the state machine finishes evaluating [...]`), but then the code is put into `OnStateEnter()` instead with no additional information.

I think replacing it with a description on either
- why `OnStateExit()` would be a bad place **or**
- why `OnStateEnter()` is the better place

instead, would clarify this implementation further. Would that make sense to you too?

And if I completely missed the point or its just a typo, definitely correct me and feel free to just close this convo. : D

---

That aside, never stumbled across `StateMachineBehaviour` before and seems like the perfect way to implement this. (Another thing learned about Unity thanks to this project. Digging the implementation. 👀👌)

ViMaSter (dismissed):
And another PR where I learn new things about Unity! \o\

Added a thread about either a typo or place to improve the comments a bit, but LGTM, definitely nothing worth blocking for!

Assets/Scripts/UI/DestroyOnFinish.cs
PhoebeMitchell: Good catch, it was originally an OnStateExit method but changed it to current implementation because it never actually exits the state. Changed the name but forgot the comment

ViMaSter (approved):
Re-approving; updated comment looks good, thanks for addressing it! 👍

notexplosive (approved):
❕❕❕❕💥

Issue #62, opened 2021-09-15
Actor specific letter sounds


JuiceGrape:
## Description

When dialogue is scrolling, the sound of letters popping up should change based on the person speaking. The sound should be able to be defined per actor

## Acceptance Criteria

This story will be finished when actors can be assigned different sounds in their scriptable objects, have their sounds assigned to them, and all this is testable in a unity scene.

PR #135, opened 2021-12-30
Feature/62 (and bug/111) Actor specific sounds

### LINKED ISSUE #111: [BUG] Text sound does not play
## Describe the bug
No sound plays when text appears.

## Steps To Reproduce
Play any script and you will hear no sound plays.

## Expected behavior
Sound should play when a non-punctuation character appears.

## Additional context
Wait until #103 is approved before addressing this bug.
Modify the currently unused OnLetterAppear method if needed.

notexplosive:
## Summary
Fixes #111 and adds feature #62. I mashed them into one PR because both were pretty tiny.

## Testing instructions
- Crack open Inky-TestScene and listen to Arin's beautiful `maleTalk` chirps. (sounds pretty dreadful when he talks fast, maybe some future work there).
- After you've done that go into the Assets/ScriptableObjects/Actors/Arin and change his `DialogueChirp` field to something like `chug` and listen to that boy chug.

## Additional information

- `[ ]` Changes UI
#111 and #62

### DIFF
--- a/Assets/Scripts/AudioController/AudioController.cs
+++ b/Assets/Scripts/AudioController/AudioController.cs
@@ -92,7 +92,7 @@ public class AudioController : MonoBehaviour, IAudioController
     /// <param name="soundEffectName">Name of sound effect asset, must be in `Resources/Audio/SFX`</param>
     public void PlaySFX(string soundEffectName)
     {
-        AudioClip soundEffectClip = GetSFXResource(soundEffectName);
+        AudioClip soundEffectClip = AudioController.GetSfxResource(soundEffectName);
         _sfxAudioSource.PlayOneShot(soundEffectClip);
     }
 
@@ -146,7 +146,7 @@ public class AudioController : MonoBehaviour, IAudioController
         _musicAudioSource.Play();
     }
 
-    public static AudioClip GetSFXResource(string soundEffectName)
+    public static AudioClip GetSfxResource(string soundEffectName)
     {
         return Resources.Load("Audio/SFX/" + soundEffectName) as AudioClip;
     }
--- a/Assets/Scripts/AudioController/AudioController.cs
+++ b/Assets/Scripts/AudioController/AudioController.cs
@@ -157,6 +157,11 @@ public class AudioController : MonoBehaviour, IAudioController
         _musicAudioSource.Play();
     }
 
+    /// <summary>
+    /// Acquires the Sfx Resource from Audio/SFX/*
+    /// </summary>
+    /// <param name="soundEffectName">Name of sound effect asset you want</param>
+    /// <returns></returns>
     private static AudioClip GetSfxResource(string soundEffectName)
     {
         return Resources.Load<AudioClip>("Audio/SFX/" + soundEffectName);
--- a/Assets/Scripts/AudioController/AudioController.cs
+++ b/Assets/Scripts/AudioController/AudioController.cs
@@ -1,5 +1,6 @@
 using System.Collections;
 using UnityEngine;
+using UnityEngine.Serialization;
 
 public class AudioController : MonoBehaviour, IAudioController
 {
@@ -17,8 +18,9 @@ public class AudioController : MonoBehaviour, IAudioController
     /// One day this will come from the "Settings," but for now it lives on a field
     /// </summary>
     [Tooltip("SFX Volume level set by player")]
+    [FormerlySerializedAs("_settingsSFXVolume")]
     [Range(0f, 1f)]
-    [SerializeField] private float _settingsSFXVolume = 0.5f;
+    [SerializeField] private float _settingsSfxVolume = 0.5f;
 
     [Tooltip("Total duration of fade out + fade in")]
     [Range(0f, 4f)]
@@ -58,7 +60,7 @@ public class AudioController : MonoBehaviour, IAudioController
             _musicAudioSource.volume = _musicFader.NormalizedVolume * _settingsMusicVolume;
 
         if (_sfxAudioSource != null)
-            _sfxAudioSource.volume = _settingsSFXVolume;
+            _sfxAudioSource.volume = this._settingsSfxVolume;
     }
 
     /// <summary>
@@ -90,17 +92,17 @@ public class AudioController : MonoBehaviour, IAudioController
     /// Plays sound effect of desired name.
     /// </summary>
     /// <param name="soundEffectName">Name of sound effect asset, must be in `Resources/Audio/SFX`</param>
-    public void PlaySFX(string soundEffectName)
+    public void PlaySfx(string soundEffectName)
     {
         AudioClip soundEffectClip = AudioController.GetSfxResource(soundEffectName);
-        PlaySFX(soundEffectClip);
+        PlaySfx(soundEffectClip);
     }
 
     /// <summary>
     /// Play given audio clip immediately
     /// </summary>
     /// <param name="soundEffectClip">Clip to play</param>
-    public void PlaySFX(AudioClip soundEffectClip)
+    public void PlaySfx(AudioClip soundEffectClip)
     {
         _sfxAudioSource.PlayOneShot(soundEffectClip);
     }
--- a/Assets/Scripts/AudioController/AudioController.cs
+++ b/Assets/Scripts/AudioController/AudioController.cs
@@ -94,7 +94,7 @@ public class AudioController : MonoBehaviour, IAudioController
     /// <param name="soundEffectName">Name of sound effect asset, must be in `Resources/Audio/SFX`</param>
     public void PlaySfx(string soundEffectName)
     {
-        AudioClip soundEffectClip = AudioController.GetSfxResource(soundEffectName);
+        AudioClip soundEffectClip = GetSfxResource(soundEffectName);
         PlaySfx(soundEffectClip);
     }
 
--- a/Assets/Scripts/AudioController/AudioController.cs
+++ b/Assets/Scripts/AudioController/AudioController.cs
@@ -93,6 +93,15 @@ public class AudioController : MonoBehaviour, IAudioController
     public void PlaySFX(string soundEffectName)
     {
         AudioClip soundEffectClip = AudioController.GetSfxResource(soundEffectName);
+        PlaySFX(soundEffectClip);
+    }
+
+    /// <summary>
+    /// Play given audio clip immediately
+    /// </summary>
+    /// <param name="soundEffectClip">Clip to play</param>
+    public void PlaySFX(AudioClip soundEffectClip)
+    {
         _sfxAudioSource.PlayOneShot(soundEffectClip);
     }
 
--- a/Assets/Scripts/AudioController/AudioController.cs
+++ b/Assets/Scripts/AudioController/AudioController.cs
@@ -92,7 +92,7 @@ public class AudioController : MonoBehaviour, IAudioController
     /// <param name="soundEffectName">Name of sound effect asset, must be in `Resources/Audio/SFX`</param>
     public void PlaySFX(string soundEffectName)
     {
-        AudioClip soundEffectClip = Resources.Load("Audio/SFX/" + soundEffectName) as AudioClip;
+        AudioClip soundEffectClip = GetSFXResource(soundEffectName);
         _sfxAudioSource.PlayOneShot(soundEffectClip);
     }
 
@@ -144,5 +144,10 @@ public class AudioController : MonoBehaviour, IAudioController
         _musicAudioSource.clip = Resources.Load("Audio/Music/" + songName) as AudioClip;
         _musicAudioSource.volume = 0f; // Always set volume to 0 BEFORE playing the audio source
         _musicAudioSource.Play();
+    }
+
+    public static AudioClip GetSFXResource(string soundEffectName)
+    {
+        return Resources.Load("Audio/SFX/" + soundEffectName) as AudioClip;
     }
 }
--- a/Assets/Scripts/AudioController/AudioController.cs
+++ b/Assets/Scripts/AudioController/AudioController.cs
@@ -157,7 +157,7 @@ public class AudioController : MonoBehaviour, IAudioController
         _musicAudioSource.Play();
     }
 
-    public static AudioClip GetSfxResource(string soundEffectName)
+    private static AudioClip GetSfxResource(string soundEffectName)
     {
         return Resources.Load<AudioClip>("Audio/SFX/" + soundEffectName);
     }
--- a/Assets/Scripts/AudioController/AudioController.cs
+++ b/Assets/Scripts/AudioController/AudioController.cs
@@ -152,13 +152,13 @@ public class AudioController : MonoBehaviour, IAudioController
     /// <param name="songName">Name of song asset, must be in `Resources/Audio/Music`</param>
     private void SetCurrentTrack(string songName)
     {
-        _musicAudioSource.clip = Resources.Load("Audio/Music/" + songName) as AudioClip;
+        _musicAudioSource.clip = Resources.Load<AudioClip>("Audio/Music/" + songName);
         _musicAudioSource.volume = 0f; // Always set volume to 0 BEFORE playing the audio source
         _musicAudioSource.Play();
     }
 
     public static AudioClip GetSfxResource(string soundEffectName)
     {
-        return Resources.Load("Audio/SFX/" + soundEffectName) as AudioClip;
+        return Resources.Load<AudioClip>("Audio/SFX/" + soundEffectName);
     }
 }
--- a/Assets/Scripts/UI/AppearingDialogueController.cs
+++ b/Assets/Scripts/UI/AppearingDialogueController.cs
@@ -9,6 +9,12 @@ public class AppearingDialogueController : MonoBehaviour, IAppearingDialogueCont
 {
     [Tooltip("Drag a DirectorActionDecoder component here.")]
     [SerializeField] private DirectorActionDecoder _directorActionDecoder;
+
+    [Tooltip("Drag a AudioController here.")]
+    [SerializeField] private AudioController _audioController;
+    
+    [Tooltip("Drag a NameBox component here.")]
+    [SerializeField] private NameBox _namebox;
     
     [Tooltip("Drag a TextMeshProUGUI component here.")]
     [SerializeField] private TextMeshProUGUI _textBox;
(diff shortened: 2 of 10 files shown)


Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -145,4 +154,9 @@ private void SetCurrentTrack(string songName)
         _musicAudioSource.volume = 0f; // Always set volume to 0 BEFORE playing the audio source
         _musicAudioSource.Play();
     }
+
+    public static AudioClip GetSfxResource(string soundEffectName)
```
PhoebeMitchell: nit: Capitalise SFX to be consistent with PlaySFX method name.
Or better: Rename PlaySFX to PlaySfx to follow C# style guidelines. You could change _settingsSFXVolume as well if you do that.

Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -145,4 +154,9 @@ private void SetCurrentTrack(string songName)
...
+    {
+        return Resources.Load("Audio/SFX/" + soundEffectName) as AudioClip;
```
PhoebeMitchell: nit: Use generic Resources.Load<AudioClip> instead of casting.

Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -92,7 +92,16 @@ private AudioSource CreateAudioSource(string gameObjectName)
     /// <param name="soundEffectName">Name of sound effect asset, must be in `Resources/Audio/SFX`</param>
     public void PlaySFX(string soundEffectName)
     {
-        AudioClip soundEffectClip = Resources.Load("Audio/SFX/" + soundEffectName) as AudioClip;
+        AudioClip soundEffectClip = AudioController.GetSfxResource(soundEffectName);
```
PhoebeMitchell: nit: I personally dislike qualifying static methods from the same class, mostly because they have a horrible shade of grey in Rider.

Assets/Scripts/UI/AppearingDialogueController.cs
```diff
@@ -9,6 +9,12 @@ public class AppearingDialogueController : MonoBehaviour, IAppearingDialogueCont
 {
     [Tooltip("Drag a DirectorActionDecoder component here.")]
     [SerializeField] private DirectorActionDecoder _directorActionDecoder;
+
+    [Tooltip("Drag a AudioController here.")]
```
PhoebeMitchell: nit: *an AudioController

Assets/Scripts/UI/AppearingDialogueController.cs
```diff
@@ -54,6 +61,7 @@ private void Awake()
     {
         _textInfo = _textBox.textInfo;
         _directorActionDecoder.Decoder.AppearingDialogueController = this;
+        this._defaultDialogueChirpSfx = AudioController.GetSfxResource("maleTalk");
```
PhoebeMitchell: nit: Unnecessary "this" qualifier

PhoebeMitchell: I know we talked about avoiding the Unity editor but could this be assigned in editor? Especially as _audioController is also being assigned in editor.

Assets/Scripts/UI/AppearingDialogueController.cs
```diff
@@ -147,6 +177,16 @@ public float GetDelay(char character)
         var pair = _punctuationDelay.FirstOrDefault(charFloatPair => charFloatPair.Item1 == character);
         return pair.Item1 == '\0' ? DefaultPunctuationDelay : pair.Item2;
     }
+
+    /// <summary>
+    /// Helper method, does what it says.
```
PhoebeMitchell: nit: Kind of a redundant summary. Could be improved or removed.

Assets/Scripts/UI/AppearingDialogueController.cs
```diff
@@ -147,6 +177,16 @@ public float GetDelay(char character)
...
+    /// </summary>
+    /// <param name="character"></param>
+    /// <returns></returns>
```
PhoebeMitchell: Add return value and parameters

Assets/Scripts/UI/AppearingDialogueController.cs
```diff
@@ -124,6 +133,27 @@ private IEnumerator PrintTextCoroutine(int startingIndex)
...
+    /// Play dialogue chirp sound effect for current actor, if it exists
+    /// </summary>
+    /// <param name="currentActor">Speaker actor</param>
+    /// <param name="currentCharacter">Character to play chirp on (skipped if punctuation or ignored)</param>
+    private void PlayDialogueChirp(ActorData currentActor, char currentCharacter)
+    {
+        if (currentActor != null)
+        {
+            if (!IsPunctuationOrIgnored(currentCharacter))
+            {
```
PhoebeMitchell: nit: if statements could be combined and inverted

Assets/Scripts/AudioController/AudioController.cs
PhoebeMitchell: Does this need to be static? AppearingDialogueController already has a reference to AudioController, and anything else that uses this method will also have a reference to AudioController so they can use the PlayMusic or PlaySFX methods.

Assets/Scripts/UI/NameBox.cs
```diff
@@ -22,7 +22,13 @@ public void SetSpeaker(ActorData actorData)
...
 
+        CurrentActor = actorData;
         _text.text = actorData.DisplayName;
         _image.color = actorData.DisplayColor;
     }
+
+    /// <summary>
+    /// The most recently set ActorData, reflects the name currently shown in the NameBox.
+    /// </summary>
+    public ActorData CurrentActor { get; private set; }
```
PhoebeMitchell: nit: Move this to top of class, after private fields to follow C# style guidelines.

PhoebeMitchell (dismissed):
Looking good! Gotta love setting the sound to airGuitar and playing the text at full speed.

Assets/Scripts/UI/AppearingDialogueController.cs
notexplosive: Good eye!

Assets/Scripts/AudioController/AudioController.cs
notexplosive: My thinking is that this method does some string concatenation with the parameter and then talks to the `Resources` API. Neither of those things require an instance to accomplish. 

My belief is that if it _can_ be static (without anti-pattern trickery like singletons) I see no reason why it shouldn't be.

If you feel strongly about it I can un-static it. Not dying on this hill.

Assets/Scripts/UI/AppearingDialogueController.cs
```diff
@@ -140,18 +141,17 @@ private IEnumerator PrintTextCoroutine(int startingIndex)
...
-            if (!IsPunctuationOrIgnored(currentCharacter))
-            {
-                var chirp = currentActor.DialogueChirp;
-                if (chirp == null)
-                {
-                    chirp = this._defaultDialogueChirpSfx;
-                }
-                _audioController.PlaySFX(chirp);
-            }
+            chirp = this._defaultDialogueChirpSfx;
```
PhoebeMitchell: nit: unnecessary "this". Forgot to comment on this before

Assets/Scripts/AudioController/AudioController.cs
PhoebeMitchell: Don't feel _that_ strongly about it, considering it's just the Resources.Load method which is static anyway.
Could be private though, now nothing else is using it.

PhoebeMitchell (dismissed):

Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -90,9 +92,18 @@ private AudioSource CreateAudioSource(string gameObjectName)
     /// Plays sound effect of desired name.
     /// </summary>
     /// <param name="soundEffectName">Name of sound effect asset, must be in `Resources/Audio/SFX`</param>
-    public void PlaySFX(string soundEffectName)
+    public void PlaySfx(string soundEffectName)
```
lyndsiWilliams, line 95: This method and the one below (ln 105) are both named `PlaySfx` identically, could that mess something up? Shouldn't they be named differently?

Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -141,8 +152,13 @@ private bool IsCurrentlyPlayingMusic()
     /// <param name="songName">Name of song asset, must be in `Resources/Audio/Music`</param>
     private void SetCurrentTrack(string songName)
     {
-        _musicAudioSource.clip = Resources.Load("Audio/Music/" + songName) as AudioClip;
+        _musicAudioSource.clip = Resources.Load<AudioClip>("Audio/Music/" + songName);
         _musicAudioSource.volume = 0f; // Always set volume to 0 BEFORE playing the audio source
         _musicAudioSource.Play();
     }
+
+    private static AudioClip GetSfxResource(string soundEffectName)
```
lyndsiWilliams, line 165: This method is missing a summary ✨

Assets/Scripts/UI/AppearingDialogueController.cs
```diff
@@ -124,6 +134,26 @@ private IEnumerator PrintTextCoroutine(int startingIndex)
...
+    /// <param name="currentActor">Speaker actor</param>
+    /// <param name="currentCharacter">Character to play chirp on (skipped if punctuation or ignored)</param>
+    private void PlayDialogueChirp(ActorData currentActor, char currentCharacter)
+    {
+        if (currentActor == null || CharShouldBeTreatedAsPunctuation(currentCharacter))
+        {
+            return;
+        }
+        
+        var chirp = currentActor.DialogueChirp;
```
lyndsiWilliams, line 149: Does `currentActor.DialogueChirp` have a type it can be defined with? Or is it a weird changey-type that works better with `var`?

lyndsiWilliams (commented):
I have just a couple of questions, also I found a missing summary 😁

Assets/Scripts/AudioController/AudioController.cs
notexplosive: Yep! It's entirely safe to have method overloading in C#. Although their names are identical they take different parameters so the compiler knows which one you're calling.

`PlaySfx("talk")` will call the `PlaySfx()` that takes a `string`.
`PlaySfx(someAudioClip)` will call the `PlaySfx()` that takes an AudioClip (assuming `someAudioClip` is an AudioClip)

Assets/Scripts/UI/AppearingDialogueController.cs
notexplosive: The short answer is "yes, it's an AudioClip, I can just define it as an AudioClip." I'm happy to do that, but there's a deeper answer here. If you don't want a C# lecture then feel free to ignore, but maybe helpful!

I think you're misunderstanding how `var` works (it's not like JavaScript where types can change on a given variable). `var` is resolved at _compile time_. So once you assign a type using `var` the type is defined and known to the language (if you hover over it in your IDE it should tell you it's an AudioClip. I'm not sure if VS Code does that but VisualStudio 2017 does, and so does Rider).

In JavaScript I can get away with:

```js
// JavaScript (and probably TypeScript)
var x = 55; // x is any type
x = "hello world"; // yep, a-OK in JavaScript land
```

```cs
// C#
var x = 55; // x is an integer because 55 is an integer
x = "hello world"; // Nope! Compile error! Red squigglies for you!
    ~~~~~~~~~~~~~

var d = 1.0; // d is a double because 1.0 is a double
var f = 1f; // f is a float because 1f is a float
var c = SomeFunctionThatReturnsACat(); // c is a Cat because the function returned a Cat
```

So var in C# is _just_ a convenient shorthand, it doesn't actually cause any superpowers in the language (at least not that I'm aware of?). 

Although in C++ there's a thing called `auto` which works kind of like `var` that can do some dark magic.

Assets/Scripts/AudioController/AudioController.cs
lyndsiWilliams: Oh that's awesome, it's smart so it just knows which method to use depending on the parameter! Thanks for wrinkling my brain a little more with this 😁

Assets/Scripts/UI/AppearingDialogueController.cs
lyndsiWilliams: Ooo so it'll just know what the type it is depending on how it's applied at compile time? Would that ever affect performance?

It doesn't show me the type when I hover over `var` in VScode, that's why I figured I'd request a description on these. Maybe there's a plugin or something I'm missing, it doesn't show with my current ones though. I assumed this was an AudioClip or something similar but wasn't sure myself, I was thinking it'd be better for readability to define it.

Overall though I'm fine with leaving it as `var` as long as it doesn't affect performance, I can deal if I'm the only one that doesn't get the magic onHover-type-definition-functionality ✨

ViMaSter: Did somebody say C# lesson? 👀

> So var in C# is just a convenient shorthand, it doesn't actually cause any superpowers in the language (at least not that I'm aware of?).

**Absolutely correct!** It only tells the compiler "you figure out what I'm using here and **after that** make sure it stays that type" . It's also not possible to use it as a parameter for that reason:
```cs
Foo(var a); // this will not compile
```
From what I've seen, modern codebases usually default to `var` instead of explicit types. 
The only performance impact would be during compile-time, so nothing an end-user will notice. The compiler needs to validate that, whatever you're assigning has the correct type anyway, whether or not there's already a type definition present i.e. `float a = 1.0f` or it's used later on and the compiler needs to assign one itself `var a = 1.0f; foo(a);`. If there's an impact to compiling, I'd consider it negligible. (We're talking µs. : D) 

> It doesn't show me the type when I hover over var in VScode, that's why I figured I'd request a description on these.

An editor **can** still figure out which type it will be, but VSCode's language server is still pretty basic. I'm afraid I don't know of any plugins that can extend this functionality, but I also only use VS Code for web stuff or via github.dev; VS Community (or maybe ReSharper for VS Community) actually does what you're asking: 
image

That said: There actually is a rare scenario where explicit types make sense.  
Perhaps you've seen methods with "template arguments" like
```cs
JsonConvert.DeserializeObject<float>("1.0");
JsonConvert.DeserializeObject<int>("1.0");
JsonConvert.DeserializeObject<List<int>>("[1, 2, 3, 4]");
```
It's possible to skip the template argument (the text inside `<>`), if the usage implies it. For example:
```cs
  var a = JsonConvert.DeserializeObject<float>("1.0");   // a will be of type float
float b = JsonConvert.DeserializeObject("1.0");          // b will be of type float
float c = JsonConvert.DeserializeObject<float>("1.0");   // c will be of type float, but <float> is redundant
  var d = JsonConvert.DeserializeObject("1.0");          // this line will compile, BUT d will be of type "object" instead

// given...
public void Foo(float A) { ... }

Foo(d); // will not compile! as `d` is of type `object`...
Foo((float)d); // ...and would needed to be cast first
```

Making something of type `object` (which is a shorthand for `System.Object`), allows you to assign **anything** to it, as **literally anything** in C# is derived from `System.Object`:
```cs
// this is valid C# code
object a = "abc";
a = 1.0f;
a = 1;
```
However, this is generally a bad idea:  
While you cannot cause memory leaks or cause a .NET app to crash by downcasting in C#, it hides information. To reference [another convo](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/135/files?authenticity_token=NuUUUI57gPX0wetyn13VBB3zsEWeKt%2FYZDpM8d%2BKnPguyqkFQbXRduMumFW2B7fElwLoTjGEoDagWshGplRQkw%3D%3D&file-filters%5B%5D=.cs#r776929909): You could have a method accept `Foo(object A)` and then cast `A` to a `string` or `AudioClip` in your method body. But: This hides your intent.  
If you ever decide to use `Foo()`, your editor of choice, cannot provide with you with any info regarding parameter `A`. You might pass a `float`, thinking it's perfectly fine, the method would TECHNICALLY be entered, but not actually able to do anything. 

---

There's one other type, which is [`dynamic`](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/types/using-type-dynamic).  
`dynamic` will cause behavior you might be familiar with from JavaScript: Code that compiles fine, but actually crashes, once your machine actually attempts to execute it.  
C# is _generally_ considered type-safe. If an object is of a type (say `float`), if you're able to access it, it's able to do anything a `float` can. The only exception being, if the field is initialized with `null`.
That means, the compiler can make assumptions about what a variable can do, before the code is ever executed, as it's either of that declared type (`float`) or not initialized.
With `dynamic` this check is bypassed. C# is forced to wait until the line of code is actually executed, to see if it's a legal operation or not.  
This _can_ have advantages, but always comes with the disadvantage of **shifting** the discoverability of issues with it **right**.
Given this _very-work-such-presentation-wow_ slide:
image
The earlier something can be tested and related issues detected and fixed, the cheaper it is. (Imagine noticing a broken brake, after producing 2000 cars.)  
This is actually one of the main reasons TypeScript is gaining popularity: JavaScript makes it easy to do anything. But TypeScript is what makes doing the wrong stuff hard.

notexplosive: Vim coming here with the college lecture hot dang!

To answer Amythia's earlier question:

> Would [using `var`] ever affect performance?

Short answer: Nope!

Longer answer: Hmm... I think the compiler would need to think a tiny bit harder but the builds are almost instant anyway. If there were a perceptible perf cost it would only be felt on our side when we build the game, it would not affect players.

notexplosive: Amythia-- once you've read this convo can you either:
- resolve it for me (leaving it unresolved so you can get the benefits of Vim's lecture :) )
- OR: comment if you still want me to use `AudioClip` instead of `var`. I haven't made the change yet but if you'd still rather I use the exact type I can make that change.

lyndsiWilliams: Wow times a thousand, thank you for all this! Now I can very comfortably say that I am fine with `var` here as it is, it'll impact _very little_. And now I know why! 🧠 😁

lyndsiWilliams (approved):
Everything looks great! 😁

PhoebeMitchell (approved):

Issue #128, opened 2021-12-30
[BUG] "None" and "Unknown" actors not recognised


PhoebeMitchell:
## Describe the bug
With changes made in #103, the None and Unknown actors no longer work.

## Steps To Reproduce
Find a script with the actions &ACTOR:None or &ACTOR:Unknown and see how they produce an error.

## Expected behavior
&ACTOR:None should hide the namebox, and &ACTOR:Unknown should set the namebox text to ???

## Additional context
Maybe best to add these as edge cases to the ActorController (i.e. if (actor == "None")).
This would avoid having to add "None" and "Unknown" actors to the actor list scriptable object.

notexplosive:
What is the namebox color for "Unknown"? Is it always the same or is it the color of the associated actor? Maybe there should be a new `&SPEAK` command like `&SPEAK_UNKOWN:Tutorial_Boy` which means: "Tutorial boy is speaking, but the player doesn't know that" so we get all of Tutorial Boy's metadata (his dialogue chirp, namebox color, etc) _except_ his name.

PhoebeMitchell:
In the Lua build it appears to use the same colour as Dan, as seen when Jory is introduced, and when Tutorial Boy is introduced. This could be because Dan is the last person to speak in both of those instances but after Brent speaks it is still blue. Reference:
https://youtu.be/Skop4WFElhE?t=977
Maybe best to ask the designers. I would agree it would make sense to have it be the same colour as the associated actor.

notexplosive:
![image](https://user-images.githubusercontent.com/6920005/147788347-717133ba-b219-4cac-9497-53d63432383d.png)
Designers sound on-board with the box being the color of the speaker

PR #136, opened 2021-12-31
Bugfix/128 None and Unknown Actors

notexplosive:
## Summary
This PR adds two new commands (that need to be added to the Wiki just before this PR goes through). Technically this is a bug fix because these things worked before. But now they work again (maybe) even better-er than before!

- `&SPEAK_UNKNOWN:<actorName>` which is similar to `&SPEAK:<actorName>` except the name is obscured as `???`, but we keep the NameBox color (see gif)

- `&NARRATE` which is equivalent to `&SPEAK:None` this disables the NameBox, typically used for narration.

unknown-name-2

## Testing instructions
I modified `AppearingDialogInk` to test out my changes. But feel free to write up your own ink using `&SPEAK_UNKNOWN` and &NARRATE`

## Additional information

- `[kinda]` Changes UI
- `[kinda]` Introduces new feature
#128


Assets/Resources/InkDialogueScripts/AppearingDialogInk.ink
```diff
@@ -1,3 +1,22 @@
+&NARRATE
```
PhoebeMitchell, line 1: Could you move this file and the JSON file to the TestScripts folder? I know all the other scripts aren't in there yet, but may as well make a start.

Assets/Scripts/Actor/ActorController.cs
```diff
@@ -37,6 +44,9 @@ private void Start()
         }
 
         _directorActionDecoder.Decoder.ActorController = this;
+        
+        _onNewSpeakingActor += _nameBox.SetSpeaker;
+        _onNarrate += _nameBox.SetSpeakerToNarrator;
```
PhoebeMitchell: What's the reasoning behind making these events? Could they not be called directly when needed?

Assets/Scripts/UI/NameBox.cs
```diff
@@ -18,11 +18,23 @@ private void Awake()
...
         gameObject.SetActive(actorData.DisplayName != "");
+        var nameToDisplay = actorData.DisplayName;
 
-        _text.text = actorData.DisplayName;
+        if (speakingType == SpeakingType.SpeakingWithUnknownName)
+        {
+            nameToDisplay = "???";
+        }
+
+        _text.text = nameToDisplay;
```
PhoebeMitchell: Could this be done with a conditional operator? (_text.text = speakingType == SpeakingType.SpeakingWithUnknownName ? ...)

Assets/Scripts/UI/NameBox.cs
```diff
@@ -18,11 +18,23 @@ private void Awake()
     /// Sets the text and color of the name box to that of the specified actor
     /// </summary>
     /// <param name="actorData">An ActorData object to get the name and color from</param>
-    public void SetSpeaker(ActorData actorData)
+    /// <param name="speakingType">The speaking type that is in use</param>
+    public void SetSpeaker(ActorData actorData, SpeakingType speakingType)
     {
         gameObject.SetActive(actorData.DisplayName != "");
```
PhoebeMitchell: Maybe replace this with just SetActive(true) now

Assets/Scripts/Actor/ActorController.cs
```diff
@@ -185,18 +195,27 @@ public void PlayEmotion(string emotion, string actorName = null)
...
     {
         try
         {
-            _currentSpeakingActor = _actorInventory[actor];
-            _onNewSpeakingActor.Invoke(_currentSpeakingActor);
+            if (actorName == NarratorActorName)
+            {
+                _currentSpeakingActor = null;
+                _onNarrate?.Invoke();
+            }
```
PhoebeMitchell: This could be moved to its own method which is called by ActionDecoder. This would save having to deal with passing NarratorActorName back and forth. Unless the idea was to keep &SPEAK:None as a valid action? Personally I think it should be one or the other.

PhoebeMitchell (dismissed):
Two PRs in one night! Smashing it! 👯‍♀️

Assets/Scripts/Actor/ActorController.cs
ViMaSter: The more I think about it, I actually find myself agreeing with moths proposal here of **either** using `&SPEAK:None` **or** having a `&NARRATE`, but not supporting both.  

Instead of hardcoding an actor name for the narrator, could we introduce a third `SpeakingType` for the narrator instead?  
This would also remove the need to [expose `NarratorActorName` inside the `IActorController`-interface](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/136/files#diff-441524bcf9bcee899df034a0b01c2398f6492bce70e53bf3ab8159930ffcb5cdR20), which feels more like an implementation detail to me, which other classes shouldn't be concerned about, as they could simply react to set `SpeakingType` instead.

Assets/Scripts/UI/NameBox.cs
ViMaSter: I personally try to avoid conditional operators. Similarly to early-returns, I usually find having a default and overriding it only in certain scenarios (as done here) to be more readable.  
That way, if we do want to introduce another special case later on, we'd just have to add another `if() { return; }` block, instead of expanding the conditional operator (which, when not done automatically can unintentionally introduce bugs).

ViMaSter (dismissed):
Added votes and comments to already existing threads, but nothing worth blocking from my end either, well done! 👍

Assets/Scripts/UI/NameBox.cs
notexplosive: Leaving this as-is. Although I like ternary it feels too cumbersome here (maybe because the names are too long?)

Assets/Scripts/Actor/ActorController.cs
```diff
@@ -25,6 +28,8 @@ public class ActorController : MonoBehaviour, IActorController
     private ActorData _currentSpeakingActor;
     private SpeakingType _currentSpeakingType = SpeakingType.Speaking;
 
+    public string NarratorActorName { get; } = "None";
```
ViMaSter: Nit: With the latest changes this is now unused and could be removed.

ViMaSter (dismissed):
One "unused property"-nit (there's always one : D), aside from that it looks pretty solid now! 👍

PhoebeMitchell (dismissed):

ViMaSter (dismissed):
Re-approving after the `develop` integration merge.

PhoebeMitchell (dismissed):

Assets/Scripts/UI/NameBox.cs
```diff
@@ -23,12 +23,24 @@ private void Awake()
...
+        _text.text = nameToDisplay;
         CurrentActor = actorData;
-        _text.text = actorData.DisplayName;
         _image.color = actorData.DisplayColor;
     }
+
+    public void SetSpeakerToNarrator()
+    {
+        gameObject.SetActive(false);
+    }
```
PhoebeMitchell: Approved then found a bug 😅 This method doesn't set the CurrentActor property. If NARRATE is called before SPEAK then no text sound will play. If NARRATE is called after SPEAK then the sound of the previous text sound will play. Maybe replace the current actor property with a GetActorSound method which returns either the actors chirp sound, or a default sound?

PhoebeMitchell (commented):
Another bug, TextBoxCanBeHidden test fails on AppearingDialogueControllerTests. This is because the narrative script in Inky-TestScene disables the NameBox so it can't be found.
HOWEVER this test is supposed to test if the speech panel is disabled, not the name box. So probably should change the Find method to find the SpeechPanel object instead.

Assets/Scripts/UI/NameBox.cs
notexplosive: Thanks! looking at this now

FYI- Fixed the broken test mentioned in your below comment (github won't let me reply to that comment so I'm writing it here)

notexplosive: ~~I can't repro this bug because I don't hear dialogue chirps at all-- are you testing a combination of both my PRs?~~

wait i'm dumb hang on

notexplosive: Narrator will now always use the default chirp sound. We'll probably want to change this but there's some extra design work there (how will the dialogue script define that?)

That can be part of this PR if we have an idea we like now-- or it can wait until later and I can open an issue.

Assets/Scripts/UI/AppearingDialogueController.cs
```diff
@@ -137,16 +137,16 @@ private IEnumerator PrintTextCoroutine(int startingIndex)
...
     {
-        if (currentActor == null || CharShouldBeTreatedAsPunctuation(currentCharacter))
+        if (currentActorChirp == null || CharShouldBeTreatedAsPunctuation(currentCharacter))
         {
             return;
         }
         
-        var chirp = currentActor.DialogueChirp;
+        var chirp = currentActorChirp;
         if (chirp == null)
```
PhoebeMitchell: Bottom if statement is never reached if currentActorChip is null

PhoebeMitchell (changes requested):
The default chirp does not play 😭

Assets/Scripts/UI/NameBox.cs
PhoebeMitchell: I'm happy with just having one default sound for now. We can add it in future if needed.

Assets/Scripts/UI/AppearingDialogueController.cs
notexplosive: whoops.

PhoebeMitchell (approved):
All looks good now 👍

ViMaSter (approved):
Test run, so LGTM!

PR #137, opened 2022-01-01
feature/130-Speech-box-continue-arrow

### LINKED ISSUE #130: Speech box continue arrow
## Description
Add this little arrow to the speech panel. Use an Animator component to animate it.
Screenshot 2021-12-30 at 04 56 58

Wait for #103 to be accepted before starting this feature.

## Acceptance Criteria

This story will be finished when the arrow has been added.

PhoebeMitchell:
## Summary

Adds the continue arrow to the SpeechPanel prefab. Hooks it up to the correct events in StandardInclusion prefab to make it disappear and appear when text starts and ends respectively.

continuearrow

## Testing instructions

Open Inky-TestScene and play a script. The arrow should appear at the end of every line of text

## Additional information

- `[x]` Changes UI
  
closes #130


Assets/Animations/UI/ContinueArror.controller
```diff
@@ -0,0 +1,12 @@
+%YAML 1.1
```
ViMaSter: Hmm... For some reason Unity flags this as an invalid Animation Controller for me and prints this instead of playing the animation: (running Windows 10, 19043)
image
Wanted to check if it also happens [in the build for this PR](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/suites/4803380329/artifacts/134837133), but I'm stuck on an animating Arin sprite with no text box, if I select `New Game`.
image

Nit: This file is also got a typo: `ContinueArro>r<.controller`

ViMaSter (changes requested):
Hm... It seems like the animation is actually broken for me. I've put a bit more info in the thread below, blocking this merge for now, until that's resolved.

Assets/Animations/UI/ContinueArror.controller
notexplosive: I fetched and checked out this branch and I don't see the warning message you're seeing. The arrow shows up fine for me.

notexplosive: When I load the build I also just see Arin in a black void... but I'm not sure what the old behavior was. Investigating...

If I walk the same steps in the editor I get the same result. Because Main Menu takes you to "Transition Test" which is just Arin in a black void. Maybe that's a separate bug we should fix?

notexplosive (dismissed):
Since this change has zero .cs changes, I don't really know how to review it (my least favorite part of using Unity). But it runs correctly so I'll assume it's correct.

Assets/Animations/UI/ContinueArror.controller
PhoebeMitchell: This problem should be fixed in latest commits

ViMaSter (dismissed):
Works now, thanks for addressing that Animator issue!

---

Nit: I randomly had the `AppearingDialogInk` script active and on the "This text appears instantly" line no arrow is shown:
image

On the line before and after that it's happily bopping though:
SPA9boMi5E

Not worth blocking though, it's just an inconsistency. Unless it actually a bug, in which case either make a new issue or fix it and I'll re-test and approve! : D

ViMaSter (changes requested):
Sorry for having to hand it back again; the arrow in the `Inky-TestScene` is completely gone now. : D
On 2948c85b920ddf15c4f2dca196e291ff42ca796e (the latest commit of this branch) it's gone:
image

On 6928f799fd0e11b452a8e41c30234a212ec4ece6 (commit I tested on for my last approval):
image

The behavior is consistent across `AddRecordTest` and `AppearingDialogInk` in both versions.

Assets/Animations/UI/ContinueArror.controller
ViMaSter: Also [created a bug report](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/issues/143) for the main menu behavior I encountered.

ViMaSter (approved):
Seems like that last commit did it; works for me as well now! :)

notexplosive (approved):
Rocket Emoji

PR #138, opened 2022-01-01
Bugfix/120 emotions can't be played on actors multiple times in a row

### LINKED ISSUE #120: [BUG] Emotions can't be played on actors multiple times in a row
## Describe the bug
If PLAY_EMOTION is used to play an emotion on an actor, then used a second time, it will not play again. Because of this the script gets stuck and cannot continue.

## Steps To Reproduce
Make a narrative script where PLAY_EMOTION is used twice in a row.
Play the script and see how it gets stuck.

## Expected behavior
You should be able to repeat the emotions as many times as you want.

## Additional context
Animator's Play method has a normalizedTime parameter. Maybe try explicitly setting this to 0 when calling Play.

PhoebeMitchell:
## Summary

Fixes a bug where emotions could not be played multiple times in a row on an actor, due Animator's Play method not restarting animations that are already playing.

Look at him go!
multipleemotions

## Testing instructions

Play a narrative script that uses the PLAY_EMOTION action multiple times in a row (such as the provided MultipleEmotionsTest) and witness the animations playing properly.

## Additional information

- `[ ]` Changes UI
  
closes #120


ViMaSter (approved):
Quick and simple fix that works. Also always love a dedicated test script for a quick test run; real nice! 👍

notexplosive (approved):
I'm glad it was easy but the fact that the fix involved adding a `0, 0)` to an API call causes Angry-Unity-Noises for me

Issue #114, opened 2021-12-30
[BUG] Cannot present evidence when PRESENT_EVIDENCE action is called


PhoebeMitchell:
## Describe the bug
When PRESENT_EVIDENCE is called the evidence menu appears but you cannot present evidence.

## Steps To Reproduce
Find a script with the PRESENT_EVIDENCE action.
Try to select evidence to present.

## Expected behavior
You should be able to select evidence to present. Presenting the correct evidence should continue the story. Presenting the wrong evidence should give a failure state.

## Additional context
Presenting evidence only works in cross examination mode. It requires an Ink choice. HOWEVER when PRESENT_EVIDENCE is called it opens the evidence menu which prevents the story from continuing so the choice can never be reached, meaning you cannot select evidence.
In the screenshot the story pauses on PRESENT_EVIDENCE so the choice is not reached.
<img width="281" alt="Screenshot 2021-12-30 at 03 28 47" src="https://user-images.githubusercontent.com/67964179/147719425-a495338c-79f7-4821-bdbc-e801b8994037.png">

PR #142, opened 2022-01-02
Refactor proposal: Deduct game mode from .ink script

ViMaSter:
## Summary
To facilitate the idea of "1 scene per episode", Unity needs to be able to deduct the game state and find all required resources **only** by parsing a .ink-file. 
This PR moves one still existing dependency from the Unity editor to the `.ink` file: the `DialogueControllerMode`, which determines if a scene is a cross examination or pure dialogue.

(Also as this was referred to as `ScriptType`, `DialogueMode` and `DialogueControllerMode` in various places across the codebase, I have unified this to `GameMode` [since a change of this variable effectively changes the entire rules of the game] but I am open to other suggestions too.)

Every .ink-script is now required to start with the `&MODE:` action line set to either `&MODE:CrossExamination` or `&MODE:Dialogue`.
This PR updates all .ink-files to `Dialogue`, except for `Ross_Cool_X.ink` which set to `CrossExamination`.
Also automated tests were added, that ensure human-readable error messages are printed if an `.ink`-file does not meet this new requirement.

Before: 
image
After: 
image

## Testing instructions

1. Run the newly added `Dialogue` tests 
2. Run any previously working scene (I've used `CrossExamination - Test Scene` and `Evidence Menu - Test Scene`)
3. Ensure the correct UI elements are shown and the game still behaves like before this change

## Additional information

- `[ ]` Changes UI


Assets/Scripts/TextDecoder/Dialogue.cs
```diff
@@ -1,3 +1,5 @@
+using System;
+using Ink.Runtime;
 using UnityEngine;
 
 [System.Serializable]
```
PhoebeMitchell: nit: System qualifier can be removed

Assets/Scripts/TextDecoder/Dialogue.cs
```diff
@@ -7,16 +9,36 @@ public struct Dialogue
...
     
-    [field: Tooltip("The dialogue mode the narrative script will use (dialogue or cross examination).")]
-    [field: SerializeField] public DialogueControllerMode ScriptType { get; private set; }
+    public GameMode ScriptMode
+    {
+        get
+        {
+            const string expectedFileStart = "&MODE:";
+            var firstLine = new Story(NarrativeScript.text).Continue().Trim();
+            var availableModes = ((GameMode[])Enum.GetValues(typeof(GameMode)));
```
PhoebeMitchell: nit: outer parentheses can be removed

Assets/Scripts/TextDecoder/Dialogue.cs
```diff
@@ -7,16 +9,36 @@ public struct Dialogue
...
         NarrativeScript = narrativeScript;
-        ScriptType = type;
     }
 
     [field: Tooltip("Drag an Ink narrative script here.")]
     [field: SerializeField] public TextAsset NarrativeScript { get; private set; }
     
-    [field: Tooltip("The dialogue mode the narrative script will use (dialogue or cross examination).")]
-    [field: SerializeField] public DialogueControllerMode ScriptType { get; private set; }
+    public GameMode ScriptMode
```
PhoebeMitchell: Not an expert on when to use properties vs methods but this seems pretty complex for a property. Maybe it should a method instead?

PhoebeMitchell: Seems like this whole thing is so StoryController can see if a script is a CrossExamination. May I suggest instead the MODE action is used to switch between Dialogue and CrossExamination mode? Furthermore, if Dialogue mode is the default then this whole struct isn't required. The narrative scripts can just consist of TextAssets.

Assets/Scripts/TextDecoder/DialogueController.cs
```diff
@@ -18,13 +18,30 @@ public enum CrossExaminationChoice
     Evidence
 }
```
PhoebeMitchell: While we're editing this script, is this enum actually ever used? Ross_X_Cool seems to play OK without it

PhoebeMitchell (commented):
Nice to see a refactor from our resident C# expert 😄 A MODE action is actually a feature we might need for issue #114 because we currently need to be in cross examination mode to present evidence; the previous alternative would be switching to a tiny narrative script every time we need to present anything in game. With that in mind I've left some comments talking about streamlining it further, and of course some lovely nits.

Assets/Scripts/TextDecoder/DialogueController.cs
ViMaSter: That's true, but while we might not have used it yet, the Ace Attorney games definitely (and our engine already) allow the user to not just object or make a decision during cross examination, but also present a piece of evidence to further progress the examination.

But I'm not sure if we're supporting the full loop already, since I wouldn't know how to write an .ink-script that reacts onto a **specific** piece of evidence from the top of my head. Once I've gone over your other comments tomorrow, I'll see if I can get an example into this PR as well, so we're not carrying over dead code with this PR. Definitely well spotted in any case!

Assets/Resources/InkDialogueScripts/ActorTestInk.ink
```diff
@@ -1,3 +1,4 @@
+&MODE:Dialogue
```
notexplosive: What's the consequence for forgetting to add this line?

Presumably _most_ scripts will be Dialogue so having Dialogue as the default doesn't seem unreasonable.

Assets/Scripts/TextDecoder/Dialogue.cs
```diff
@@ -7,16 +9,36 @@ public struct Dialogue
...
+    {
+        get
+        {
+            const string expectedFileStart = "&MODE:";
```
notexplosive: Design concern: It seems like we're keying on the exact string `&MODE:`, which assumes that we will always use the "ampersand, all caps, colon" syntax that we've been using. Is there any way we can extract that so if we change from ampersand to (for example) dollar sign we don't need to remember to change this code?

As a random idea: what if MODE worked like a normal command and you could change modes mid-script? Or is that too crazy?

notexplosive (dismissed):
Thanks for doing this :)

Assets/Scripts/TextDecoder/Dialogue.cs
ViMaSter: > Design concern: It seems like we're keying on the exact string &MODE:, which assumes that we will always use the "ampersand, all caps, colon" syntax that we've been using. Is there any way we can extract that so if we change from ampersand to (for example) dollar sign we don't need to remember to change this code?

[Moths](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/142#discussion_r777155381) and you actually brought up two brilliant points here:
1. The only place where we have a dependency on "knowing what kind of mode a script is" was inside the `StoryController`, to call the `_onCrossExaminationStart`-event which was only used to `Initialize()` the `PenaltyManager`.
2. There's no good reason for one script to be limited to one mode.

Two changes here:
1. `StoryController._dialogueList` is effectively a "playlist" of scripts to run inside a given Unity scene (it's what makes 1-scene-per-episode possible). This is still the case, however it also previously manually fired a `_onCrossExaminationStart`-event for which it required to know the "game mode" of a given Ink script. Now that we have `&MODE:` [I have simply moved this logic to the `ActionDecoder`](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/142/commits/e4d3114d9c8e98ea4885c59210e2b7971be1ef96#diff-b31a4dc38266afe394a2db706c096fb7ecaf862c17ddd3e1a495a1c3cea7b49aR396-R399) and...

2. We generally prefer direct invocations over events, so I decided to "leave the ~~campsite~~ codebase cleaner than I found it": `_onCrossExaminationStart` is removed and the `StoryController` instead has a hard dependency on a `PenaltyManager` and calls `Initialize()` on it.  
  (I personally didn't see the need for a separate `IPenaltyManager`, but if you feel it'd be better to have it either right now or at a later point in time, I'm definitely open to it!)

Assets/Scripts/TextDecoder/Dialogue.cs
ViMaSter: Super good point! Since it matches well with a suggestion from Explosive [I've put most details there](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/142#discussion_r777207122), but regarding

> Furthermore, if Dialogue mode is the default then this whole struct isn't required. The narrative scripts can just consist of TextAssets.

While implementing this is a one-line change at this point, I personally actually like having each file begin with an explicit `&MODE:Dialogue` for legibility:  
Since we're aiming to have Unity be controlled by only a script, seeing a `&MODE:Dialogue` at the top of each file, helps whoever is editing or reading a script file to understand the full picture. Making it optional (/redundant) would add another bit of knowledge someone needs to just "keep in mind" while working on scripts or (even worse in my opinion) force writers to go into Unity to find a "magic default value" we developers set up.  
If they can just search a script for `MODE` and see every time a script is set to a specific mode, this uncertainty is immediately eliminated. Otherwise they'd have to leave their text editor, open up Unity, remember which script actually manages the game mode, find the game object which contains it in a given scene, just to see a simple `GameMode: Dialogue`.

Open to other ideas as usual! I just wanted to share my motivation behind the change. Feel free to either mark this as resolved or let me know if you'd be happier with an alternative.

Assets/Resources/InkDialogueScripts/ActorTestInk.ink
ViMaSter: > What's the consequence for forgetting to add this line?

`DialogueController.OnContinueStory()` would fail printing `Unhandled dialogue type` to the console.

> Presumably most scripts will be Dialogue so having Dialogue as the default doesn't seem unreasonable.

[Moths' brought up a similar point in this thread](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/142#discussion_r777208631), so I'll merge this into the other thread so we don't have the same discussion twice. : D

Assets/Scripts/TextDecoder/Dialogue.cs
ViMaSter: Was actually thinking the same thing when I wrote it. Due to [the latest changes](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/142#discussion_r777207122) this entire property was redundant anyway, so it's just gone completely now. :)

Assets/Scripts/TextDecoder/DialogueController.cs
ViMaSter: Welp, good catch:
image
VS was thinking at least the `Press`-value was used, but I assume IntelliSense simply mixed it up with the `Controls.PlayerActions.Press`, as when actually attempting to see the reference it's pointing to, it couldn't find any.🙃

`CrossExaminationChoice` isn't used anywhere else in the code either, so I've just removed it!

Assets/Scripts/TextDecoder/Dialogue.cs
PhoebeMitchell: > StoryController instead has a hard dependency on a PenaltyManager and calls Initialize() on it

As MODE primarily does stuff to the DialogueController maybe OnCrossExaminationStart should also be on the DialogueController. Though thinking about it, setting up the scene for a CrossExamination seems like a job for the SceneController. Maybe OnCrossExaminationStart should be renamed too as its no longer an event.

Assets/Scripts/TextDecoder/Dialogue.cs
PhoebeMitchell: I don't feel too strongly about it, but having a required &MODE at the top of a file is another thing the writers have to remember. There's nothing really special about Dialogue mode, its just the default way the game plays. CrossExamination mode is a special case that causes the system to handle Ink choices differently. TBH if we found a workaround for handling certain Ink choices differently we might not need a separate Dialogue and CrossExamination mode to begin with.

Assets/Scripts/TextDecoder/Dialogue.cs
ViMaSter: Yeah, I can see that. Not too happy with it either, but also don't have a strong enough an opinion on it, to actually make a change. : D

I'd say, let's just keep it until we actively run into advantages or disadvantages of it's current place when we implement other game modes, like the "free roaming evidence gathering" part?

Assets/Scripts/TextDecoder/Dialogue.cs
ViMaSter: > I don't feel too strongly about it, but having a required &MODE at the top of a file is another thing the writers have to remember.

Ah, yeah, that's a pretty reasonable way to look at it too! Since both @notexplosive and you [agree on that matter](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/142#discussion_r777160958) and I can see those points as well, I've made the "Dialogue" game mode the default now and removed the thereby redundant `&MODE:Dialogue` in https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/142/commits/f5d2b93995874ee3bf23853480d08031d34356ec.

If that works for the both of you, feel free to just upvote (depending on who of you first : D) and then close this conversation, unless you spot something else related to this!

Assets/Scripts/TextDecoder/Dialogue.cs
```diff
@@ -1,22 +1,18 @@
+using System;
 using UnityEngine;
 
-[System.Serializable]
+[Serializable]
 public struct Dialogue
```
PhoebeMitchell: This struct is essentially just a TextAsset at this point. Would it make sense to get rid of the struct altogether, and just use the TextAsset?

PhoebeMitchell (dismissed):
Good stuff! Just one comment left on the Dialogue struct

Assets/Scripts/TextDecoder/Dialogue.cs
PhoebeMitchell: Update: Been working on an object preloading system and converted this struct into a class (https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/feature/script-pre-loading/Assets/Scripts/TextDecoder/NarrativeScript.cs) so maybe it'd be ok to keep this as it is for now

ViMaSter: a) Ooo~, nice to see the refactoring to load things from Ink scripts alone is already in full effect. 👀 

b) Actually was planning to update it after your comment, but yeah, that's actually a good idea and reminded me of how [Explosive converted `strings` to `AssetNames`](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/95a7d486dda673ae7538bc20b90ff9d39a848176/Assets/Scripts/TextDecoder/AssetName.cs) if applicable:  
While the underlying data in both `string` and `AssetName`s is just "an array of characters", not every `string` is an `AssetName`; and similarly not every `TextAsset` is automatically a `Dialogue`. Having them as separate classes should make it super clear, what kind of data is expected when.  
(Sadly C# has no equivalent to C++'s `typedef`s. That would effectively make it a one-liner.)

Also gonna mark it as resolved then! :)

notexplosive (dismissed):
👩🏻‍🚀🚀

notexplosive (dismissed):
👩🏻‍🚀🚀💢‼

PhoebeMitchell (dismissed):

notexplosive (dismissed):
morjit

PhoebeMitchell (dismissed):

notexplosive (dismissed):

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -614,7 +619,22 @@ private void SET_ACTOR_POSITION(int oneBasedSlotIndex, ActorAssetName actorName)
         OnActionDone?.Invoke();
     }
     #endregion
+    #region DialogueController
+    /// <summary>Changes the game mode. (This decides how the user is able ot progress with the story.)</summary>
```
PhoebeMitchell: typo: ot -> to

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -614,7 +619,22 @@ private void SET_ACTOR_POSITION(int oneBasedSlotIndex, ActorAssetName actorName)
...
+    /// <param name="mode">Name of game mode to put the plazer in</param>
```
PhoebeMitchell: typo: plazer -> player?

PhoebeMitchell (dismissed):
Tests run. Just some typos.

PhoebeMitchell (dismissed):

notexplosive (dismissed):
I think it's good?

notexplosive (approved):

PhoebeMitchell (approved):

PR #141, opened 2022-01-01
Feature/47 normalize asset names

notexplosive:
## Summary
All assets now have NamesLikeThis where the first letter of each word is capital.

## Testing instructions
Pull down the branch and:
-Try to find a script/scene that's broken with these changes. Cross Exam (which feels the most complex) works on my machine
-Try to find an asset with a wrong name.

## Additional information

- `[ ]` Changes UI
- `[ ]` Introduces new feature
- `[ ]` Removes existing feature
- `[x]` Has associated resource: 
#47

PR #146, opened 2022-01-08
Feature/script pre loading

### LINKED ISSUE #108: [BUG] Cross examination lives are not disabled on cross examination loop exit
## Describe the bug
The cross examination lives remain visible when the cross examination loop is not active.

## Steps To Reproduce
Play a script in cross examination mode
Press C to press witness
Witness how the lives are still visible

## Expected behavior
The lives should go invisible when exiting the cross examination loop.

## Additional context
There is an event on the DialogueController called OnCrossExamination loop active which passes a bool. You should be able to hook this up to the IsActive property on the GameObject that holds the lives so it gets enabled and disabled whenever the event is called.

### LINKED ISSUE #109: [BUG] Playing two cross examination scripts in a row gives double the number of lives
## Describe the bug
When two narrative scripts are set to cross examination mode in the StoryController and they play one after the other, the more lives are added to the player's existing number of lives.

## Steps To Reproduce
Assign two narrative scripts to the StoryController and set them to cross examination mode.
Play the scene and play through the scripts
Witness how more lives are added when the next scene starts.

## Expected behavior
When a cross examination narrative script ends it should clean up after itself, and destroy any remaining lives.

## Screenshots
lives

## Additional context
DialogueController has an OnDialogueFinished event for when a narrative script ends. Maybe use this?

### LINKED ISSUE #110: [BUG] Cross examination lives appear in front of black screen
## Describe the bug
The cross examination lives appear in front of the black screen used to fade in and out.

## Steps To Reproduce
Play a script, that has the &FADE_IN action at the beginning, in cross examination mode.
Witness how the lives appear in front of the blackness.

## Expected behavior
The lives should be behind the black screen.

## Screenshots
blackness

## Additional context
Rearrange the objects in the hierarchy to change their order, and apply changes to the BaseCanvas prefab. Make sure any changes are reflected in the standard inclusion prefab by making a new scene, adding a standard inclusion prefab and checking everything is in the correct order.

PhoebeMitchell:
## Summary

Implements a feature that reads an Ink JSON file and loads any assets required. This is handled by a new NarrativeScript class which replaces the previous Dialogue struct. Scripts can access the loaded and stored objects through the NarrativeScript class' ObjectStorage property.

This feature replaces the previous method of assigning actors and evidence available in a scene. This will allow for the game to be expanded upon more easily without requiring extensive use of the Unity editor.

ActorData, Evidence, and AudioClip objects are supported. ~~Support for loading BGScenes is planned, but want to wait for the asset normalisation PR #141 to go through.~~ BGScene preloading is now also supported.

The evidence menu tests scripts broke so took the opportunity to rework them to be a lot faster, cleaner, and less reliant on Ink scripts and Unity Scenes as well a fixing a bug where the cursor was positioned in the wrong place to click some buttons.

~~Marked as draft until: #142 is merged~~

Updated to include #142. Also closes #108 closes #109 and closes #110 because it broke the previous penalty implementation, so included those fixes when updating it.

## Testing instructions

Play scripts in Inky-TestScene and make sure they still work.
Run the associated tests.

## Additional information

- `[ ]` Changes UI
- `[x]` Removes existing feature


Assets/Scripts/TextDecoder/Preloading/ObjectPreloader.cs
```diff
@@ -0,0 +1,203 @@
+using System;
+using UnityEngine;
+
+/// <summary>
+/// Assigned to an ActionDecoder in place of the usual controller interfaces.
+/// Action decoder calls the methods which are used to load any required objects.
+/// Objects are then stored in the assign ObjectStorage object.
+/// </summary>
+public class ObjectPreloader : IObjectPreloader
```
ViMaSter: I think the way the lines are parsed here is a bit heavily nested. I could see a more direct approach working:

Instead of having a separate `IObjectPreloader` which reacts to the `ActionDecoder` calling the controllers, it could be used more directly: *
- Create a new `ActionDecoderBase` class
- This `ActionDecoderBase` class is IDENTICAL to the current `ActionDecoder`, except...
  - the class is `abstract`
  - every `ACTION_NAME` method
    - is `abstract` too
    - has no implementation
    - is `protected`
    - (i.e. `protected abstract void ADD_EVIDENCE(string evidence);`)
- The `ActionDecoder` class
  - should be named `IngameActionDecoder` (or something along those lines)
  - derives from `ActionDecoderBase`
  - implements every `ACTION_NAME`-method like `ActionDecoder` does right now

Now, your code:
- creates a `PreloadingActionDecoder` which also derives from `ActionDecoderBase`
- implements every `ACTION_NAME`-method directly
- has no need for implementations of `I*Controller`-interfaces* 

---

*I believe moving to feature classes like `ActionDecoder`, `ShoutPlayer` or the `ObjectStorage` introduced with this PR is preferable over the "controller architecture" we're using at the moment.  
Given that, avoiding more dependencies on those interfaces, seems like a good thing and, additionally, your `ObjectPreloader` doesn't really care about a lot of the properties like `AutoSkip`, `TextBoxHidden`, etc. and wouldn't be tied to those either.

Assets/Scripts/TextDecoder/Preloading/ScriptReader.cs
```diff
@@ -0,0 +1,67 @@
+using System.Collections.Generic;
+using System.Linq;
+using Ink.Runtime;
+using Object = Ink.Runtime.Object;
+
+public static class ScriptReader
```
ViMaSter: Is there a reason to have this class and `NarrativeScript` split up? As the methods inside this class are exclusively used inside their tests and instances of `NarrativeScript`, I personally don't see any need to have them in their own utility class.

If a `NarrativeScript` accepts a `IObjectPreloader` inside its constructor, `ScriptReaderTests` could simply become `NarrativeScriptTests`.

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuMouseTests.cs
```diff
@@ -56,46 +41,77 @@ public IEnumerator MouseShouldHighlightEvidenceMenuItems()
...
-            yield return _inputTestTools.PressForFrame(_inputTestTools.Mouse.leftButton, 123);
-        
-            Transform outerMenuItem = GameObject.Find("EvidenceMenuItem").transform;
-            yield return _inputTestTools.SetMousePositionWorldSpace(outerMenuItem.position + (outerMenuItem.GetComponent<RectTransform>().rect.size.x / 2) * canvasScale);
-        
-            Assert.AreEqual(outerMenuItem.GetComponent<EvidenceMenuItem>().CourtRecordObject.DisplayName,
-                menu.SelectedButton.GetComponent<EvidenceMenuItem>().CourtRecordObject.DisplayName);
+            yield return SelectButton(incrementButton);
+            yield return LeftClick();
+            yield return SelectButton(firstMenuItem.transform);
```
ViMaSter: Yes. So much yes. This is abstract enough to not have to fuss with things like input devices, low-level enough to not over-complicate things, AND way easier to read. So nice. 👍

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuMouseTests.cs
```diff
@@ -56,46 +41,77 @@ public IEnumerator MouseShouldHighlightEvidenceMenuItems()
...
+        }
+
+        private EvidenceMenuItem GetFirstMenuItem()
+        {
+            return GetMenuItems().First(menuItem => menuItem.gameObject.name == "EvidenceMenuItem").GetComponent<EvidenceMenuItem>();
+        }
 
-            global::EvidenceMenu evidenceMenu = InputTestTools.FindInactiveInScene<global::EvidenceMenu>()[0];
-            Assert.True(evidenceMenu.isActiveAndEnabled);
+        private IEnumerator SelectButton(Transform tf)
```
ViMaSter: Nit: Maybe change this to `HoverOver`? Selecting can mean a lot and - on first instinct - I actually expected a click too, which turned out to be a separate method. (Which is good code-wise, but was unexpected for me, given the name.)

Also super nit: I personally don't like abbreviating symbols in code; for me, `transform` reads easier than `tf`.

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuTest.cs
```diff
@@ -0,0 +1,68 @@
...
+        [UnitySetUp]
+        public IEnumerator SetUp()
+        {
+            yield return EditorSceneManager.LoadSceneAsyncInPlayMode(SCENE_PATH, new LoadSceneParameters(LoadSceneMode.Additive));
+            EvidenceController = Object.FindObjectOfType<EvidenceController>();
+            EvidenceMenu = TestTools.FindInactiveInScene<global::EvidenceMenu>()[0];
+            Menu = EvidenceMenu.GetComponent<Menu>();
+            CanvasTransform = Object.FindObjectOfType<Canvas>().transform;
+            var dialogueController = Object.FindObjectOfType<global::DialogueController>();
+            yield return TestTools.WaitForState(() => !dialogueController.IsBusy);
```
ViMaSter: Super duper nit: I think an empty line before and after (yield) `return`s makes would make this a bit more readable.

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuKeyboardTests.cs
```diff
@@ -1,195 +1,187 @@
 using System.Collections;
+using System.Collections.Generic;
 using System.Linq;
 using NUnit.Framework;
-using Tests.PlayModeTests.Scenes.MainMenu;
 using Tests.PlayModeTests.Tools;
 using TMPro;
+using UnityEditor.SceneManagement;
 using UnityEngine;
+using UnityEngine.SceneManagement;
```
ViMaSter: Nit: These namespaces aren't used in this file and should be safe to remove:
- `System.Collections.Generic`
- `Tests.PlayModeTests.Tools`
- `UnityEditor.SceneManagement` (actually added twice)

Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -68,13 +71,23 @@ void Update()
...
+    {
+        AudioClip song = _dialogueController.NarrativeScript.ObjectStorage.GetObject<AudioClip>(songName);
+        PlaySong(song);
+    }
+
+    /// <summary>
+    /// Overload for PlaySong which allows songs to be played using a direct reference.
+    /// </summary>
+    /// <param name="song">The song to play.</param>
+    public void PlaySong(AudioClip song)
```
ViMaSter: Adding this here but it applies to a few methods on various controllers (`PlaySong`, `PlaySfx`, `AddEvidence`, etc. ):

Would it be a good idea to actually remove any methods on controllers that don't use the new `ObjectStorage` and expect `string`s (soon to be `AssetName`s)?

The interfaces already only expose overloads using usages `string`s (soon to be `AssetName`s) and not using them feels like circumventing systems, that we're explicitly setting in place with this PR to keep things more stable.

Even in tests it makes sense to me, to either mock things or test the entire integration, over just testing the unit.  
Having said that, we're already kind of planning on moving away from controllers anyway, so do we even have to bother? : D

Any opinions from you on that matter?

Assets/Scripts/TextDecoder/Preloading/ObjectStorage.cs
```diff
@@ -0,0 +1,44 @@
+using System;
+using System.Collections.Generic;
+using UnityEngine;
+using Object = UnityEngine.Object;
+
+public class ObjectStorage : IObjectStorage
```
ViMaSter: Similarly to `ScriptReader` and `NarrativeScript`, is there a need to split `ObjectStorage` and `ResourceLoader`?
Right now, `ObjectStorage`s are only ever filled using `ResourceLoader`s, so I think we could safely move the functionality there. Opinions?

Assets/Scripts/TextDecoder/Preloading/ResourceLoader.cs
```diff
@@ -0,0 +1,18 @@
+using UnityEngine;
+
+/// <summary>
+/// Wraps Resources.Load in a class accessible via interface
+/// in order to avoid dependencies on UnityEngine
+/// </summary>
+public class ResourceLoader : IObjectLoader
```
ViMaSter: Nit: The line between `Resource` and `Object` is a bit blurry to me (i.e. `class ResourceLoader : IObjectLoader`). I think sticking to either `Resource` OR `Object` keeps things a bit simpler.

Assets/Scripts/Actor/ActorController.cs
```diff
@@ -1,17 +1,15 @@
 using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.Events;
-using UnityEngine.Serialization;
 
 public class ActorController : MonoBehaviour, IActorController
 {
+    [Tooltip("Drag the DialogueController here")]
+    [SerializeField] private DialogueController _dialogueController;
```
ViMaSter: (This applies to all controllers where a `DialogueController`-reference was added.)

The only thing we're effectively referencing to, is the `NarrativeScript.ObjectStorage`. Given that we're rethinking the controllers and the architecture associated with it, I'd prefer it we move the `NarrativeScript` to a new `class GameState`, which all controllers refer to instead.
This...
 - reduces coupling to the other controllers (whose purpose is a bit nebulous at the moment and ideally replaced with more specific feature-classes like the `ObjectStorage`, `ResourceLoader` or `ScreenShaker`; like in this PR)
 - makes what data the controllers are actually after more obvious

ViMaSter (changes requested):
...how is this less code **_and_** that much more readable?! Amazing work!  
As usual, anything that adds tests has me on board from the get-go, but you also made the existing ones that much more readable.

Got a few open questions and a proposal for an alternative on how the scripts get parsed during resource loading, but this is a fantastic way to get these discussions going and talk about **actual implementations**. 

IMO this is already heading in a good direction, I'm just gonna hold off on approval because of the `ObjectPreloader` implementation, but more on that in the thread about it.  
Added a few other threads as well, ideally we can come up with something for those too, but I think those are nothing really worth blocking over.

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuMouseTests.cs
PhoebeMitchell: "Select" is the definition Unity uses for highlighting buttons without clicking on them. Agree its can be a bit misleading though, and I think I've used select in other place to refer to clicking on them.

And I forgot this isn't a monobeheviour with an already existing transform property 😅

Assets/Scripts/AudioController/AudioController.cs
PhoebeMitchell: This method was added because it was the easiest way to test this class, or rather, get the existing test to work. Might be possible to create a DialogueController too and pass a narrative script to it before doing the test.

> The interfaces already only expose overloads using usages strings (soon to be AssetNames) and not using them feels like circumventing systems, that we're explicitly setting in place with this PR to keep things more stable.

Honestly I was under the impression that these classes should be concerned with just doing the things they're told to do, as opposed to also getting the object with a string reference. So they'd just be passed an AudioClip. This would follow the humble object idea of Monobehaviours handling Unity stuff only. But then again, I guess _something_ has to get the object using the string reference.

Assets/Scripts/TextDecoder/Preloading/ObjectStorage.cs
PhoebeMitchell: I did this so ObjectPreloader wasn't dependent on Resources.Load. Supposedly there are "better" asset loading options, so this would allow the ObjectLoader to be easily swapped out. This is just rhetoric though, not exactly sure of the supposed advantages. Personally I think it'd be cool if new Ink scripts could be added to the final build and loaded, without needing to use Unity (though this is an undiscussed feature). Unsure if Resources.Load would be able to accommodate that.

Assets/Scripts/TextDecoder/Preloading/ResourceLoader.cs
PhoebeMitchell: The idea was this is an object loader that specifically uses Resources.Load.
This whole class can be removed though if its decided this separation of Unity stuff isn't needed.

Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuMouseTests.cs
PhoebeMitchell: Changed to HoverOverButton and transform

Assets/Scripts/Actor/ActorController.cs
PhoebeMitchell: Instead of a new class NarrativeScripts are now entirely handled by the StoryController which has been refactored slightly. It now initialises all the scripts in the array, and stores the current narrative script which the other "controller" classes access via a property.
I felt this made more sense than having a separate class store just the current narrative script while they are managed by the story controller, at least until a bigger refactor of the controllers occurs. Thoughts?

Also penalties being enabled is currently broken but planning to fix that once #142 is merged, and should be able to address #108, #109, and #110 in the process.

Assets/Scripts/SceneLoading/SceneLoader.cs
```diff
@@ -29,7 +30,12 @@ private void Awake()
     /// </summary>
     public void LoadScene(string sceneName)
     {
-        if (!Busy)
+        if (Busy)
+        {
+            return;
+        }
+        
+        if (SceneManager.GetSceneByName(sceneName).IsValid())
```
ViMaSter: Nit: This could be flattened out even further: 
```cs
public void LoadScene(string sceneName)
{
    if (Busy)
    {
        return;
    }

    if (SceneManager.GetSceneByName(sceneName).IsValid())
    {
        Debug.LogWarning($"Could not load scene. Scene name {sceneName} is invalid.");
        return;
    }

    _sceneLoadOperation = SceneManager.LoadSceneAsync(sceneName, LoadSceneMode.Single);
    if (_sceneLoadOperation == null)
    {
        return;
    }

    Busy = true;
    Transition();
}

File: Assets/Scripts/TextDecoder/NarrativeScript.cs [hunk 12]
```diff
@@ -0,0 +1,107 @@
...
+        ReadContent(story.mainContentContainer.namedOnlyContent?.Values.ToList(), lines);
+
+        var actions = new HashSet<string>(lines.Where(line => line[0] == '&'));
+        foreach (var action in actions)
+        {
+            try
+            {
+                actionDecoder.OnNewActionLine(action);
+            }
+            catch (ScriptParsingException exception)
```
ViMaSter: Any reason exceptions are silently discarded here? Usually they should only occur, when the script has invalid syntax, at which point I would except all parsers to halt.

Assets/Scripts/TextDecoder/Preloading/ObjectStorage.cs
ViMaSter: As far as I know, `Resources.Load()` is the best way to load individual Assets in Unity, but, as you said, requires them to be part of the project inside a `Resources`-folder and present at build-time.  
The only alternatives I see, would be
 - [`AssetBundle.LoadFromFile()`](https://docs.unity3d.com/ScriptReference/AssetBundle.LoadFromFile.html) which is mostly DLC-oriented
   or
 - [`File.ReadAllBytes()`](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.readallbytes?view=netframework-4.7.2), which would allow loading files at runtime, but isn't available on all platforms  
   (like Web, iOS or Android)

For now I think merging both classes is fine, as we'd have to built a system which switches between loaders during build-time, which we probably should only consider, if we ever decide to tackle runtime narrative script loading.

Assets/Scripts/TextDecoder/Preloading/ResourceLoader.cs
ViMaSter: Gonna resolve this, [as I basically opened the same conversation twice](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/146/files#r780630624). : D

ViMaSter (dismissed):
Less code still! \o/ Thanks for addressing the `ActionDecoder` split-off! Looks super clean to me now! :)

I think this is one of the few cases where I really miss C++ references. Every script that has a `StoryController` property, is using it only as a reference to its underlying `NarrativeScript`s or `ObjectStorage`, which it can't have directly, as an update on the `StoryController` wouldn't update any of the references; just having the AudioController have a `private ObjectStorage*& activeStorage` would be so nice. 😞 

Anyway! Given that, I think I'm okay with the way we're using the StoryController now, but I would strongly prefer a more logic-specific name like `InkScriptPlaylist` however, just so we can slowly get rid of the "Controller"-suffix. I would still reserve that keyword for the [MVC](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller)-context or a `Pawn`/`Controller`-relationship; neither of which is present in our codebase. : D

Assets/Scripts/TextDecoder/NarrativeScript.cs
PhoebeMitchell: Having an abstract method for every action means we would need to update three files for every new action added, ActionDecoderBase, ActionDecoder, and ObjectPreloader as abstract methods must be implemented. This kind of goes against the idea of making actions easier to add. Furthermore having all the methods be abstract would mean the ObjectPreloader would still be full of empty methods which is what I thought we wanted to avoid.

Therefore I only included methods that required objects to be loaded, however this leads to ScriptParsingExceptions when loading the objects as the methods are not present on ObjectPreloader, so I put in the try/catch to prevent that, and as its intended functionality I didn't put any warning or error.

Could do some reflection and linq stuff to get available method names, but I feel like this is way more readable.

edit: hmm but yeah it would make sense to not catch exceptions if scripts actually did contain invalid syntax

ViMaSter: Totally valid point from you with the reusability though. 

Once this is merged, I could add a sub-type, `MethodNotFoundScriptParsingException` which we can handle (and ignore) here, but still have it throw on regular `ScriptParsingException`s. Would that work for you?

PhoebeMitchell: Yep, that sounds like it could work.

ViMaSter: Implemented the exception in 22f3be6 and updated this try-catch-block in 86523b0.

ViMaSter (dismissed):
Added 22f3be62b61e568351525fbd89395bc58b01acc4 and 86523b015903ade2c47e36935f45bc4cd7ea2277 to address the Exception handling and the rest looks good to me now. 👍

ViMaSter (dismissed):

.idea/.idea.GG-JointJustice-Unity/.idea/workspace.xml
```diff
@@ -0,0 +1,170 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ViMaSter: At the moment we're not commiting any editor-specific settings files to the repo, so I'm not sure if we want to keep this file here.

That said gonna use this change to ask if this is something we might actually want to allow:  
On paper, yes, this file is irrelevant for me (VS Community + Code user) and will pop up in any PR moving forward; the same could also be said for any VS config files for non VS-users here.
Then again, I work on this project on two machines and also have to constantly re-teach VS certain abbreviations, ignored warnings or other settings, because settings I do on machine 1 never reach machine 2. As they are project-specific, there's no way to sync them to my Microsoft or JetBrains account globally, since (at least my) tools rely on settings files inside the repo.*

Just committing them feels weird, as we'd end up with a lot of tool-specific files, then again having these things sync would be nice.

Open for opinions or maybe suggestions for tools that might fix this issue somehow. (I'd be surprised if there's no tools for this exact problem; just never made the effort to look for them since this is the first project where this has become a noticeable issue. 👀)

---

*as I'm writing this, I'm wondering why MS or JB doesn't just store it inside their accounts (if logged in) based on the git remotes path. "Hey, you're checking out `git@github.com:Studio-Lovelies/GG-JointJustice-Unity.git`? Here's the settings from your other machine!"

Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -0,0 +1,291 @@
+using System;
```
ViMaSter: These newly added files at
- `Assets/Resources/InkDialogueScripts/EvidenceAddingTestScript.json`
- `Assets/Resources/InkDialogueScripts/Ross_Cool_X.json`
- `Assets/Scripts/TextDecoder/ActionDecoder.cs`
should probably be instead be modifications to these respective files instead
- `unity-ggjj/Assets/Resources/InkDialogueScripts/EvidenceAddingTestScript.json`
- `unity-ggjj/Assets/Resources/InkDialogueScripts/Ross_Cool_X.json`
- `unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs`

These files exist as well, so might be that there's changes missing in those files from here, just as a heads-up!

unity-ggjj/Assets/Scripts/Actor/Actor.cs
```diff
@@ -1,3 +1,5 @@
+using UnityEngine;
+
```
ViMaSter: Nit: This namespace is not used anymore and should be safe to remove.
```suggestion
```

unity-ggjj/Assets/Scripts/Actor/ActorController.cs
```diff
@@ -1,18 +1,15 @@
-using System;
 using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.Events;
-using UnityEngine.Serialization;
 
 public class ActorController : MonoBehaviour, IActorController
 {
+    [Tooltip("Drag a NarrativeScriptPlaylist here")]
```
ViMaSter: Nit: Mismatching tooltip text and property class.
```suggestion
    [Tooltip("Drag a DialogueController here")]
```

unity-ggjj/Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -1,9 +1,11 @@
 using System.Collections;
 using UnityEngine;
-using UnityEngine.Serialization;
 
 public class AudioController : MonoBehaviour, IAudioController
 {
+    [Tooltip("Drag a NarrativeScriptPlaylist here")]
```
ViMaSter: Nit: Mismatching tooltip text and property class.
```suggestion
    [Tooltip("Drag a DialogueController here")]
```

unity-ggjj/Assets/Scripts/BGScenes/SceneController.cs
```diff
@@ -5,6 +5,9 @@
 
 public class SceneController : MonoBehaviour, ISceneController
 {
+    [Tooltip("Drag a NarrativeScriptPlaylist here")]
```
ViMaSter: Nit: Mismatching tooltip text and property class.
```suggestion
    [Tooltip("Drag a DialogueController here")]
```

unity-ggjj/Assets/Scripts/Evidence/EvidenceController.cs
```diff
@@ -1,28 +1,28 @@
+using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.Events;
 using UnityEngine.Serialization;
 
 public class EvidenceController : MonoBehaviour, IEvidenceController
 {
+    [Tooltip("Drag a NarrativeScriptPlaylist here")]
```
ViMaSter: Nit: Mismatching tooltip text and property class.
```suggestion
    [Tooltip("Drag a DialogueController here")]
```

unity-ggjj/Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -110,17 +121,16 @@ public void PlaySfx(AudioClip soundEffectClip)
...
         if (IsCurrentlyPlayingMusic())
         {
             yield return _musicFader.FadeOut(_transitionDuration / 2f);
         }
 
-        SetCurrentTrack(songName);
+        SetCurrentTrack(song);
         yield return _musicFader.FadeIn(_transitionDuration / 2f);
 
         _isTransitioningMusicTracks = false;
```
ViMaSter: Nit: Not _strictly_ part of this PR, but this variable is `private`, only ever set in this method and has no `SerializedField` underneath. I think that should make it safe to remove, unless I'm missing a use case.

unity-ggjj/Assets/Scripts/TextDecoder/NarrativeScriptPlaylist.cs
```diff
@@ -0,0 +1,78 @@
...
+    [field: Tooltip("A narrative script to be played on game over")]
+    [field: SerializeField] public NarrativeScript GameOverScript { get; private set; }
+    
+    [Tooltip("Write the name of the next scene to load here.")]
+    [SerializeField] private string _nextSceneName;
+
+    private SceneLoader _sceneLoader;
+    private int _narrativeScriptIndex = -1;
+
+    public NarrativeScript NarrativeScript => NarrativeScripts[_narrativeScriptIndex];
```
ViMaSter: Nit: This seems to be unused, as the `Get*Script()` methods below are used instead (which I think is actually a good thing!).

I'd argue, that this property should be safe to remove, until we ever actually introduce a component, that actually needs the current one on demand.
```suggestion
```

unity-ggjj/Assets/Scripts/BGScenes/BGSceneList.cs
```diff
@@ -1,21 +1,51 @@
...
+        
+        InstantiateBgScenes(_narrativeScriptPlaylist.GameOverScript);
+    }
+
+    /// <summary>
+    /// Instantiates BGScenes from a given narrative script and adds them to the dictionary
     /// </summary>
-    private void Awake()
+    /// <param name="narrativeScript">The narrative script to get BGScenes from</param>
+    private void InstantiateBgScenes(NarrativeScript narrativeScript)
```
ViMaSter: Nit: As the class is named `BGScene` I'd prefer is all usages or even indirect references respect that same capitalization (`InstantiateBGScenes` instead of `InstantiateBgScenes`).

---

Just a random note if that's Rider not cooperating with abbreviations and MakingFUNCasingFixes: ReSharper actually has an extra "abbreviation" section, so maybe that's something shared across the JetBrains apps. 
image
And if you already knew that, forgot I said anything. : D

unity-ggjj/Assets/Scripts/BGScenes/BGSceneList.cs
```diff
@@ -1,21 +1,51 @@
...
+    [SerializeField] private NarrativeScriptPlaylist _narrativeScriptPlaylist;
+    
+    private readonly Dictionary<string, BGScene> _bgScenes = new Dictionary<string, BGScene>();
     private BGScene _activeScene;
 
     /// <summary>
-    /// Initializes the background scene dictionary
+    /// Gets all available BGScenes and instantiates them.
+    /// </summary>
+    public void Start()
```
ViMaSter: Nit: I believe Unity lifecycle methods aren't required to be `public` and as re-calling them is probably not a good idea, I'd suggest making it `private`.
```suggestion
    private void Start()
```

unity-ggjj/Assets/Scripts/BGScenes/BGSceneList.cs
```diff
@@ -26,13 +56,13 @@ private void Awake()
     /// <returns>The new active scene. Can be null if an error occurred.</returns>
     public BGScene SetScene(SceneAssetName sceneName) //TODO: Change this when making file names universal
```
ViMaSter: Nit: I believe this `TODO:` is safe to remove now. : D
```suggestion
    public BGScene SetScene(SceneAssetName sceneName)
```

unity-ggjj/Assets/Scripts/BGScenes/BGSceneList.cs
```diff
@@ -1,21 +1,51 @@
...
+    /// <param name="narrativeScript">The narrative script to get BGScenes from</param>
+    private void InstantiateBgScenes(NarrativeScript narrativeScript)
     {
-        _scenesInChildren = new Dictionary<string, BGScene>();
-        foreach (BGScene scene in GetComponentsInChildren<BGScene>(true))
+        var backgroundScenes = narrativeScript.ObjectStorage.GetObjectsOfType<BGScene>();
+        foreach (var bgScene in backgroundScenes)
         {
-            _scenesInChildren.Add(new SceneAssetName(scene.gameObject.name), scene);
+            if (_bgScenes.Keys.Contains(bgScene.name))
```
ViMaSter: Nit: `Dictionary<K,V>`s contain a `.ContainsKey(K key)` method, which uses whatever underlying implementation is provided and should be quicker than iterating over the `Keys` and filtering them using LINQ.
This also should make `using System.Linq;` redundant.

```suggestion
if (_bgScenes.Keys.Contains(bgScene.name))
```

unity-ggjj/Assets/Scripts/UI/PenaltyManager.cs
```diff
@@ -1,27 +1,68 @@
...
-    /// Creates a specified number of penalties.
+    /// Creates penalty UI objects on examination start
     /// </summary>
-    /// <param name="penaltyCount">The number of penalties to create.</param>
-    public void Initialize(int penaltyCount)
+    public void OnCrossExaminationStart()
     {
-        for (int i = 0; i < penaltyCount; i++)
+        gameObject.SetActive(true);
+        if (_penaltyObjects.Count == 0)
```
ViMaSter: Nit: Invert this conditional to reduce nesting.

unity-ggjj/Assets/Scripts/TextDecoder/Preloading/ObjectStorage.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        return _objects.ContainsValue(obj);
+    }
+}
+
+public interface IObjectStorage
+{
+    public T GetObject<T>(string objectName) where T : Object;
+    public IEnumerable<T> GetObjectsOfType<T>() where T : Object;
+}
```
ViMaSter: Nit: I personally prefer each class having it's own file, but also believe we never decided on an official rule for it. (And I'm pretty sure I've also violated it myself once or twice. : D)

Any opinions from you on it?

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -627,13 +627,34 @@ private void SET_ACTOR_POSITION(int oneBasedSlotIndex, ActorAssetName actorName)
...
-            StoryController.OnCrossExaminationStart();
+            case GameMode.Dialogue:
+                PenaltyManager.OnCrossExaminationEnd();
+                break;
+            case GameMode.CrossExamination:
+                PenaltyManager.OnCrossExaminationStart();
+                break;
+            default:
+                Debug.LogError($"Invalid GameMode: {mode}");
+                break;
```
ViMaSter: Nit: Having this error is actually a good idea! I think this line could actually even be changed to `throw` an exception. I'd With every current game mode now having specific logic executed here, if we're ever in a scenario, where `mode` is set to a value that we don't expect, something has gone wrong bad enough to stopping execution over. 
```suggestion
                throw new NotSupportedException($"Switching to game mode '{mode}' is not supported");
```

(This would also force us to revisit this `switch`, every time we add a new game mode. That way, rather than this method just silently succeeding, we always have to make a conscious decision on whether we want something to happen here for that new game mode. Having something like `case GameMode.NewGameMode: break;` is totally fine, **if** it has been added deliberately.)

unity-ggjj/Assets/Scripts/TextDecoder/DialogueController.cs
```diff
@@ -60,18 +62,27 @@ public GameMode GameMode
 
     [Tooltip("This event is called when the _isBusy field is set.")]
     [SerializeField] private UnityEvent<bool> _onBusySet;
-
-    private Story _inkStory;
+    
+    private bool _isBusy;
```
ViMaSter: Nit: This field appears to be redundant due to `public bool IsBusy { get; private set; }` on line 44. It seems to be unused too, so it should be safe to remove.
```suggestion
```

unity-ggjj/Assets/Scripts/TextDecoder/NarrativeScript.cs
```diff
@@ -0,0 +1,103 @@
...
+    /// <param name="script">An Ink narrative script</param>
+    /// <param name="actionDecoder">An optional action decoder, used for testing</param>
+    public NarrativeScript(TextAsset script, IActionDecoder actionDecoder = null)
+    {
+        Script = script;
+        Initialize(actionDecoder);
+    }
+
+    /// <summary>
+    /// Initialises script values that cannot be set in the Unity inspector
```
ViMaSter: Nit: Inconsistency: `Initialises` -> `Initializes`
```suggestion
    /// Initializes script values that cannot be set in the Unity inspector
```

unity-ggjj/Assets/Scripts/TextDecoder/NarrativeScriptPlaylist.cs
```diff
@@ -0,0 +1,78 @@
...
+    private void Awake()
+    {
+        _sceneLoader = GetComponent<SceneLoader>();
+        InitializeNarrativeScripts(NarrativeScripts);
+        InitializeNarrativeScripts(FailureScripts);
+        GameOverScript.Initialize();
+    }
+
+    /// <summary>
+    /// Loops through a collection of narrative scripts and initialises each one
```
ViMaSter: Nit: Inconsistency: `Initialises` -> `Initializes`
```suggestion
    /// Loops through a collection of narrative scripts and initializes each one
```

unity-ggjj/Assets/Scripts/TextDecoder/NarrativeScriptPlaylist.cs
```diff
@@ -0,0 +1,78 @@
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Serialization;
```
ViMaSter: Nit: I think no symbols from this namespace are used inside this file so it should be safe to remove.
```suggestion

```

unity-ggjj/Assets/Scripts/TextDecoder/NarrativeScriptPlaylist.cs
```diff
@@ -0,0 +1,78 @@
...
+    [field: SerializeField] public List<NarrativeScript> NarrativeScripts { get; private set; }
+    
+    [field: Tooltip("List of narrative scripts to be used when taking penalties")]
+    [field: SerializeField] public List<NarrativeScript> FailureScripts { get; private set; }
+    
+    [field: Tooltip("A narrative script to be played on game over")]
+    [field: SerializeField] public NarrativeScript GameOverScript { get; private set; }
+    
+    [Tooltip("Write the name of the next scene to load here.")]
+    [SerializeField] private string _nextSceneName;
```
ViMaSter: Nit: I'm unsure if this field and it's associated logic are still relevant after this PR. This field's value is used if we "run out of lines" to execute and I would make the argument, this is something we should throw and abort execution for instead; it's actually an unexpected state:

Generally, the game's flow is planned to be completely defined via the `.ink`-scripts; thereby running out of lines _shouldn't_ really happen. If we encounter the final dialogue for the current state of the game, we should instead transition to credits (or the main menu for the time being) and do so via the `&SCENE:`-action, to give that power to writers instead of limiting it to people working in the editor only. Doing so, the Unity scene is left before "no lines" are left inside the narrative script.

Does that approach make sense to you too? Maybe there's also another case where this could prove useful I've missed, so if I did, do let me know!

unity-ggjj/Assets/Scripts/TextDecoder/NarrativeScriptPlaylist.cs
```diff
@@ -0,0 +1,78 @@
...
+    /// Loops through a collection of narrative scripts and initialises each one
+    /// </summary>
+    /// <param name="narrativeScripts">The collection of narrative scripts to initialise</param>
+    private static void InitializeNarrativeScripts(IEnumerable<NarrativeScript> narrativeScripts)
+    {
+        foreach (var narrativeScript in narrativeScripts)
+        {
+            narrativeScript.Initialize();
+        }
+    }
```
ViMaSter: Nit: This could be achieved without the need for a separate method while still being what I would consider "readable":
```suggestion
        NarrativeScripts.ForEach(script => script.Initialize());
        FailureScripts.ForEach(script => script.Initialize());
        GameOverScript.Initialize();
    }
```

That said, if you feel the extra method makes things more legible, feel free to reject this!

unity-ggjj/Assets/Scripts/UI/PenaltyManager.cs
```diff
@@ -1,27 +1,68 @@
...
+    [Tooltip("Drag a NarrativeScriptPlaylist here")]
+    [SerializeField] private NarrativeScriptPlaylist _narrativeScriptPlaylist;
+    
     [Tooltip("Drag the prefab for the penalty UI object here.")]
     [SerializeField]private Animator _penaltyObject;
 
+    [SerializeField] private int _penaltyCount = 5;
+
+    [Header("Events")]
+    [SerializeField] private UnityEvent<NarrativeScript> _onPenaltiesDepleted;
```
ViMaSter: Nit: Being the "interface over events"-guy I am, I think I personally would actually prefer adding `StartSubStory` to `IDialogueController` and using `_directorActionDecoder.DialogueController.StartSubStory()` instead of calling this event.

That way, we have less code, achieve the same goal and also the intent (queueing a game over) is clear, without having to look at the event subscribers in the editor.

Any opinions from you?

unity-ggjj/Assets/Scripts/UI/PenaltyManager.cs
```diff
@@ -1,27 +1,68 @@
...
+    /// <summary>
+    /// Destroy all penalty objects
+    /// </summary>
+    public void ResetPenalties()
+    {
+        while (_penaltyObjects.Count > 0)
+        {
+            var penalty = _penaltyObjects.Dequeue();
+            Destroy(penalty.gameObject);
+        }
```
ViMaSter: Nit: I'm totally fine if this gets dismissed as personal preference, but `while`-loops with implied exit conditions and no timeout/fallback exit condition scare me. 👀  
Implied in the sense, that `.Count` is never **directly** changed in the body (i.e. `_penaltyObjects.Count--`, which is impossible given the classes public interface) but only as a side-effect of `.Dequeue()`.  
And "no fallback" along the lines of "ran more than 10000 times for whatever reason? break! to guarantee the process never freezes".

I personally would probably prefer safely looping over the enumerable first and then clearing it afterwards instead:
```suggestion
        foreach (var penaltyAnimator in _penaltyObjects)
        {
            Destroy(penaltyAnimator.gameObject);
        }
        _penaltyObjects.Clear();
```

unity-ggjj/Assets/Tests/EditModeTests/NarrativeScriptTests.cs
```diff
@@ -0,0 +1,95 @@
+using System.Collections.Generic;
+using Ink;
+using Moq;
+using NUnit.Framework;
+using UnityEngine;
+
+public class NarrativeScriptTests
+{
+    public readonly string TestScript = "This is a test script" +
```
ViMaSter: Nit: As it's not an interpolated string, this field can actually be marked as a constant:
```suggestion
    public const string TestScript =    "This is a test script" +
```

unity-ggjj/Assets/Tests/EditModeTests/NarrativeScriptTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        objectPreloaderMock.Setup(mock => mock
+                .OnNewActionLine("&PLAY_SONG:ABoyAndHisTrial"))
+            .Callback(() => methodCalls[9]++);
+        
+        new NarrativeScript(new TextAsset(story.ToJson()), objectPreloaderMock.Object);
+
+        foreach (var pair in methodCalls)
+        {
+            Assert.AreEqual(1, pair.Value);
+        }
```
ViMaSter: We should assign Good Boy coins per case added to test assemblies. Testing and asserting just the right things. Real nice. 👀👌 

Buncha (purely syntactical) nits:
- I personally would prefer to use the action lines itself as index for `methodCalls` over hardcoding `int`s; magic numbers like `[7]` mapping to `&ADD_RECORD:TutorialBoy` are a bit difficult to read and maintain, if we ever decide to extend the test in between values
- Instead of having to set up a dictionary with each line, we could also just use one list, have it store all method calls that occurred and later use `.Count(line => line == "&ACTION:WeCareAbout")` on it, to see how often any given line was processed
- Extracting all the action lines we test for from the `TestScript` could keep this test dynamic and (at least to me) seems fine to me from a legibility perspective
- `Moq` actually allows you to `Setup` a mock once with "any value" using `It.IsAny<string>()`  and supplied values can then be received as a parameter inside `.Callback`; this removes the need to set up each value individually
- [C# 7 introduced discards (`_ = ...`)](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/discards); anything assigned to `_` is marked as "deliberately discarded". It's a way of explicitly throwing away a return value. Using it has no _technical_ benefit, but keeps most linters quiet and informs coders that a return value wasn't forgotten accidentally, but rather deliberately ignored

All of it in action:
```suggestion
        var parser = new InkParser(TestScript);
        var story = parser.Parse().ExportRuntime();

        var uniqueActionLines = TestScript.Split('\n').Distinct().Where(line => line.StartsWith("&")).ToList();
        var methodCalls = new List<string>();

        var objectPreloaderMock = new Mock<IActionDecoder>();
            objectPreloaderMock.Setup(mock => mock.OnNewActionLine(It.IsAny<string>())).Callback<string>((line) => methodCalls.Add(line));
        
        _ = new NarrativeScript(new TextAsset(story.ToJson()), objectPreloaderMock.Object);

        foreach (var uniqueActionLine in uniqueActionLines)
        {
            Assert.AreEqual(1, methodCalls.Count(line => line == uniqueActionLine));
        }
```

As usual, that's just a **suggestion**. : D If there's any specific points you don't agree with (I know generating test data is [rightfully!] a controversial subject), I'm always open for discussions on what parts we might want to skip.

unity-ggjj/Assets/Tests/EditModeTests/ObjectPreloaderTests.cs
```diff
@@ -0,0 +1,77 @@
...
+        var objects = Resources.LoadAll(path);
+
+        foreach (var o in objects)
+        {
+            var actorPositionParameter = action == "SET_ACTOR_POSITION" ? "1," : "";
+            
+            var obj = (T)o;
+            _objectPreloader.OnNewActionLine($"&{action}:{actorPositionParameter}{obj.name}");
+            var storedActor = _objectStorage.GetObject<T>(obj.name);
+            Assert.AreEqual(storedActor, obj);
```
ViMaSter: Nit: I personally prefer keeping variable names human-readable and avoid abbreviations, even in a seemingly obvious context. (`i` in `for`-loops aside. 👀)
```suggestion
        var resourcesInsidePath = Resources.LoadAll(path);

        foreach (var genericResource in resourcesInsidePath)
        {
            var actorPositionParameter = action == "SET_ACTOR_POSITION" ? "1," : "";
            
            var typeSpecificResource = (T)genericResource;
            _objectPreloader.OnNewActionLine($"&{action}:{actorPositionParameter}{typeSpecificResource.name}");
            var storedActor = _objectStorage.GetObject<T>(typeSpecificResource.name);
            Assert.AreEqual(storedActor, typeSpecificResource);
        }
```

unity-ggjj/Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuKeyboardTests.cs
```diff
@@ -1,195 +1,183 @@
...
-using Tests.PlayModeTests.Tools;
 using TMPro;
 using UnityEngine;
 using UnityEngine.TestTools;
 using UnityEngine.UI;
 
 namespace Tests.PlayModeTests.Scripts.EvidenceMenu
 {
-    public class EvidenceMenuKeyboardTests
+    public class EvidenceMenuKeyboardTests : EvidenceMenuTest
```
ViMaSter, line 11: Seeing these tests constantly getting refactored and more and more readable **and** stable each time, is a delight.

`EvidenceMenuKeyboardTests` in this PR got this perfect balance between comments for "what is happening", descriptive method names for "how things happen" and shares just the right things with it's mouse counterpart. Incredible work. 👍

ViMaSter (changes requested):
[Ooooh... Sooooo niiiice!](https://youtu.be/9g6B_7NFyWA?t=298)

Thanks to _someone_ <sup>Oh wait, that was me... 🤔</sup> moving the project to a subfolder, there's been a few files which are back from the dead. 
Once those are gone, everything else is just nice-to-have and could totally be resolved in another PR, to get this huge change list in first.

.idea/.idea.GG-JointJustice-Unity/.idea/workspace.xml
PhoebeMitchell: The .idea folder is supposed to be included in the .gitignore file, which is currently in ggjj-unity. It should be in the root directory right? I can easily delete the .idea folder, I don't have any important settings saved.

Assets/Scripts/TextDecoder/ActionDecoder.cs
PhoebeMitchell: Deleted the rogue files

unity-ggjj/Assets/Scripts/BGScenes/BGSceneList.cs
PhoebeMitchell: Yeah I was following convention, but your reasoning makes sense. Updated, and thanks for the tip.

unity-ggjj/Assets/Scripts/BGScenes/BGSceneList.cs
PhoebeMitchell: Whoops haha.

unity-ggjj/Assets/Scripts/BGScenes/BGSceneList.cs
PhoebeMitchell: Changed Keys.Contains to ContainsKey. I think thats what you meant? Your suggested change seems to just be removing white space.

unity-ggjj/Assets/Scripts/TextDecoder/Preloading/ObjectStorage.cs
PhoebeMitchell: Moved to its own file. But where should the file be? I like having interfaces with the scripts that use them, but we also have a dedicated interface folder. Thoughts?

unity-ggjj/Assets/Scripts/TextDecoder/NarrativeScript.cs
PhoebeMitchell: tfw british

unity-ggjj/Assets/Scripts/TextDecoder/NarrativeScriptPlaylist.cs
PhoebeMitchell: I agree, and it also doesn't make sense for the one scene for all scripts approach we're going for in this PR. Are you suggesting having, for example, a &LOAD_SCRIPT action at the end of every narrative script? This would require a way of either reloading a scene and injecting the script to play (which I think needs to be done at some point anyway), or destroying the assets for the current scene and loading new ones (or not destroying assets that need to be used in the new script if we want to be SUPER efficient). Maybe this should be its own issue and PR as I'm not sure how much time I will be able to dedicate at the moment with university.

.idea/.idea.GG-JointJustice-Unity/.idea/workspace.xml
PhoebeMitchell: Moved .gitignore to the root directory,

ViMaSter: `.gitignore`s can actually be put inside subdirectories; that just means, that these rules only apply inside that folder and any subdirectories (if the qualifiers match).  
For example, [the docs generator currently has its own `.gitignore`-file](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/028be977d9ceecd7d783f2021b9d6db5d4f360aa/docs_generator/.gitignore), which only contains .NET-specific settings.  
When I added the generator, I deliberately moved the `.gitignore` we had in our root to the subfolder of the Unity project.  
To me, these rules were specific to our Unity-project, so I didn't want to enforce them to (let's say) the generated docs.  
Things like `[Ll]ibrary/` being ignored is Unity specific and if - for whatever reason - we want to use that folder name outside it, we can't, if the `.gitignore` on the root-level contains that rule.

A quick note: With the `.gitignore` being in the root, [this line](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/feature/script-pre-loading/.gitignore#L7) actually no longer prevents the `UserSettings` folder inside the Unity subfolder from being added. The leading slash (`/[Uu]ser...`) means the rule is limited to only folders **next to** that `.gitignore`, which means it's not observed inside subdirectories. Thereby `/unity-ggjj/UserSettings/` would be exempt (compared to `[Bb]uild/` which limits **every** folder or subfolder named `build`). A way to fix this would be to change every rule to `/unity-ggjj/[Uu]ser...`, etc., but then this file can quickly get unreadable if it contains every rule for the entire repository.

How about this for a potential solution: I think we can keep the `.gitignores` where they were originally, but also add a global `.gitignore` with [these settings suggested from JetBrains](https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore).

---

A completely random guess, since I haven't used any of their tools directly, but maybe this is actually the underlying cause for the issues and could completely eliminate the need to update the `.gitignore` files:

In Visual Studio Code, settings are stored in the **highest folder** currently open; for example:
- If I open `/` and save changes, the VS Code settings reside inside `/.vscode/`
- If I open `/unity-ggjj` however, they reside inside `/unity-ggjj/.vscode/`

Could it be that your editor is pointing to the repository root `/`, instead of to the Unity subfolder `/unity-ggjj` and that's the reason the `.idea`-file gets created in the root?  
You could try that, by undoing the `.gitignore` changes, moving the `/.idea`-folder you already have to `/unity-ggjj/.idea` and then open `/unity-ggjj` directly, instead of `/` and seeing if
- all your settings remain
- you can still edit all files you need
- whatever Unity-specific tools inside your Editor still work

(If this works, it would also prevent any files from the `docs_generator` or `docs` polluting your workspace; no need to keep track of those files, if only the Unity project is being modified.)

unity-ggjj/Assets/Scripts/BGScenes/BGSceneList.cs
ViMaSter: Oh whoops, yeah, sorry. 🙃 You did the right thing haha

unity-ggjj/Assets/Scripts/TextDecoder/Preloading/ObjectStorage.cs
ViMaSter: Personally not a big fan of splitting things by "what they are", rather than "what they do", so I think having a `/Preloading/IObjectStorage.cs` is preferable over `/Interfaces/Preloading/IObjectStorage.cs`.
Did the same thing for [`/Assets/Scripts/TextDecoder/Parser`](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/tree/develop/unity-ggjj/Assets/Scripts/TextDecoder/Parser) too, where both the base class and derivatives never leave that folder.
Would that make sense to you too?

---

\* As an aside: I actually really enjoy web development with C#, but ASP.NET has this **awful** convention of having one folder for `Controller` (which route your URLs to C# methods), one for `Views` (what gets rendered) and one for `Models` (data you pass around). If you have an app that does subscription management, video playback, blog posts and authorization, instead of going
- Subscription management
  - Controllers
  - Views
  - Models
- Video playback
  - Controllers
  - Views
  - Models
- ...

you go

- Controllers
  - Subscription Management
  - Video Playback
  - ...
- Views
  - Subscription Management
  - Video Playback
  - ...

There's no good reason for it as far as I can tell it just... is. And no one dares to question it.
Not related to this project at all, just got reminded of it. : D

unity-ggjj/Assets/Scripts/TextDecoder/NarrativeScriptPlaylist.cs
ViMaSter: Just to focus on the field first: I'd say let's remove it for now as (I believe) it's completely unused, since we don't have any sort of scene progression yet, right?  
If that works for you, feel free to just make that change and mark this thread as resolved or let me know if you already have an alternative in mind you still want to make a part of this branch.

That said: I think splitting this up into a separate issue is definitely a good call and that `&LOAD_SCRIPT`-actions fit perfectly into something I was already thinking about:
[I've moved the "scene and script progression" aspect of it to a discussion](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/discussions/170), so we can focus this thread on the field specifically, but it's a super important topic, and I didn't want it to get drowned out and forgotten as "just another PR nitpick"! I hope that's okay for you too!

---

Unrelated note that I just wanted to put somewhere: Definitely feel free to hit the re-request review button once you've gotten around to addressing everything you wanted to and are waiting for feedback again; just so I don't lose track of when things are ready and don't start reviewing them while you're still working on them. 👀 
image

.idea/.idea.GG-JointJustice-Unity/.idea/workspace.xml
PhoebeMitchell: Reverted this change and deleted the offending folder

unity-ggjj/Assets/Scripts/TextDecoder/Preloading/ObjectStorage.cs
PhoebeMitchell: Kept the interface in Assets/Scripts/TextDecoder/Preloading

unity-ggjj/Assets/Scripts/TextDecoder/NarrativeScriptPlaylist.cs
PhoebeMitchell: Removed scene loading from NarrativeScriptPlaylist

unity-ggjj/Assets/Scripts/TextDecoder/NarrativeScriptPlaylist.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    /// <summary>
+    /// Increments the narrative script index and returns the
+    /// corresponding script in the narrative scripts array
+    /// </summary>
+    /// <returns>The next narrative script in the array</returns>
+    public NarrativeScript GetNextNarrativeScript()
+    {
+        _narrativeScriptIndex++;
+        return _narrativeScriptIndex < NarrativeScripts.Count ? NarrativeScripts[_narrativeScriptIndex] : null;
```
ViMaSter: Nit: Haven't seen this on my first review pass unfortunately, but given that we're aiming for a chapter progression system (at least roughly) similar to what's outlined in https://github.com/Studio-Lovelies/GG-JointJustice-Unity/discussions/170, I think ending in a situation where the index exceeds the count of available scripts should probably throw an exception. It's an unexpected case we should never enter and -  if we do - have no way of recovering from. Does that work for you?
```suggestion
        if (_narrativeScriptIndex >= NarrativeScripts.Count)
        {
            throw new NotSupportedException("This playlist is on its last script already");
        }
        return NarrativeScripts[_narrativeScriptIndex];
```

unity-ggjj/Assets/Scripts/TextDecoder/NarrativeScriptPlaylist.cs
```diff
@@ -0,0 +1,46 @@
...
+    }
+    
+    /// <summary>
+    /// Gets a random failure state from the list.
+    /// </summary>
+    /// <returns>Inky dialogue script containing a failure sub-story.</returns>
+    public NarrativeScript GetRandomFailureScript()
+    {
+        return FailureScripts.Count == 0 ? null : FailureScripts[Random.Range(0, FailureScripts.Count)];
```
ViMaSter: Nit: Similarly to the comment above, I would make the argument that, if this method is called and no failure scripts are available, we've entered an unexpected and unrecoverable state. (I was thinking about whether or not this could actually already be asserted during `Start()`, but if we're in a linear narrative, we might not need one as there's no way to fail.)

I could imagine something along those lines, if that makes sense to you too?
```suggestion
        if (FailureScripts.Count <= 0)
        {
            throw new NotSupportedException("This playlist contains no failure scripts");
        }
        return FailureScripts[Random.Range(0, FailureScripts.Count)];
```

unity-ggjj/Assets/Scripts/UI/PenaltyManager.cs
```diff
@@ -30,5 +71,28 @@ public void Decrement()
...
+    {
+        if (_penaltyObjects.Count == 0)
+        {
+            _directorActionDecoder.Decoder.DialogueController.StartSubStory(_narrativeScriptPlaylist.GameOverScript);
+        }
+    }
+
+    /// <summary>
+    /// Sets the number of penalties the player has left
+    /// Used for testing
```
ViMaSter: More of a comment than an actionable: I personally try to avoid code inside systems under test that are only needed and ran during testing; it's modifying the system in a way that's impossible during regular operation and since we want to ideally be testing these regular operations, we'd deviate by using it.

I think there's actually a way to remove the need for this method unless I've missed something:  
- `PenaltyTest.NumberOfPenaltiesCanBeSet` would be gone, if the method is gone
- `Tests.PlayModeTests.Scenes.CrossExamination.ViaKeyboard.GameOverPlaysOnNoLivesLeft` loads `Assets/Scenes/TestScenes/CrossExamination - TestScene.unity` and `_penaltyCount` is actually exposed to the Unity Inspector; if we set it to `1` in that scene, we're effectively achieving the same behavior

Does that make sense?

---

But, with that said:
- The same argument that is being used for not using stubs/mocks however and I actually like using them, so feel free to discard it with me praising "do as I say, not as I do". : D
- It's covered by tests, so while it might introduce behavior not observed outside of test cases, its impact is at least covered.
- I can also see how this method could come in handy if we maybe want to add difficulty or accessibility options later on. (Though, right now we're not planning for it, so I think I personally would rather not have it in for now and just re-implement it, once we have those systems fully designed and there's a requirement for it; until such time, the required implementation might also have changed, so I wouldn't make any assumptions about what kind of implementation is needed yet.)

unity-ggjj/Assets/Tests/PlayModeTests/Tools/InputTestTools.cs
```diff
@@ -123,5 +123,16 @@ public IEnumerator ClickAtPositionWorldSpace(Vector2 position)
             yield return SetMousePositionWorldSpace(position);
             yield return PressForFrame(Mouse.leftButton);
         }
+
+        /// <summary>
+        /// Holds the X Key until a DialogueController is not busy
+        /// </summary>
+        /// <param name="dialogueController">The dialogue controller to wait for</param>
+        public IEnumerator ProgressStory(DialogueController dialogueController)
```
ViMaSter, line 131: Oooooooo~, that's sooooo nifty! You're killing it with the test abstractions. 👀👌

unity-ggjj/Assets/Tests/PlayModeTests/Tools/TestTools.cs
```diff
@@ -10,12 +10,12 @@ namespace Tests.PlayModeTests.Tools
 {
     public class TestTools
     {
-        public static IEnumerator WaitForState(Func<bool> hasReachedState, double timeoutInSeconds = 10)
+        public static IEnumerator WaitForState(Func<bool> hasReachedState, double timeoutInSeconds = 10, Func<IEnumerator> action = null)
```
ViMaSter: Nit: This behavior is actually suuuper nice. I just think, for readability, I would actually prefer to see it split off into a separate method.
- `WaitForState()` (effectively does what this method did before this change) which only takes the `hasReachedState`-check function and timeout
- `DoUntilStateIsReached()` (effective does what this method did after this change) which takes all three parameters (and maybe reorder them to `action`, `hasReachedState`, `timeoutInSeconds`

ViMaSter (dismissed):
Few nits (including an initial oversight from me /o\\) but absolutely nothing that couldn't be done in another PR if there's a second approval before you get the chance to address them. (And if you are okay with it of course; your PR, your call!)

This looks like a super solid implementation now; thanks for going over all that feedback so quickly! 👍

unity-ggjj/Assets/Scripts/TextDecoder/NarrativeScriptPlaylist.cs
PhoebeMitchell: Corrected it's to its and committed suggestion

unity-ggjj/Assets/Scripts/UI/PenaltyManager.cs
PhoebeMitchell: I would agree with not having test specific stuff in the game code. Reverted back to how it was, and set number of penalties in CrossExamination Test Scene to 1.

However I would like to move towards being able to run tests on the planned main Game scene, without having individual scenes for each test. Maybe this is wishful thinking though.

unity-ggjj/Assets/Tests/PlayModeTests/Tools/TestTools.cs
PhoebeMitchell: Split into WaitForState and DoUntilStateIsReached. WaitForState works by calling DoUntilStateIsReached with the correct parameters. Is that ok?

ViMaSter: Just looked at the implementation - looks perfect! 👍 (And moving the magic number to a const is a super nice touch.)

unity-ggjj/Assets/Scripts/UI/PenaltyManager.cs
PhoebeMitchell: Then we'd avoid stuff like this randomly breaking in one scene for no reason x)
Screenshot 2022-01-30 at 00 47 23

ViMaSter: > However I would like to move towards being able to run tests on the planned main Game scene, without having individual scenes for each test. Maybe this is wishful thinking though.

Oh I think that's definitely something we should have too! Just to expand on my reasoning a bit:
In my mind, every time I look at (building or reviewing) new tests in our game, I try to categorize them into one of two categories:
- **unit tests** which ensure that a specific unit (which doesn't have to consist of only only one class; it could be a web of multiple classes or even "anything that's needed to get a MonoBehaviour derivative to work") does what it's supposed to, fails safely and ideally as much functionality of that unit is covered
- what I would call **smoke tests** (effectively sending smoke through a pipe and watching for leaks; which could be something like "playing the game from main menu start to credits without any wrong choices" and "reporting if the game ever gets stuck") that mainly ensures that both our Ink scripts are "solvable" and a player is able to execute whatever actions they need to using mouse and keyboard, to complete the game

The advantage of unit tests is that they (ideally) have few dependencies. Thereby, they should break less often (only if there's a direct change to the underlying system under test) and, if they do, give a clear indication of what's wrong. But all they do is check that code cleanly executes what it's instructed to do.  
The smoke tests ensure that the "player journey" can actually be completed which, in turn, has dependencies on almost everything. But they're needed too, as clean code is nice, but if whatever that working code is instructed to do still prevents you from finishing the game, it's still a broken game.

Given that these "smoke tests" have a lot of dependencies, ideally they're few and only written if whatever they depend on is as stable as it can be, as tests that are difficult to maintain (either because they are complex to understand or break often) can be frustrating and (at least speaking for me : D) when I'm frustrated, I'm less productive, demotivated and less inclined to contribute new code. So while we're not able to complete our game loop from start to finish, I prefer limiting tests to be as independent and isolated as possible.

Once we can, we should definitely make sure we not only cover the underlying architecture, but "actually playing the game" too, since hey, we're actually still making a game. : D

I hope that makes some degree of sense? 👀 If you want to, we can also open up a discussion for it, if you disagree or maybe have a different perspective on it and think there's a better way to structure these things in the future. After all it's just "one potential structure" that I just stuck to, since so far it has proven useful to me. It's nothing we've ever formally agreed on and maybe having it written down somewhere (like the code style guide) or actually structuring the tests in the code to mirror whatever structure we agree on would be a good idea too. (When I split the PlayMode tests into `Scripts` and `Scenes`, I was actually basing it on this structure of "unit tests" and "smoke tests" in my mind respectively.)  
That said, if you see any better way forward, I definitely wanna hear it! :)

**EDIT:** To add to your last response: Great example. : D I think coming up with a list of "supported scenes" and automated tests for them at some point might be a good idea.

---

Oh and feel free to mark this as resolved if you've made it this far. ...brevity has never been my strong suit. 👀💦

ViMaSter (approved):

unity-ggjj/Assets/Scripts/TextDecoder/NarrativeScriptPlaylist.cs
ViMaSter: Its/it's. It ALWAYS trips me up. Thanks for the fix. 👍

notexplosive (approved):

PR #156, opened 2022-01-21
Added save file management (via C# and UNLOCK_CHAPTER action line)

### LINKED ISSUE #116: Save game
## Description
The game must save when the player reaches new parts of the cases, so they can resume after exiting.
It's probably best to avoid PlayerPrefs.
DO NOT use C#'s BinaryFormatter.
Maybe save data to AppData, or documents? It must work independent of platform. Discuss with programming team.

## Acceptance Criteria
This story will be finished when a system is in place that write data to a file, then load that data when the game opens.

ViMaSter:
## Summary
This PR implements a persistent storage using Unity's [`PlayerPrefs`](https://docs.unity3d.com/ScriptReference/PlayerPrefs.html) and the classes inside the `Savefiles` namespace.

---

#### [Savefiles.SaveData](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/feature/116-savefiles/unity-ggjj/Assets/Scripts/Savefiles/SaveData.cs)
A `plain-old data object`/`passive object`/`pure data structure` which data is serialized.  
Constructing an instance of it by hand, results in whatever we want the default settings to be.  
Currently only contains a version number and a [bit field](https://docs.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=netframework-4.8) of unlocked chapters for testing purposes.

#### [Savefiles.Proxy](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/feature/116-savefiles/unity-ggjj/Assets/Scripts/Savefiles/Proxy.cs)
This class is our layer of abstraction to Unity's persistent storage:
- `HasExistingSaveData()` returns a bool
- `Load()` throws if the method above returns false or returns the `Savefiles.SaveData` currently stored
- `UpdateCurrentSaveData(callback)` takes a function that can modify save data, runs it and then saves the values (a good example is the [`UNLOCK_CHAPTER`-action implementation](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/feature/116-savefiles/unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs#L619-L628))

This class also takes care of upgrading save files between game versions. `UpgradeSaveData()` is currently basically empty and only needs to change, if we ever make breaking changes between versions. (For example if we first store the screen resolution in two `float`s but later decide to move it to a `Vector2`.)

##### Before
image
##### After
image

## Testing instructions

1. Play the `Savefiles - TestScene.unity` once to unlock chapters
2. Play the scene again, notice chapters already being unlocked
3. Without running the game, inside the Unity Editor, press `Cmd+Shift+D`/`Ctrl+Shift+D` or go to `Edit/Delete Save Data`
4. Play the `Savefiles - TestScene.unity` scene again and notice all previously unlocked chapters being locked
5. Make sure existing and the [newly added tests](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/feature/116-savefiles/unity-ggjj/Assets/Tests/EditModeTests/SaveDataTests.cs) complete successfully

## Additional information

- `[ ]` Changes UI
  - Savefiles
  - Closes https://github.com/Studio-Lovelies/GG-JointJustice-Unity/issues/116


unity-ggjj/Assets/Scripts/Savefiles/ChapterExtensions.cs
```diff
@@ -0,0 +1,10 @@
+namespace Savefiles
```
PhoebeMitchell: Nit: Capital F for Files (change folder name as well)

unity-ggjj/Assets/Scripts/Savefiles/SaveData.cs
```diff
@@ -0,0 +1,50 @@
...
+                Chapter1 = 1 << 0,
+                Chapter2 = 1 << 1,
+                Chapter3 = 1 << 2,
+                BonusChapter1 = 1 << 3
+            }
+
+            public Chapters UnlockedChapters;
+        }
+
+        public Progression GameProgression;
```
PhoebeMitchell: Make this a property?

unity-ggjj/Assets/Scripts/Savefiles/Proxy.cs
```diff
@@ -0,0 +1,91 @@
+using System;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using UnityEngine;
+
+namespace Savefiles
+{
+    public class Proxy
```
PhoebeMitchell: Could this class name be more informative? e.g. SaveProxy, PlayerPrefsProxy

unity-ggjj/Assets/Scripts/Savefiles/SaveData.cs
```diff
@@ -0,0 +1,50 @@
...
+                None = 0,
+                Chapter1 = 1 << 0,
+                Chapter2 = 1 << 1,
+                Chapter3 = 1 << 2,
+                BonusChapter1 = 1 << 3
+            }
+
+            public Chapters UnlockedChapters;
+        }
+
```
PhoebeMitchell: Could this be moved outside of SaveData? Or does it need to be its own class at all? Either of these changes would avoid qualifying Progression.Chapters with SaveData in other scripts.

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -15,6 +16,7 @@ public class ActionDecoder
     public IAudioController AudioController { get; set; }
     public IEvidenceController EvidenceController { get; set; }
     public IAppearingDialogueController AppearingDialogueController { get; set; }
+    private readonly Savefiles.Proxy _savefileProxy = new Savefiles.Proxy();
```
PhoebeMitchell: Nit: Move this above the properties. It's not in our style doc but i like private fields and properties being separate.

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -613,6 +615,19 @@ private void SET_ACTOR_POSITION(int oneBasedSlotIndex, ActorAssetName actorName)
         ActorController.AssignActorToSlot(actorName, oneBasedSlotIndex);
         OnActionDone?.Invoke();
     }
+
+    /// <summary>Unlocks a new chapter inside the chapter select. **(This is persistent, even when the game is restarted!)**</summary>
+    /// <param name="chapter">Name of the chapter to unlock</param>
+    /// <example>&amp;UNLOCK_CHAPTER:CHAPTER_2</example>
+    /// <example>&amp;UNLOCK_CHAPTER:BONUS_CHAPTER_2</example>
+    /// <category>Progression</category>
+    private void UNLOCK_CHAPTER(Savefiles.SaveData.Progression.Chapters chapter)
```
PhoebeMitchell: Redundant qualifier Savefiles

unity-ggjj/Assets/Scripts/Savefiles/Proxy.cs
```diff
@@ -0,0 +1,91 @@
...
+            // otherwise simply parse the stored SaveData JSON blob
+            return JsonConvert.DeserializeObject<SaveData>(saveDataFromPlayerPrefs);
+        }
+
+        public delegate void OperationOnCurrentSaveData(ref SaveData saveData);
+
+        /// <summary>
+        /// Public interface to use when making changes to the current SaveData
+        /// </summary>
+        /// <param name="operation"></param>
```
PhoebeMitchell: Add parameter description

unity-ggjj/Assets/Scripts/UI/SaveFileWatcher.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        _text.text = "Checking for new state in 3.\n\n" + currentState;
+        yield return new WaitForSeconds(1.0f);
+        _text.text = "Checking for new state in 2..\n\n" + currentState;
+        yield return new WaitForSeconds(1.0f);
+        _text.text = "Checking for new state in 1..\n\n" + currentState;
+        yield return new WaitForSeconds(1.0f);
+        StartCoroutine(Loop());
+    }
+}
```
PhoebeMitchell, line 41: As this script is just used as a demo, maybe move it to one of the test assemblies

unity-ggjj/Assets/Tests/EditModeTests/SaveDataTests.cs
```diff
@@ -0,0 +1,109 @@
...
+{
+    private readonly Proxy _proxy = new Proxy();
+
+    [SetUp]
+    [OneTimeTearDown]
+    public void EnsureDefaultSaveFileExists()
+    {
+        // Force creation of a new default SaveData object
+        Proxy.DeleteSaveData();
+        _proxy.UpdateCurrentSaveData((ref SaveData saveData) => saveData = new SaveData(SaveData.LatestVersion));
```
PhoebeMitchell: These tests delete data saved in the demo. Not sure how big a problem this is but I could see it being a minor annoyance if I was trying to test or debug something and deleted my save data. Would it be possible to make it so it saves to a separate key for tests, and delete that data when tests are done?

PhoebeMitchell (dismissed):
yay, saving. I like

unity-ggjj/Assets/Scripts/Savefiles/SaveData.cs
ViMaSter: Just for reference: Technically they absolutely could! Serialization isn't tied to child-classes or namespaces.
This is probably a stylistic question, but I personally actually really like that, if this data is referenced, we need to be explicit about it.
`Progression` or `Chapters`  (or later on things like `MusicVolume`, `Resolution`, `Difficulty`), can mean a lot in various contexts (imagine like a script which listens to changes on a `MusicVolume`-slider in UI settings, to persist these values, once a dialog box is closed).  
Having to explicitly specify `SaveData.Progression.Chapters` or `SaveData.AudioSettings.MusicVolume` or a global `using Chapters = SaveData.Progression.Chapters` at the top of that file, makes it pretty clear, that objects are related to the values persisted inside save files, not some other script-specific context.

Could you post a draft of how you think these classes would be better laid-out? A nested list or some pseudo-code is totally fine; maybe there's a better way I'm just not seeing at the moment! : o

unity-ggjj/Assets/Scripts/Savefiles/SaveData.cs
ViMaSter: I actually deliberately kept all members inside `SaveData` a field. For me, a property indicates some underlying logic (for example special getters that do some kind of initialization logic like with, or having getter and setter have different access rights) but as the purpose of this class is purely to get filled and serialized, I wanted to avoid that.

Any specific reason that made you prefer a property over a field here?

unity-ggjj/Assets/Scripts/Savefiles/Proxy.cs
ViMaSter: I was struggling with that one actually, once I wrapped all of these classes inside a namespace (on it's own `SaveFileProxy` would've been fine). : D
`PlayerPrefsProxy` is a pretty good name though, going with that!

unity-ggjj/Assets/Tests/EditModeTests/SaveDataTests.cs
ViMaSter: Actually had that exact thought when building it, but was both not bothered enough to change it, yet annoyed enough to not be happy with it. Glad you're seeing the potential issue here too.

Having `Savefile.Proxy` accept a key as a constructor and create sub-savefiles could work, yet having multiple places to store data isn't really needed for the game. Adding that interface might invite misuse or confusion about how this class should be used.

I'm not sure if Unity **guarantees** that `TearDown`s and `SetUp`s get called if anything fails during test execution, but I could have the tests load and store the current save on `SetUp` and attempt to save it again on `TearDown`. Would that work for you?

ViMaSter (commented):
Thanks so much for the quick review again! Gonna address some of the smaller stuff and added a few responses to some of your concerns!

unity-ggjj/Assets/Scripts/UI/SaveFileWatcher.cs
ViMaSter: Hmm... This is actually an interesting one. I 100% agree, "demo" scripts like this one, should definitely not be part of the game's assembly, but in my mind we don't really have a better place:

- `EditMode`: reserved for automated tests
- `PlayMode`: 👆 same here
- `Editor`: reserved for anything extending the editor
- `GG-JointJustice`: anything the user actually needs to run the game
- `GG-JointJustice-Old`: ideally gone soon (reminded me to actually create an issue for that)

`Editor` _could_ technically work; it's code only needed by developers, but it's not really extending the editor, but I think adding a new `Demo` assembly should be the way to go. In there, we can have anything needed to show off functionality, but that has no actual use inside the game, automated tests or extends Editor functionality.

Would that make sense to you too or do you think this might be overcomplicating things?

unity-ggjj/Assets/Tests/EditModeTests/SaveDataTests.cs
ViMaSter: Running the EditMode tests now backups and restores files on `Setup` and `TearDown` respectively. I deliberately tried to break a test too and my save was still intact after running the tests.

If that's not ideal, let me know of any alternatives you think might be better to consider, definitely still open for alternatives here! :)

unity-ggjj/Assets/Scripts/Savefiles/SaveData.cs
PhoebeMitchell: Because it's been drilled into me that "public field bad" :p
Also its in the style doc

unity-ggjj/Assets/Scripts/UI/SaveFileWatcher.cs
PhoebeMitchell: I'm fine with having a Demo assembly

unity-ggjj/Assets/Scripts/Savefiles/SaveData.cs
PhoebeMitchell: Thanks, your "SaveData.Progression.Chapters or SaveData.AudioSettings.MusicVolume" example makes this make more sense to me.

unity-ggjj/Assets/Scripts/Savefiles/PlayerPrefsProxy.cs
```diff
@@ -0,0 +1,91 @@
...
+using Newtonsoft.Json.Linq;
+using UnityEngine;
+
+namespace SaveFiles
+{
+    public static class PlayerPrefsProxy
+    {
+        public static bool HasExistingSaveData()
+        {
+            return PlayerPrefs.HasKey("SaveData");
```
notexplosive: On WebGL are we sharing this dictionary with anything else? Should we make this string something like `"GGJJ_SaveData"`?

Also, extract this string into a static/constant.

unity-ggjj/Assets/Scripts/Savefiles/PlayerPrefsProxy.cs
```diff
@@ -0,0 +1,91 @@
...
+            // we need to upgrade them to the current format
+            if (currentlyStoredSaveData.Version < SaveData.LatestVersion)
+            {
+                UpgradeSaveData(currentlyStoredSaveData.Version, JObject.Parse(saveDataFromPlayerPrefs), ref currentlyStoredSaveData);
+                Save(currentlyStoredSaveData);
+                return currentlyStoredSaveData;
+            }
+
+            // otherwise simply parse the stored SaveData JSON blob
+            return JsonConvert.DeserializeObject<SaveData>(saveDataFromPlayerPrefs);
```
notexplosive: can't you just `return var currentlyStoredSaveData` here rather than deserializing again?

unity-ggjj/Assets/Scripts/Savefiles/PlayerPrefsProxy.cs
```diff
@@ -0,0 +1,91 @@
...
+        /// is lower than <see cref="SaveData.LatestVersion"/>, we need to migrate this object from the old format to the current one
+        /// </remarks>
+        /// <param name="outdatedVersion">The version the <see cref="outdatedSettingsJSON"/> is based on</param>
+        /// <param name="outdatedSettingsJSON">A JObject representing the outdated JSON-blob of the loaded settings</param>
+        /// <param name="upgradedSaveData">Reference to a <see cref="SaveData"/> instance which needs to contain the final, upgraded object</param>
+        private static void UpgradeSaveData(int outdatedVersion, JObject outdatedSettingsJSON, ref SaveData upgradedSaveData)
+        {
+            // currently void, as there is only SaveData of version 1 at this time
+            upgradedSaveData.Version = SaveData.LatestVersion;
+        }
```
notexplosive: What are your thoughts on making SaveData immutable. `UpgradeSaveData(ref data)` becomes `var newData = GetUpgradedSaveData(data)` (avoids the use of ref)

unity-ggjj/Assets/Scripts/UI/SaveFileWatcher.cs
notexplosive: Just dropping by to say I am also fine with a Demo assembly.

unity-ggjj/Assets/Tests/EditModeTests/ActionDecoderTests.cs
```diff
@@ -125,8 +146,14 @@ public void RunCommandWithTooFewArguments(string methodName)
...
         var method = decoder.GetType().GetMethod(methodName, BindingFlags.Instance | BindingFlags.NonPublic);
         Assert.NotNull(method, $"Couldn't find method with name '{methodName}' on object of type '{nameof(ActionDecoder)}'");
-        var generatedParameters = method.GetParameters().Select(parameterInfo => _validData[parameterInfo.ParameterType]).Select(validParameter => validParameter.ToString()).ToList();
-        var lineToParse = $"&{methodName}{(generatedParameters.Any() ? ":" : "")}{string.Join(",", string.Join(",", generatedParameters.Skip(1)))}";
+        var generatedParameters = method.GetParameters().Select(parameterInfo => {
+            if (!ValidData.ContainsKey(parameterInfo.ParameterType))
+            {
+                throw new NotImplementedException($"In order for this test to run, you need to specify a valid ink-script example for '{parameterInfo.ParameterType}' inside '{nameof(ActionDecoderTests)}.{nameof(ValidData)}'");
+            }
+            return ValidData[parameterInfo.ParameterType];
```
notexplosive: You've copy-pasted this block a few times. Any chance this can be extracted into a method? At the very least that string that's used over and over again.

notexplosive (dismissed):
Few notes. Looking good!

unity-ggjj/Assets/Scripts/Savefiles/PlayerPrefsProxy.cs
ViMaSter: > On WebGL are we sharing this dictionary with anything else? Should we make this string something like "GGJJ_SaveData"?

If I'm reading [the Unity docs about it](https://docs.unity3d.com/ScriptReference/PlayerPrefs.html#:~:text=On%20WebGL%2C%20Unity%20stores%20PlayerPrefs%20data%20using%20the%20browser%27s%20IndexedDB%20API.%20For%20more%20information%2C%20see%20IndexedDB.) right, I'm expecting Unity to already take care of that: Unlike Cookies or Local Storage, `IndexedDB` allows for one domain to have multiple databases each with multiple tables: 
image
(Hit F12, paste this into the `Console` tab, hit enter and then check the `Application` tab)
```js
const db = await indexedDB.open("coolDB", 1);

db.onupgradeneeded = event => { 
    db.result.createObjectStore("coolTable")
};
db.onsuccess = async event => { 
    transaction = db.result.transaction("coolTable", "readwrite");
    store = transaction.objectStore("coolTable");
    await store.add({a: 1, b: 2}, 1337)
}
```

Given that sites like itch.io or Kongregate host(ed) multiple Unity games on the same domain, I'm assuming that Unity generates a new Database based on the product name we specify, so we shouldn't expect any sort of conflicts. (That said: Once we have a WebGL build it would be good to check, I just wouldn't expect Unity to create a new IndexedDB database for every key, and rather have one per "product".)

Also that reminds me that we currently don't run WebGL builds automatically; gonna add an Issue for that. 
 (**EDIT:** https://github.com/Studio-Lovelies/GG-JointJustice-Unity/issues/172)

> Also, extract this string into a static/constant.

Good catch; gonna update that!

unity-ggjj/Assets/Scripts/UI/SaveFileWatcher.cs
ViMaSter: Awesome! Added an issue for it https://github.com/Studio-Lovelies/GG-JointJustice-Unity/issues/173 and I'm gonna mark this as resolved for this PR, as I'd rather move all demo assets in one go, rather than create the Assembly here but then only move this code (or end up blowing up this PR if I move all files now.)

(But do open this back up if that's not okay with anyone!)

unity-ggjj/Assets/Tests/EditModeTests/ActionDecoderTests.cs
ViMaSter: I'm always cautious with abstracting things in tests, but always happy to, if things become too repetitive.

Feel free to mark this as resolved, if the implementation in 18fa6d3680474409496a71ba1d4bd04bdc3689ed works for you! (`RunCommandWithIncorrectParameterTypes()` still has its own implementation, as - unlike the other tests - we make use of **invalid** example data here, if the parameter type has any.)

unity-ggjj/Assets/Scripts/SaveFiles/PlayerPrefsProxy.cs
```diff
@@ -0,0 +1,102 @@
...
+            // we need to upgrade them to the current format
+            if (currentlyStoredSaveData.Version < SaveData.LatestVersion)
+            {
+                currentlyStoredSaveData = CreateUpgradedSaveData(currentlyStoredSaveData.Version, JObject.Parse(saveDataFromPlayerPrefs));
+                Save(currentlyStoredSaveData);
+                return currentlyStoredSaveData;
+            }
+
+            // otherwise simply pass the data currently present
+            return currentlyStoredSaveData;
```
PhoebeMitchell: Is the first return redundant here?

unity-ggjj/Assets/Scripts/Savefiles/Editor/ClearSavefileButton.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        if (!EditorUtility.DisplayDialog("Delete Save Data", "Are you sure you want to delete your Save Data?\r\n\r\nYou will lose ALL progress in the game!", "Delete", "Cancel"))
+        {
+            return;
+        }
+
+        Debug.LogWarning("Save Data has been successfully deleted!");
+        Proxy.DeleteSaveData();
+    }
+}
```
PhoebeMitchell: I feel like it might make more sense to have this in Assets/Scripts/Editor instead of Assets/Editor/Scripts.
Will there be anything except for scripts that would go in an "Editor" folder?

Assuming you decide against the above, maybe make an assembly definition for this and other editor scripts.

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -618,6 +625,19 @@ private void SET_ACTOR_POSITION(int oneBasedSlotIndex, ActorAssetName actorName)
...
```
PhoebeMitchell: You dismissed my comment on the redundant Savefiles qualifier without explanation in my last review. Did you decide to keep it? "SaveData" seems explantory enough to me without the Savefiles qualifier.

PhoebeMitchell: Nit: The data in SaveData is capitalised but the the files in Savefiles is not. I think files should be capitalised because SaveFiles is two words.

unity-ggjj/Assets/Scripts/UI/SaveFileWatcher.cs
```diff
@@ -0,0 +1,38 @@
...
+public class SaveFileWatcher : MonoBehaviour
+{
+    private Text _text;
+
+    // Start is called before the first frame update
+    private void Start()
+    {
+        // make sure a saveFile exists
+        SaveFiles.PlayerPrefsProxy.UpdateCurrentSaveData((ref SaveData _) => {});
+
```
PhoebeMitchell: This is a little confusing. If I understand correctly this is to check that a save file exists, and if none exists it makes one, and this is achieved by calling UpdateCurrentSaveData and telling it to do nothing?

It might be more readable to use the HasExistingSaveData method, and a separate method for creating a new save data.

unity-ggjj/Assets/Scripts/UI/SaveFileWatcher.cs
```diff
@@ -0,0 +1,38 @@
...
+        _text = GetComponent<Text>();
+        StartCoroutine(Loop());
+    }
+
+    IEnumerator Loop()
```
PhoebeMitchell: Nit: Missing private access modifier

unity-ggjj/Assets/Scripts/SaveFiles/SaveData.cs
```diff
@@ -0,0 +1,50 @@
...
+                Chapter1 = 1 << 0,
+                Chapter2 = 1 << 1,
+                Chapter3 = 1 << 2,
+                BonusChapter1 = 1 << 3
+            }
+
+            public Chapters UnlockedChapters;
+        }
+
+        public Progression GameProgression;
```
PhoebeMitchell, line 36: Would still prefer if this was an auto-property, if only for consistency's sake.

unity-ggjj/Assets/Scripts/SaveFiles/PlayerPrefsProxy.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// </summary>
+        /// <param name="operation">Method that makes changes to the currently saved data by modifying a SaveData object</param>
+        public static void UpdateCurrentSaveData(OperationOnCurrentSaveData operation)
+        {
+            if (!HasExistingSaveData())
+            {
+                Save(new SaveData(SaveData.LatestVersion));
+            }
+            var currentData = Load();
+            operation(ref currentData);
```
PhoebeMitchell: This saves then loads immediately. Would it be possible to use the new SaveData that was just saved in the load?

unity-ggjj/Assets/Tests/EditModeTests/ActionDecoderTests.cs
```diff
@@ -21,10 +22,10 @@ public class ActionDecoderTests
         })
         .SelectMany(list => list);
     private static IEnumerable<string> AvailableActionsWithNoOptionalParameters => AvailableActionMethods.Where(method => method.GetParameters().Any() && method.GetParameters().All(parameter => !parameter.IsOptional)).Select(methodInfo => methodInfo.Name);
-    private static IEnumerable<string> AvailableActionsWithInvalidParameters => AvailableActionMethods.Where(method => method.GetParameters().Any(parameter => _invalidData.Keys.Contains(parameter.ParameterType))).Select(methodInfo => methodInfo.Name);
+    private static IEnumerable<string> AvailableActionsWithInvalidParameters => AvailableActionMethods.Where(method => method.GetParameters().Any(parameter => InvalidData.Keys.Contains(parameter.ParameterType))).Select(methodInfo => methodInfo.Name);
 
     // These are example values for parameters of "Ink"-script action lines for the ActionDecoder
-    private static readonly Dictionary<Type, object> _validData = new Dictionary<Type, object> {
+    private static readonly Dictionary<Type, object> ValidData = new Dictionary<Type, object> {
         {typeof(string), "ValidString"},
```
PhoebeMitchell, line 29: I assume this change to PascalCase is to be more in line with the rest of the file?

https://github.com/Studio-Lovelies/GG-JointJustice-Unity/discussions/163
Here you suggest private fields should use _prefixedCamelCase. Should they be PascalCase when they are static or readonly? I don't have a preference but would like it added to any updated style docs we may produce.

PhoebeMitchell (dismissed):
All looks good and functional so approving. Some questions though

unity-ggjj/Assets/Editor/Scripts/Savefiles/ClearSavefileButton.cs
```diff
@@ -0,0 +1,19 @@
+using SaveFiles;
+using UnityEngine;
+
```
PhoebeMitchell: Nit: remove blank line

unity-ggjj/Assets/Scripts/SaveFiles/PlayerPrefsProxy.cs
ViMaSter: True, good catch! Result of me [addressing Explosive's feedback here](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/156#discussion_r796269913) and not taking into consideration the full picture.

unity-ggjj/Assets/Scripts/Savefiles/Editor/ClearSavefileButton.cs
ViMaSter: > I feel like it might make more sense to have this in Assets/Scripts/Editor instead of Assets/Editor/Scripts.

Folder structure! My favorite! (/o\\)

I think my original logic goes back to the argument we had (in another PR? losing track : D) about either
-  "what is is" containing "what it is for" (i.e. `Views/AccountManagement/BillingInformation.html`) _or_
-  "what it is for" containing "what it is" (i.e. `AccountManagement/Views/BillingInformation.html`

Right now my logic was to first separate scripts by feature (`/Assets/Scripts/Savefiles`) and then editor-utility code in an `Editor`-subfolder [to have it be excluded from the game build](https://docs.unity3d.com/Manual/SpecialFolders.html#:~:text=Editor%20scripts%20add%20functionality%20to%20Unity%20during%20development%2C%20but%20aren%E2%80%99t%20available%20in%20builds%20at%20runtime.%20Scripts%20in%20a%20Editor%20folder%20run%20as%20Editor%20scripts%2C%20not%20runtime%20scripts.) (which may be what you were concerned about, regarding making an extra Assembly definition? Anything inside **any** folder named `Editor` is automatically excluded for game builds _because Unity_).

(If we hadn't already used a `Scripts` folder, I would even go as far as `/Assets/SaveFiles/Game` and `/Assets/SaveFiles/Editor` but wanted to adhere to our existing structure. 👀) 

> `Will there be anything except for scripts that would go in an "Editor" folder?`

_Technically_ we can extend the editor with Resources like sprites and sound effects just like in-game. So if we ever decide to spice things up with sprites or SFX (or even things like config files), they could live there, so that they don't get included in the game build.  
Do I realistically see us using that in the near future though? Not really. : D

I can definitely see both ways and I'm absolutely okay with going with your approach, if that structure makes more sense to everyone!  
How about we go with the current implementation for this PR and I open another discussion to see how we want to approach that moving forward (and create a separate refactoring-branch changing all affected code to whatever we agree to use then)? (If your editor assembly concern is covered by the `Editor`-directory name oddity from Unity, that is. 👀)

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
ViMaSter: Oh whoops, sorry for that! I think I accidentally closed that one! Addressing it in my next commit and thanks for bringing it up again.

ViMaSter: That was actually a side effect of me (incorrectly) using `Savefiles` instead of `SaveFiles` as folder name; already fixed in 324a38bd923e4014946aff4da4467a3ae1fbfff8!

unity-ggjj/Assets/Scripts/UI/SaveFileWatcher.cs
ViMaSter: > If I understand correctly this is to check that a save file exists, and if none exists it makes one, and this is achieved by calling UpdateCurrentSaveData and telling it to do nothing?

Exactly, yes! `UpdateCurrentSaveData` always succeeds; if there's an existing file you can modify it, if not it creates a default one first.

I actually merged these methods, to have the public interface of that class to be as small as possible, but would something like this work better for you?
```cs
class SaveFiles {
  class PlayerPrefsProxy {
    public void EnsureSaveDataExists() {
      if (!HasExistingSaveData()) {
          Save(new SaveData(SaveData.LatestVersion));
      }
    }
  }
}
```
A separate `Create()` method would feel weird to me:  
If it doesn't take any parameter, it's not obvious what it is creating.
If it does, it would mean the default value of a stored save file can be overwritten by whoever is using the Proxy, rather than be dictated by the SaveData itself.

unity-ggjj/Assets/Tests/EditModeTests/ActionDecoderTests.cs
ViMaSter: Hmm... Actually a good point to think about.  
ReSharper defaults to PascalCase for properties - no matter the qualifiers - so that's just what I went with (blindly 👀) : 
image

I don't think we have a private property anywhere else inside our current codebase?

Open for whatever makes the most sense for everyone, no real preference here.

unity-ggjj/Assets/Scripts/Savefiles/Editor/ClearSavefileButton.cs
PhoebeMitchell: I did not know Editor was excluded from the game build but good to know. My main "concern" is having all scripts in the same place so they're easy to access in my IDE. Not a big fan of '"what it is for" containing "what it is"' in the case of game components because it makes for a bit of top heavy structure, with the assets folder containing a folder for every thing in the game, with most of the things in those folders being things I don't need in my IDE.

That said I see the merits of having editor stuff separate. My suggestion for a different assembly definition was so we couldn't access editor stuff from the game assembly, but we can't access global assembly stuff anyway so that's kind of a moot point.

unity-ggjj/Assets/Scripts/UI/SaveFileWatcher.cs
PhoebeMitchell: That looks good. Mostly just want the `(ref SaveData _) => {}` hidden

unity-ggjj/Assets/Tests/EditModeTests/ActionDecoderTests.cs
PhoebeMitchell: I'm looking at line 28 which isn't a property. Rider seems to default to _prefixedCamelCase for private static and private readonly, but PascalCase for private static readonly. This seems a little odd to me. Feels like they should all be _prefixedCamelCase. I'm happy to go down the path of least resistance and just accept Rider defaults which I assume are the same as ReSharper defaults.

unity-ggjj/Assets/Scripts/Savefiles/Editor/ClearSavefileButton.cs
ViMaSter: > My main "concern" is having all scripts in the same place so they're easy to access in my IDE. [...] with the assets folder containing a folder for every thing in the game, with most of the things in those folders being things I don't need in my IDE

Makes sense, definitely... Unsure about other integrations, but when I select Visual Studio in the Preferences and then ask Unity to regenerate the project files for it and open that directly, it only contains C# scripts for me (not even .ink-files). Maybe there's integrations for Rider (or Fleet) too?

image

Then again, the whole idea behind VS Code for example, is that you don't have VS Code-specific workspaces or solutions, rather you just point it at a folder / file and - if there is a `.vscode`-folder in the opened folder - it can treat that as a workspace-esque thing. I think most JetBrains Software operates similarly, in which case this might not work and 
 definitely will fill up whatever explorer widget you have rather quickly.

Let me know if generating a project file is a thing you can do for your IDE and if not I'll move it to the root level `Editor`-folder; I'm not married to the idea and if it inconveniences you and others, I see no reason to stick to it!

unity-ggjj/Assets/Tests/EditModeTests/ActionDecoderTests.cs
ViMaSter: Whoops, for some reason ended up on the wrong line, my bad!
image
And yes, same here; it's the combination of `static` and `readonly`.

I'm also kinda torn on it; I can definitely see the argument for private -> `_prefixedCamelCase` taking precedence, but also anything "more constant" being "more capitalized" (thereby `PascalCase`).

I personally think going for the path of least resistance for now is a good idea too (since we both seem to enjoy the same default settings : D), but I'll update the discussion to contain all settings and then we can change them to whatever we feel more comfortable with as a team. But as always, feel free to re-open this, if that doesn't work for you or you spontaneously come up with a reason to already change it for this PR!

**EDIT:** [Post on the discussion](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/discussions/163#discussioncomment-2103020).

unity-ggjj/Assets/Scripts/SaveFiles/SaveData.cs
ViMaSter: I'm actually running into a problem with that: Since properties return a temporary value , it's not possible to modify child fields using an extension method (accessing a property calls the `get`ter, which returns a new copy of the value, rather than exposing the underlying field directly, which prohibits `ref this`), which I actually find to be rather useful:
image

A possible workaround would be to directly change the value:
image

If you find this to be more readable, I'm okay with keeping it, but as I mentioned, I would actually prefer having serialized data to be as plain as possible, without any language-specific implementation details (as JSON has no concept of properties and fields like C# does).

unity-ggjj/Assets/Scripts/Savefiles/Editor/ClearSavefileButton.cs
PhoebeMitchell: Oh yeah, Rider has options to only show C# files. It's more of a folder hierarchy thing, and limiting the number of folders called Scripts.
Screenshot 2022-02-04 at 03 50 33
Looking at this though it seems editor stuff is put in its own assembly anyway which it doesn't seem to do when part of the GG-JointJustice assembly.
Screenshot 2022-02-04 at 03 57 03
I think this is fine, and I'm kind of arguing against something we're not doing anyway lol. Feel free to mark this as resolved when you've read it.

unity-ggjj/Assets/Scripts/SaveFiles/SaveData.cs
PhoebeMitchell: Ah, now I think of this class in terms of a JSON object this whole thing makes a lot more sense to me. Carry on. I think this is good.

PhoebeMitchell (dismissed):
Looks good. Thanks for addressing my comments :D
Hope you don't mind but I took the liberty of updating the SaveFiles - TestScene. The old BGSceneList object seems to be reappearing where it shouldn't be even though I thought I'd deleted it, so I re-deleted it and updated the reference to it in SceneController, so I can approve without my review going stale.

unity-ggjj/Assets/Scripts/Savefiles/SaveData.cs
ViMaSter: Gonna close this because of https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/156#discussion_r799163871, hope that's alright. 👀

unity-ggjj/Assets/Scripts/SaveFiles/PlayerPrefsProxy.cs
```diff
@@ -0,0 +1,107 @@
...
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using UnityEngine;
+
+namespace SaveFiles
+{
+    public static class PlayerPrefsProxy
+    {
+        private const string PLAYER_PREFS_KEY = "SaveData";
+        public static bool HasExistingSaveData()
```
lyndsiWilliams, line 17: This method is missing a summary ✨

unity-ggjj/Assets/Scripts/SaveFiles/PlayerPrefsProxy.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            return PlayerPrefs.HasKey(PLAYER_PREFS_KEY);
+        }
+
+        public static void DeleteSaveData()
```
lyndsiWilliams, line 25: This method is missing a summary ✨

unity-ggjj/Assets/Scripts/SaveFiles/PlayerPrefsProxy.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            PlayerPrefs.DeleteKey(PLAYER_PREFS_KEY);
+        }
+
+        public static SaveData Load()
```
lyndsiWilliams, line 34: This method is missing a summary ✨

unity-ggjj/Assets/Scripts/SaveFiles/PlayerPrefsProxy.cs
```diff
@@ -0,0 +1,107 @@
...
+            // if the layout of SaveData is changed, manually grab old fields from `outdatedSettingsJSON`
+            // here and assign them to their new location
+            //////////
+
+            newSaveFile.Version = SaveData.LatestVersion;
+
+            return newSaveFile;
+        }
+
+        private static void Save(SaveData saveData)
```
lyndsiWilliams, line 106: This method is missing a summary ✨

unity-ggjj/Assets/Scripts/SaveFiles/PlayerPrefsProxy.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            PlayerPrefs.SetString(PLAYER_PREFS_KEY, JsonConvert.SerializeObject(saveData));
+            PlayerPrefs.Save();
+        }
+
+        public static void EnsureSaveDataExists()
```
lyndsiWilliams, line 115: This method is missing a summary ✨

unity-ggjj/Assets/Scripts/UI/SaveFileWatcher.cs
```diff
@@ -0,0 +1,37 @@
...
+    // Start is called before the first frame update
+    private void Start()
+    {
+        PlayerPrefsProxy.EnsureSaveDataExists();
+
+        _text = GetComponent<Text>();
+        StartCoroutine(Loop());
+    }
+
+    private IEnumerator Loop()
```
lyndsiWilliams, line 26: This method is missing a summary ✨

lyndsiWilliams (changes requested):
Hello it is I, the nit fairy coming to pick at all your summary-less methods 🧚 ✨

unity-ggjj/Assets/Scripts/UI/SaveFileWatcher.cs
ViMaSter: @lyndsiWilliams Added comments for all the method you mentioned!

---

If you've got the time, I'd love to hear your thoughts on https://github.com/Studio-Lovelies/GG-JointJustice-Unity/discussions/166!  
`SaveFiles.PlayerPrefsProxy.Save()` is a good example for a comment that only mirrors what the method body already describes using the same words and that - to me - only makes it worse if a comment is added:  
I my mind, since we have full source code access, the information gathered from the comment right now (that it serializes the parameter to JSON and persists it using `PlayerPrefs`), could be deducted by just looking at the method body (a press on F12 in any Visual Studio for example), so all that can possibly be added as a comment, is redundant and (as it's not compiled) additionally in danger of being forgotten to be updated, if the method body ever changes.  

This doesn't apply to all methods, but more on that in that discussion, just figured it's a good time to bring it up here! 👀

lyndsiWilliams: I think the only way I personally would be comfortable skipping methods would be if it were a _very stupidly obvious_ thing. For instance, while it seems obvious that `SaveFiles.PlayerPrefsProxy.Save()` does exactly as the name says, sometimes methods do a little more than what the name infers. I'd even be fine with summaries that say "This does exactly as the name says", I just like summaries as a reassurance.

lyndsiWilliams (approved):
This looks great, thanks for adding the summaries! Super hyped to get saving implemented! 😁

PhoebeMitchell (approved):

PR #180, opened 2022-02-07
feature/115 Case Select Menu

### LINKED ISSUE #115: Case Select Menu
## Description
A case selection menu needs to be added to the MainMenu scene. It should be easily adaptable and extendable in case its design is changed, and for when more cases are added.
Selecting a case should bring you to a part select screen. For now, base this off of the one in the LUA build (buttons across the bottom with a preview image). The buttons don't have to go to any particular scene yet as they don't exist.

## Acceptance Criteria

This story will be finished when the menu is demonstrably navigable without bugs.

PhoebeMitchell:
## Summary

Implements the case select menu. Buttons are created for each script provided on the CaseSelectMenu component. Currently mimics the design of the Lua build. Buttons do not lead anywhere yet.

menu2

## Testing instructions

Access the menu in the MainMenu scene. Try and break it. Run the tests. Do they still work?

## Additional information

- `[x]` Changes UI
  
closes #115


unity-ggjj/Assets/Scripts/Menu/MenuItem.cs
```diff
@@ -16,6 +17,9 @@ public class MenuItem : MonoBehaviour, ISelectHandler, IDeselectHandler, IPointe
     [Tooltip("Drag the menu item's text component here.")]
     [SerializeField] private TextMeshProUGUI _textMeshProUGUI;
     
+    [field: SerializeField] public UnityEvent OnItemSelect { get; private set; }
+    [field: SerializeField] public UnityEvent OnItemDeselect { get; private set; }
```
ViMaSter: These events seem only to be bound and invoked via C# (correct me if I missed anything! : D) and if that's the case, I think they could be changed to `Action` and the attribute `[SerializeField]` can also be dropped; that way they can only be bound in code and aren't rendered inside the Unity Inspector. I remember that being a good suggestion from you in the Discussions!

The term `Select` also feels weird to me, but I'm not sure if there's a better one. Instinctively, I would associate it with a confirmation (like a mouse button click or button press) rather than a mouse over, which this seems to be.  
Maybe a comment along the lines of `Invoked when the mouse cursor enters/leaves this element or arrow-keys are used to navigate to/away from it` works better? (I would've suggested renaming it to `OnMouseEnter` `OnMouseLeave` but since it also handles keyboard navigation that's not really appropriate either.)

unity-ggjj/Assets/Scripts/Menu/MenuOpener.cs
```diff
@@ -10,8 +11,8 @@
 /// </summary>
 public class MenuOpener : MonoBehaviour
 {
-    [SerializeField, Tooltip("Drag the menu controller of the menus to open or close here.")]
-    private Menu _menuToOpen;
+    [field: FormerlySerializedAs("_menuToOpen")] [field: SerializeField, Tooltip("Drag the menu controller of the menus to open or close here.")]
```
ViMaSter: Nit: Split these into individual attributes, each on their own line.

unity-ggjj/Assets/Scripts/Menu/CaseSelectMenu.cs
```diff
@@ -0,0 +1,56 @@
...
+                    menuOpener.OpenMenu();
+                    _chapterSelectMenu.Initialise(@case.Chapters, menuOpener);
+                }
+            });
+        }
+    }
+}
+
+[Serializable]
+public struct Case
```
ViMaSter: Nit: I think we never formerly agreed on something here, but I usually prefer to have each `class`, `struct` and `enum` be its own file (unless it's nested inside another class/struct); I think this would be a good idea here!

If you see any other options or disagree with it, I'm totally okay with keeping it like this inside this PR; just mark this as resolved here and open up a Discussions thread for it. That way we can also integrate it to the style guide once we've come to a conclusion.

unity-ggjj/Assets/Scripts/Menu/CaseSelectMenu.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// Subscribe instantiated buttons to update preview image
+    /// Add onClick logic to each button: either start the game or open the chapter select menu
+    /// </summary>
+    private void Awake()
+    {
+        var backButton = GetComponentInChildren<MenuItem>();
+        backButton.OnItemSelect.AddListener(() => _previewImage.color = Color.black);
+        backButton.OnItemDeselect.AddListener(() => _previewImage.color = Color.white);
+
+        foreach (var @case in _cases)
```
ViMaSter: Nit: Personally not that big of a fan of [the verbatim modifier](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/verbatim) in general. It's one of those language-specific "gotcha"s that can disrupt reading code quite a bit, in my experience.

Would changing the classes and variables to something like `NarrativeCase`/`NarrativeCases` instead work for you too?

unity-ggjj/Assets/Scripts/Menu/CaseSelectMenu.cs
```diff
@@ -0,0 +1,56 @@
...
+            {
+                if (@case.Chapters.Length == 1)
+                {
+                    // No need to show chapters, put logic for starting game here
+                }
+                else
+                {
+                    menuOpener.OpenMenu();
+                    _chapterSelectMenu.Initialise(@case.Chapters, menuOpener);
+                }
```
ViMaSter: Nit: Rather than commenting this, feel free to throw a `NotImplementedException` here.  
That way, builds crash and the Unity editor prints an error to the console if this point of the code is hit.  
It also reduces nesting, as the `else` becomes redundant.
```suggestion
                if (@case.Chapters.Length == 1)
                {
                    throw new NotImplementedException("Put logic for starting game here");
                }
                
                menuOpener.OpenMenu();
                _chapterSelectMenu.Initialise(@case.Chapters, menuOpener);
```

unity-ggjj/Assets/Scripts/Menu/CaseSelectMenu.cs
```diff
@@ -0,0 +1,56 @@
...
+        backButton.OnItemDeselect.AddListener(() => _previewImage.color = Color.white);
+
+        foreach (var @case in _cases)
+        {
+            var menuItem = Instantiate(_buttonPrefab, _buttonContainer);
+            menuItem.Text = @case.Name;
+            menuItem.OnItemSelect.AddListener(() => _previewImage.sprite = @case.PreviewImage);
+            var menuOpener = menuItem.GetComponent<MenuOpener>();
+            menuOpener.MenuToOpen = _chapterSelectMenu.Menu;
+            ((Button)menuItem.Selectable).onClick.AddListener(() =>
```
ViMaSter: Nit: This is just a suggestion, so I'm fine if you feel like it's overkill: How about we make the action listeners we're adding here `private static` methods of this class instead of inlining them and add them that way instead?

Right now, every action besides this one is fairly short, so this isn't as dramatic, but reading something like  
`((Button)menuItem.Selectable).onClick.AddListener(StartGameOrShowChapterSelection)` or  
`backButton.OnItemSelect.AddListener(HighlightButton);`  
would keep the method doing the bindings short, while maintaining the same level of legibility as inlining the actions, especially if the action listener grows in complexity.

Thinking about this a bit more, we could also have `Event Utility Classes`; something along the lines of
```cs
static class ButtonEvents {
    public void static AddHighlight() {...}
    public void static RemoveHighlight() {...}
    public void static AddSelected() {...}
    public void static RemoveSelected() {...}
}
```
which defines how a button highlight looks like **once** and then can be used by any UI code component that wants to highlight a button.

Curious to hear your thoughts on it!

ViMaSter (approved):
Looks good to me! Couple of questions and nits, but also perfectly happy if it goes in like that!

Naming things, callback management and things that would take a while to automatically test; I once again realize why I don't enjoy frontend programming, so thanks for doing it! : D

unity-ggjj/Assets/Scripts/Menu/MenuItem.cs
PhoebeMitchell: "Select" is Unity terminology for highlighting a Selectable, and these are called when the Selectable is Selected. I will add a tooltip to explain the purpose of these.

They're UnityEvents because I use them in #186 to play sounds when they are selected.

ViMaSter: Ah, that totally makes sense! I wasn't aware that "Selectables" was a Unity-specific term for UI navigation, thanks for clearing that up! :)

I think having that comment is still a good idea, but otherwise feel free mark this as resolved!

lyndsiWilliams (approved):
THE TESTS PASS! Also everything else looks great, great job on this 😁

unity-ggjj/Assets/Scripts/Menu/MenuItem.cs
PhoebeMitchell: Added tooltips for these events

unity-ggjj/Assets/Scripts/Menu/CaseSelectMenu.cs
PhoebeMitchell: Made the anonymous method into its own method. Not sure it makes sense to have the method be static as it uses the _chapterSelectMenu field.

Not sure there is need for the suggested ButtonEvents class yet either so have not implemented it. But could be useful for when starting the game gets implemented as both CaseSelectMenu, ChapterSelectMenu and the new game button will need this logic.

Highlighting is already handled with IHighlight components, which can be swapped out depending on the highlight we want.

PR #171, opened 2022-01-30
Feature/new court record menu

PhoebeMitchell:
## Summary

Adds new art for the court record menu, also known as the evidence menu.
Adjusts some logic related to displaying menu graphics (updating text, and button highlighting)
Also changes tests so they don't add a new device for each test, resulting in breaking the tests.

~~Has #146 merged so go review that first.~~

## Before/after screenshots and/or animated gif

Screenshot 2022-01-30 at 01 40 08

## Testing instructions

Run the tests.
Open the menu. Witness art.

## Additional information

- `[x]` Changes UI
- `[ ]` Introduces new feature
- `[ ]` Removes existing feature
- `[x]` Has associated resource: 
  
closes #131

PR #155, opened 2022-01-21
Replaced included default player actions with custom player actions

ViMaSter:
## Summary
This PR removes all references in our project to the `DefaultPlayerActions` that are included as an example inside the `Input System` package and instead uses the `PlayerActions` instance at `/Assets/Scripts/Input/Control.cs` were using for the rest of the game already.

### Unused UI input system actions
Unity's UI Input system consists of some behaviors we're currently not using. As we're not using them, I've deliberately left them unassigned:
- Middle Click
- Right Click
- Scroll Wheel
- Tracked Position and Orientation (not to be confused with the mouse position called `Point`, that's assigned; think a phone's gyro sensor or a Wiimote)

image

### `DirectionalButtons` vs. individual `Left`, `Right`, `Up`, `Down` actions
Our `Controls` had both `DirectionalButtons` **and** individual `Left`, `Right`, `Up`, `Down` actions. I've removed the individual actions as they a) seemed unused and b) can easily be accessed by listening to changes of the `DirectionalButtons` axis.  
It now looks like this:  
image

### `X` and `Enter`
`X` and `Enter` can now be used interchangeably. I haven't really seen the need to separate the two, but if we want to split "Dialogue actions" and "UI actions" later on, it's something we can always do.

## Testing instructions
1. Run any scene of the game
2. Make sure all actions previously executable are still executable; aside from `X` and `Enter` now able to be used interchangeably, no keybinds should have changed
3. Verify the EditMode and PlayMode tests still succeed

## Additional information

- `[x]` Changes UI
  - `X` and `Enter` key can now be used interchangeably 
- `[ ]` Introduces new feature
- `[ ]` Removes existing feature
- `[ ]` Has associated resource:
  - Closes #139

PR #179, opened 2022-02-06
Refactor/dialogue controller

PhoebeMitchell:
## Summary

This PR aims to shorten and humblify DialogueController. DialogueConstroller's functionality is now handled by NarrativeScriptPlayer (monobehaviour) and StoryPlayer (testable humble object). In the process #114 should be fixed, and evidence can now be presented using the ink macro `-> Present(evidenceName) ->`. #59 should also be implemented as part of this refactor. Menus now set an active InputModule (previously UnityInputSystemInterface) when they open and close.

This PR also attempts to fix script execution order problems by adding a `Game` monobehaviour which runs scripts in the correct order in its start method. Most of the "controller" objects now have a reference to this monobehaviour.

When fixing the tests I found ActionDecoder seemed to have reverted back to how it was before ActionDecoderBase was implemented. I assumed this was a problem with the recent repo structure changes, so it now inherits from ActionDecoderBase again.

## Testing instructions

Run the tests. Does the game work? Try all the scenes.

## Additional information

- `[ ]` Changes UI
- `[ ]` Introduces new feature
- `[ ]` Removes existing feature
- `[x]` Has associated resource: 
  
closes #114 
closes #177 
closes #59

PR #186, opened 2022-02-08
Feature/125 sounds on button select

### LINKED ISSUE #125: Sounds on button select
## Description
Buttons should play a sound when they are selected. The new game button also needs to play a unique sound when it is pressed, and should have a short delay before continuing to the next scene. This also applies to case select buttons (if they have been added yet). There is a button prefab. You should modify this so all buttons in the game are affected. Buttons in the evidence menu should also play the correct sounds. Use the Lua build for references on what sounds each button should play.

## Acceptance Criteria

This story will be finished when all buttons play the correct sounds.

PhoebeMitchell:
## Summary

Adds playing the sounds on button select and click. Sounds have been added to buttons on the main menu, case select menu, pause menu, and court record menu.

~~Has #155, #180, and #171 merged. Go review those first.~~
Everything merged, ready for review. Will probably want to merge #189 first though so review that if you can.

## Testing instructions

Access the menus listed above. Do sounds play as expected?

## Additional information

- `[ ]` Changes UI
  
closes #125


notexplosive (approved):
ez game

unity-ggjj/Assets/Scripts/Input/UnityInputSystemInterface.cs
```diff
@@ -74,7 +74,11 @@ void Controls.IUIActions.OnLeftMouseButton(InputAction.CallbackContext context)
     /// <param name="context"></param>
     void Controls.IUIActions.OnDirectionalButtons(InputAction.CallbackContext context)
     {
-        //Unused for now
+        var value = context.ReadValue<Vector2>();
+        if (context.performed && value != Vector2.zero)
+        {
+            _onDirectionalButtons.Invoke(value);
+        }
```
ViMaSter: Nit: I'm sure you already know what's coming at this point. : D
```suggestion
        if (!context.performed)
        {
            return;
        }
        
        if (value == Vector2.zero)
        {
            return;
        }
        _onDirectionalButtons.Invoke(value);
```

ViMaSter (approved):
Looks good to me. 👍

unity-ggjj/Assets/Scripts/Input/UnityInputSystemInterface.cs
PhoebeMitchell: Usually I'd agree, but I think in this case it needs to be specifically context.performed, as "!context.performed" could mean context.started and context.cancelled are true which would change the functionality. 

Saying that I realised I should use context.started instead of checking if the direction is the zero vector.

btw also added sounds to the pause menu and evidence menu buttons because they were missing for some reason, if you wanted to confirm they work.

PhoebeMitchell:
Please nobody merge this or resolve the conversation above. Waiting for #179 which contains a refactor of the input logic which is needed in this branch to fix errors in tests.

PR #189, opened 2022-02-12
Refactor/dialogue controller

### LINKED ISSUE #59: Enable and Disable controls based on system state
## Description

When the pause menu or evidence menu is open, the dialogue in the background should no longer respond to input.

## Dev guide

Disable the object responsible for the input when any menu is open, and enable it again when closing said menu.

## Acceptance Criteria

This story will be finished when the aformentioned functionality is implemented, added to the standard inclusion prefab, and testable in a unity scene.

### LINKED ISSUE #114: [BUG] Cannot present evidence when PRESENT_EVIDENCE action is called

### LINKED ISSUE #177: [BUG] DialogueController starts playing before BGScenes are loaded
## Describe the bug
BGScenes are created in the Start method of BGScenesList. However DialogueController starts playing in its start method, and relies on BGScenes being loaded, or the &SCENE action will tell BGScenesList to load a BGScene that has not been created yet. 

## Steps To Reproduce

Open Inky-TestScene and run a narrative script where the first action is a &SCENE action e.g., MissingScenesDemo. You may or may not get the following error from BGScenesList.

Screenshot 2022-02-04 at 04 19 11

Order of execution varies within Unity so it may work, it may not.

## Expected behavior
The order of execution must be NarrativeScriptPlaylist's Awake method -> BGSceneList's Start method -> DialogueController's Start method.

## Additional context
If you can do this _without_ setting the order of script execution in Unity that would be good. Personally I feel setting the script execution order is a bit of a dirty fix and is very hard to debug if the order of execution is changed for whatever reason. I feel it'd be better to write the code in a way that indicates the order in which it is run.

PhoebeMitchell:
## Summary

This PR aims to shorten and humblify DialogueController. DialogueConstroller's functionality is now handled by NarrativeScriptPlayer (monobehaviour) and StoryPlayer (testable humble object). In the process #114 should be fixed, and evidence can now be presented using the ink macro `-> Present(evidenceName) ->`. #59 should also be implemented as part of this refactor. Menus now set an active InputModule (previously UnityInputSystemInterface) when they open and close.

This PR also attempts to fix script execution order problems by adding a `Game` monobehaviour which runs scripts in the correct order in its start method. Most of the "controller" objects now have a reference to this monobehaviour.

When fixing the tests I found ActionDecoder seemed to have reverted back to how it was before ActionDecoderBase was implemented. I assumed this was a problem with the recent repo structure changes, so it now inherits from ActionDecoderBase again.

Massive issues when updating this to develop, so please review thoroughly :)

## Testing instructions

Run the tests. Does the game work? Try all the scenes.

## Additional information

- `[ ]` Changes UI
  
closes #114 
closes #177 
closes #59


unity-ggjj/Assets/Scripts/Game/Game.cs
```diff
@@ -0,0 +1,44 @@
+using UnityEngine;
+
+public class Game : MonoBehaviour
```
ViMaSter: Nit: I find `Game` to be a pretty broad term. I'd suggest changing it to something like `GameState` or - more specifically `NarrativeGameState`. That way, if we ever need similar classes for other parts of the game, we can have classes like `ExplorationGameState` or `CoolMinigameGameState`.

unity-ggjj/Assets/Scripts/Game/Game.cs
```diff
@@ -0,0 +1,44 @@
...
+        _bgSceneList.InstantiateBGSceneFromPlaylist(_narrativeScriptPlaylist);
+        _directorActionDecoder.Decoder.NarrativeScriptPlayer = _narrativeScriptPlayer;
+
+        _directorActionDecoder.Decoder.ActorController = _actorController;
+        _directorActionDecoder.Decoder.AudioController = _audioController;
+        _directorActionDecoder.Decoder.EvidenceController = _evidenceController;
+        _directorActionDecoder.Decoder.SceneController = _sceneController;
+        _directorActionDecoder.Decoder.PenaltyManager = _penaltyManager;
+        _directorActionDecoder.Decoder.AppearingDialogueController = _appearingDialogueController;
+        _directorActionDecoder.Decoder.NarrativeScriptPlayer = _narrativeScriptPlayer;
```
ViMaSter: I'm wondering if it would make more sense, to simply have the `DirectorActionDecoder` have a reference to this class <sup>(or an interface of it)</sup> and then add public members to this class <sup>(or the interface of it)</sup> to encompass all controllers or feature classes, similarly to how it's done in classes like `EvidenceMenu` after this change. Any opinions from you?

unity-ggjj/Assets/Scripts/Interfaces/IAudioController.cs
```diff
@@ -1,6 +1,9 @@
+using UnityEngine;
+
 public interface IAudioController
 {
-    void PlaySfx(string SFX);
+    void PlaySfx(string sfx);
```
ViMaSter: Nit: Not sure if we've ever added it to the coding style guide, but I think abbreviations should be either be all uppercase or all lowercase. (`SFX` instead of `Sfx`)

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayer.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    /// <summary>
+    /// Checks if a witness can be pressed then chooses the correct choice on StoryPlayer
+    /// </summary>
+    public void PressWitness()
+    {
+        if (CanPressWitness)
+        {
+            StoryPlayer.HandleChoice(1);
+        }
```
ViMaSter: Nit: I'd recommend throwing an exception, if this is called while `CanPressWitness` is false. Silently not executing logic can be difficult to debug and `CanPressWitness` exists and is publicly available for other classes to wait on.

unity-ggjj/Assets/Scripts/Input/InputModule.cs
```diff
@@ -140,4 +142,12 @@ void Controls.IUIActions.OnCursor(InputAction.CallbackContext context)
...
         //Unused for now
     }
+
+    /// <summary>
+    /// This is object as the active input
+    /// </summary>
+    public void SetAsActiveInput()
+    {
+        _inputManager.SetInput(this);
+    }
```
ViMaSter: Nit: I think this method can be made obsolete: It's possible to directly call `SetInput`on the `InputManager` with the input module to choose from a UnityEvent:
image
As they're both public methods covering the exact same logic, I think it would be a good idea to limit them to only one occurrence.

unity-ggjj/Assets/Scripts/Input/InputManager.cs
```diff
@@ -0,0 +1,22 @@
...
+    private void Awake()
+    {
+        _inputModules = GetComponentsInChildren<InputModule>().ToList();
+    }
+
+    /// <summary>
+    /// Sets a specified input module to enabled and disables all others
+    /// </summary>
+    /// <param name="inputModule"></param>
+    public void SetInput(InputModule inputModule)
```
ViMaSter, line 18: This is _suuuuuch_ a nice way to handle input and another great example of a [state machine](https://gameprogrammingpatterns.com/state.html)! Additionally it also makes things in the Editor super readable. 👍

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayer.cs
```diff
@@ -0,0 +1,64 @@
...
+        set => StoryPlayer.GameMode = value;
+    }
+
+    public bool Waiting
+    {
+        get => StoryPlayer.Waiting;
+        set => StoryPlayer.Waiting = value;
+    }
+    public bool CanPressWitness => StoryPlayer.CanPressWitness && !Waiting;
+    public bool HasSubStory => StoryPlayer.HasSubStory;
```
ViMaSter: Nit: I think this property should be safe to be inlined (have all classes that use `NarrativeScriptPlayer.HasSubStory` directly use `NarrativeScriptPlayer.StoryPlayer.HasSubStory` instead).

unity-ggjj/Assets/Scripts/NarrativeScript/StoryPlayer.cs
```diff
@@ -0,0 +1,214 @@
...
+    {
+        if (_subStory != null)
+        {
+            _subStory.PresentEvidence(courtRecordObject);
+            return;
+        }
+
+        if (!IsAtChoice || GameMode != GameMode.CrossExamination)
+        {
+            return;
```
ViMaSter: Nit: I think this should actually throw a `NotSupportedException`: `EvidenceMenu.OnEvidenceClicked()` (correctly) uses `CanPresentEvidence` to stop execution before calling this, so if this method is ever entered without being at a choice or cross examination, we've ran into an unexpected state of the game.

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayer.cs
```diff
@@ -0,0 +1,64 @@
+using UnityEngine;
+
+public class NarrativeScriptPlayer : MonoBehaviour, INarrativeScriptPlayer
```
ViMaSter: This is a super clean implementation of the ["keep objects humble"](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/wiki/Code-Style-Guideline#:~:text=When%20updating%20existing%20functionality%2C%20preferably%20de%2Dcouple%20custom%20logic%20from%20Unity%2Dspecific%20functionality%20unless%20strictly%20required)-rule we've agreed on! Real nice job! 👍

---

Nit: I'm actually starting to wonder if it would make sense to introduce a global pre- or suffix for classes that serve as a Unity-adapter for another pure class (like `StoryPlayer` in this case)...  

Like with `DirectorActionDecoder` -> `ActionDecoder`, `NarrativeScriptPlayer`->`StoryPlayer` this class hasn't got any `Narrative`-specific logic that makes this class different from its underlying logic. Instead it only serves as an adapter for Unity. Additionally the name of the underlying humble class is obfuscated (`Script` -> `Story`) too.

Maybe something along the lines of `MonoAdapter` could work?
- `ActionDecoderMonoAdapter`
- `StoryPlayerMonoAdapter`

I could also see namespaces being useful here; having humble objects and MonoBehavior-adapters inside a separate namespaces for example. That said, I remember Unity not playing nice with classes with the same name in different namespaces. 

(Imagine if we could even go as far as splitting up our "game logic" assembly into a `Game` Assembly Definition with **no** ties to UnityEngine and UnityEditor and then have an `Adapter` Assembly Definition which only makes them Unity-editor and Mono-friendly! 👀)

Ideally it's something we can add to our Style Guide too... Curious if you have any suggestions or ideas!

unity-ggjj/Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
```diff
@@ -29,7 +28,7 @@ public void OnNewActionLine(string line)
         catch (TextDecoder.Parser.ScriptParsingException exception)
         {
             Debug.LogError(exception);
-            _onActionDone.Invoke();
+            _narrativeScriptPlayer.Continue();
```
ViMaSter: I believe this entire `catch` block (and wrapping the code in question in a `try/catch`-block in general) is actually no longer necessary. If we encounter a line that fails to parse, I would expect the game to stop.

Right now we would silently continue the game, which I think is a bit dangerous, given we've failed to parse an action line. (According to Git I actually added this but I'm not entirely sure why we needed it in the first place. 🤔)

unity-ggjj/Assets/Scripts/NarrativeScript/StoryPlayer.cs
```diff
@@ -0,0 +1,214 @@
...
+        }
+        else
+        {
+            _appearingDialogueController.PrintText(nextLine);
+        }
+    }
+
+    /// <summary>
+    /// Checks if a story can continue, and handles what happens
+    /// it cannot, depending on the current GameMode
```
ViMaSter: Nit: Missing word.
```suggestion
    /// if it cannot, depending on the current GameMode
```

unity-ggjj/Assets/Scripts/NarrativeScript/StoryPlayer.cs
```diff
@@ -0,0 +1,214 @@
...
+    /// </summary>
+    /// <returns>If the story can continue (true) or (not)</returns>
+    private bool HandleCannotContinue()
```
ViMaSter: Nit: This method is only ever called in one location inside this class; I think inlining this will make it easier to follow the flow of `Continue()`.

unity-ggjj/Assets/Scripts/Game/Game.cs
```diff
@@ -0,0 +1,44 @@
...
+    [SerializeField] private AppearingDialogueController _appearingDialogueController;
+    [SerializeField] private PenaltyManager _penaltyManager;
+    [SerializeField] private BGSceneList _bgSceneList;
+    [SerializeField] private ChoiceMenu _choiceMenu;
+
+    public IActorController ActorController => _actorController;
+    public IAppearingDialogueController AppearingDialogueController => _appearingDialogueController;
+    public IObjectStorage ObjectStorage => _narrativeScriptPlayer.ActiveNarrativeScript.ObjectStorage;
+    public INarrativeScriptPlayer NarrativeScriptPlayer => _narrativeScriptPlayer;
+    public IAudioController Audio => _audioController;
```
ViMaSter: Nit: Other controllers (`ActorController` + `AppearingDialogueController`) also have the `Controller` suffix so I think I'd prefer keeping it `AudioController` in this case too out of consistency.

unity-ggjj/Assets/Scripts/UI/PenaltyManager.cs
```diff
@@ -55,12 +39,24 @@ public void OnCrossExaminationEnd()
...
+
+    /// <summary>
+    /// Creates penalties equal to the value of _penaltyCount
+    /// </summary>
+    private void CreatePenalties()
+    {
+        for (int i = 0; i < _penaltyCount; i++)
+        {
+            _penaltyObjects.Enqueue(Instantiate(_penaltyObject, transform));
+        }
```
ViMaSter: Nit: This method is only ever called in one location inside this class and the logic is pretty simple; I think inlining this will make it easier to follow the flow of `ResetPenalties()`.

unity-ggjj/Assets/Scripts/UI/PenaltyManager.cs
```diff
@@ -17,29 +15,15 @@ public class PenaltyManager : MonoBehaviour, IPenaltyManager
 
     private readonly Queue<Animator> _penaltyObjects = new Queue<Animator>();
 
-    public int PenaltiesLeft => _penaltyObjects.Count;
+    public int PenaltiesLeft { get; private set; }
```
ViMaSter, line 19: Nothing actionable, out of pure curiosity: Any reason this has received an underlying field instead of returning `_penaltyObjects.Count`?

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/Visibility/ViaKeyboard.cs
```diff
@@ -1,65 +1,52 @@
 ﻿using System.Collections;
 using Tests.PlayModeTests.Tools;
 using UnityEngine;
+using UnityEngine.Assertions;
 using UnityEngine.TestTools;
 using Object = UnityEngine.Object;
 
 namespace Tests.PlayModeTests.Scenes.VisibilityTest
 {
     public class ViaKeyboard
```
ViMaSter, line 10: Nit: Given the fact, that most "ViaKeyboard" tests no longer make use of the keyboard but use `StoryProgresser`, I think this class name can be updated and/or the classes entire file moved to `Tests.PlayModeTests.Scripts.StoryProgresser` instead.

This also applies to `unity-ggjj/Assets/Tests/PlayModeTests/Scenes/CrossExamination/ViaKeyboard.cs` and `unity-ggjj/Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuKeyboardTests.cs`.

That aside: Super happy to see this move to more of a component/unit test instead of a full integration test. Both with Unity's Input System being pretty flaky inside tests and aiming to focus our tests on our code, rather than libraries or Unity engine internals. Great update! 👍

unity-ggjj/Assets/Tests/PlayModeTests/Scripts/AppearingDialogueController/GetDelay.cs
```diff
@@ -59,14 +60,16 @@ private IEnumerator OneTimeSetup()
 
             // create controller component
             _appearingDialogueController = gameObjectInstance.AddComponent<global::AppearingDialogueController>();
+            LogAssert.Expect(LogType.Exception, "NullReferenceException: Object reference not set to an instance of an object");
```
ViMaSter, line 63: Nit: I think this exception is a result of `AppearingDialogueController.Start()` being changed to `AppearingDialogueController.Awake()`. Would it be possible to keep `AppearingDialogueController.Start()` to remove the exception inside this test or would that require a lot of additional changes?

unity-ggjj/Assets/Scripts/Actor/ActorController.cs
```diff
@@ -264,7 +238,8 @@ public void AssignActorToSlot(string actor, int oneBasedSlotIndex)
 
         try
         {
-            var actorData = _dialogueController.ActiveNarrativeScript.ObjectStorage.GetObject<ActorData>(actor);
+
+            var actorData = _game.ObjectStorage.GetObject<ActorData>(actor);
```
ViMaSter: Nit: Empty line.
```suggestion
            var actorData = _game.ObjectStorage.GetObject<ActorData>(actor);
```

unity-ggjj/Assets/Scripts/BGScenes/BGSceneList.cs
```diff
@@ -3,28 +3,18 @@
...
-    [SerializeField] private NarrativeScriptPlaylist _narrativeScriptPlaylist;
-    
     private readonly Dictionary<string, BGScene> _bgScenes = new Dictionary<string, BGScene>();
     private BGScene _activeScene;
 
     /// <summary>
     /// Gets all available BGScenes and instantiates them.
     /// </summary>
-    private void Start()
+    public void InstantiateBGSceneFromPlaylist(INarrativeScriptPlaylist narrativeScriptPlaylist)
```
ViMaSter, line 12: Nit: Given the "monoscene" idea we're trying to shoot for, I think having this method only load the scenes for the currently active script and sub-scripts of it (instead of the entire playlist) would make more sense.  
It's not much of a concern at the moment though, given we're only loading two chapters at most, but maybe this could result in a lot resources being loaded in memory but only being relevant when the respective chapter is started.

unity-ggjj/Assets/Scripts/BGScenes/SceneController.cs
```diff
@@ -37,19 +33,6 @@ public class SceneController : MonoBehaviour, ISceneController
...
-    [SerializeField] private UnityEvent _onWaitStart;
-
-    [Tooltip("This event is called when a wait action is finished.")]
-    [SerializeField] private UnityEvent _onWaitComplete;
-
-    [Tooltip("Event that gets called when the actor displayed on screen changes")]
-    [SerializeField] private UnityEvent<Actor> _onActorChanged;
-
-    [Tooltip("Event that gets called when the active bg-scene changes")]
-    [SerializeField] private UnityEvent<BGScene> _onSceneChanged;
```
ViMaSter, line 52: In the glorious words of NSP: Every time an event is gone, an angel gets its thing. Love to see it. 👍

unity-ggjj/Assets/Scripts/Interfaces/IEvidenceController.cs
```diff
@@ -1,9 +1,13 @@
...
+    List<Evidence> CurrentEvidence { get; }
+
+    void AddEvidence(Evidence evidence);
     void RemoveEvidence(string evidenceName);
-    void AddToCourtRecord(string actorName);
+    void AddToCourtRecord(ActorData actor);
     void RequirePresentEvidence();
-    void SubstituteEvidence(string initialEvidenceName, string substituteEvidenceName);
-    void OnPresentEvidence(ICourtRecordObject evidence);
+    void SubstituteEvidence(string initialEvidenceName, Evidence substituteEvidence);
```
ViMaSter: Minor nit: I think unifying all method names of this interface that modify available evidence would be useful.

I could see something along the lines of `AddToCourtRecord(Evidence evidence)/RemoveFromCourtRecord(ActorData actor)` and only having the parameters differ working. Or instead something more direct like `AddEvidence` and `AddActor` for both types of court record entries.

ViMaSter: Nit: Is there any particular reason, only some of the methods have been changed to use instances of the data classes directly? 

Personally, I think having those classes use data objects directly is actually a step in the right direction, I simply would prefer if all methods inside this controller use the same level of abstraction; **either** names of objects **or** direct instances.

unity-ggjj/Assets/Scripts/BGScenes/SceneController.cs
```diff
@@ -364,4 +332,14 @@ public void ReloadScene()
     {
         _sceneLoader.LoadScene(SceneManager.GetActiveScene().name);
     }
+
+    /// <summary>
+    /// Sets NarrativeScriptPlayer to no longer wait,
+    /// and continues the story
+    /// </summary>
+    private void WaitComplete()
```
ViMaSter: Nit: As there's various places inside this class where only `_game.NarrativeScriptPlayer.Waiting = false;` is used directly, I think having this method name be a bit more descriptive (like `ContinueStory()`) to clarify how it differs from just setting `Waiting` to `false` would be a good idea. (Also I always like starting method names with verbs.)  
At first glance it wasn't clear to me, how setting `Waiting = false` and `WaitComplete()` differ, until I looked at the method's body.

unity-ggjj/Assets/Scripts/Game/Game.cs
```diff
@@ -0,0 +1,44 @@
...
+    public IEvidenceController EvidenceController => _evidenceController;
```
ViMaSter: More food for thought than an action item: This shift of having every controller or feature class accessible through this class is a pretty big change. Mind you, having [game state classes](https://gameprogrammingpatterns.com/state.html) like this is now, definitely isn't unusual and there's also nothing inherently wrong with it. Still, it's an idea I need to warm up to, given our previous architecture.  
The main argument that keeps coming back up in my mind, is that all classes still interact with the same interfaces as before, it's just that they now additionally proxy though this game state class which - on paper - in itself only introduces another level of abstraction.

That aside, so far it has resulted in less code and more humble classes though; so given that, I think I'm actually a fan! Curious to see, where this new approach will lead us as this project grows! 👀

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/StoryPlayer.cs
```diff
@@ -0,0 +1,214 @@
+using System.ComponentModel;
+using System.Linq;
+using Ink.Runtime;
+
+public class StoryPlayer
```
ViMaSter: Nit: I think a quick summary of **why** this class exists (not how it does things) and how it fits into the rest of the game's architecture would be a good idea; things like what this class is responsible for (and also what it deliberately isn't!) and the motivation for creating it. These things can usually serve as pretty good guidance, if we come across it after not working on it for a while.

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/StoryPlayer.cs
```diff
@@ -0,0 +1,214 @@
...
+        }
+        else
+        {
+            _appearingDialogueController.PrintText(nextLine);
+        }
+    }
+
+    /// <summary>
+    /// Checks if a story can continue, and handles what happens
+    /// it cannot, depending on the current GameMode
```
ViMaSter: Nit: Missing word
```suggestion
    /// if it cannot, depending on the current GameMode
```

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/StoryPlayer.cs
```diff
@@ -0,0 +1,214 @@
...
+    /// </summary>
+    /// <param name="choiceIndex">The index of the choice to choose</param>
+    public void HandleChoice(int choiceIndex)
+    {
+        Story.ChooseChoiceIndex(choiceIndex);
+        Continue();
+    }
+
+    /// <summary>
+    /// Starts a sub-story which will run in place of the parent story until it ends or is stopped
```
ViMaSter: Nit: Could this comment be extended a bit to also contain info about by which classes or methods a sub-story can be "stopped"?

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/StoryPlayer.cs
```diff
@@ -0,0 +1,214 @@
...
+        var currentChoices = Story.currentChoices;
+        var choice = currentChoices.FirstOrDefault(choice => choice.text == courtRecordObject.InstanceName);
+        if (choice != null)
+        {
+            HandleChoice(choice.index);
+        }
+        else
+        {
+            StartSubStory(_narrativeScriptPlaylist.GetRandomFailureScript());
+        }
```
ViMaSter: Nit: You know me... 👀
```suggestion
        if (choice == null)
        {
            StartSubStory(_narrativeScriptPlaylist.GetRandomFailureScript());
            return;
        }
        
        HandleChoice(choice.index);
        
```

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlaylist/INarrativeScriptPlaylist.cs
```diff
@@ -0,0 +1,10 @@
+using System.Collections.Generic;
+
+public interface INarrativeScriptPlaylist
+{
+    public void InitializeNarrativeScripts();
```
ViMaSter: Nit: The `public` keyword is redundant on interface members, as it can contain no method bodies where `private` or `protected` members could be accessed from.  
(If you do require private methods, this could be achieved by using an `abstract class`.)
```suggestion
    void InitializeNarrativeScripts();
```

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlaylist/INarrativeScriptPlaylist.cs
```diff
@@ -0,0 +1,10 @@
...
+    
+    NarrativeScript GetNextNarrativeScript();
+    IEnumerable<NarrativeScript> GetAllNarrativeScripts();
+    NarrativeScript GetRandomFailureScript();
```
ViMaSter: Super mega ultra nit: I think grouping these methods by return type makes things a tiny bit easier to read.
```suggestion
    NarrativeScript GetNextNarrativeScript();
    NarrativeScript GetRandomFailureScript();
    IEnumerable<NarrativeScript> GetAllNarrativeScripts();
```

unity-ggjj/Assets/Scripts/TextDecoder/Preloading/IObjectStorage.cs
```diff
@@ -3,6 +3,6 @@
 
 public interface IObjectStorage
 {
-    public T GetObject<T>(string objectName) where T : Object;
-    public IEnumerable<T> GetObjectsOfType<T>() where T : Object;
+    public T GetObject<T>(string objectName) where T : class;
```
ViMaSter, line 6: Nit: Is there any specific reason for the limitation to reference types on this method? I would suggest either removing it completely or also enforcing it on `GetObjectsOfType<T>()` for consistency.

unity-ggjj/Assets/Tests/EditModeTests/ActionDecoderEnum.cs
```diff
@@ -6,47 +6,38 @@
...
-            SceneController = new Moq.Mock<ISceneController>().Object,
-        };
-    }
-
     [Test]
     public void ParseActionLineWithValidEnumValue()
     {
-        ActionDecoder decoder = CreateMockedActionDecoder();
+        const string ACTOR_DATA_NAME = "NewActorData";
+        ActionDecoder decoder = new ActionDecoder();
```
ViMaSter: Nit: If this is moved closer to it's usage (like line 21), you could use an object initializer:
```cs
ActionDecoder decoder = new ActionDecoder {
    SceneController = sceneControllerMock.Object,
    NarrativeScriptPlayer = narrativePlayerMock.Object
};
```
<sup>GitHub doesn't allow suggestions, because line 22 is identical. :(</sup>

unity-ggjj/Assets/Tests/EditModeTests/ActionDecoderEnum.cs
```diff
@@ -6,47 +6,38 @@
...
         Mock<ISceneController> sceneControllerMock = new Moq.Mock<ISceneController>();
```
ViMaSter: Nit: All types inside this method have been to changed to `var`, so I would suggest updating its type here too. Additionally, the namespace qualifier is redundant.
```suggestion
        var sceneControllerMock = new Mock<ISceneController>();
```

This and line 15 could also be moved under the
```cs
        var actorData = ScriptableObject.CreateInstance<ActorData>();
        actorData.name = ACTOR_DATA_NAME;
```
block, that way they're only initialized right before usage:

```cs

        const string ACTOR_DATA_NAME = "NewActorData";
        var actorData = ScriptableObject.CreateInstance<ActorData>();
        actorData.name = ACTOR_DATA_NAME;

        var sceneControllerMock = new Moq.Mock<ISceneController>();
        var narrativePlayerMock = new Mock<INarrativeScriptPlayer>();
        narrativePlayerMock.Setup(mock => mock.ActiveNarrativeScript.ObjectStorage.GetObject<ICourtRecordObject>(ACTOR_DATA_NAME)).Returns(actorData);
        sceneControllerMock.Setup(controller => controller.ShowItem(actorData, ItemDisplayPosition.Left));

        ActionDecoder decoder = new ActionDecoder {
            SceneController = sceneControllerMock.Object,
            NarrativeScriptPlayer = narrativePlayerMock.Object
        };

        ...
```

This also applies to `ParseActionLineWithInvalidEnumValue()`.

unity-ggjj/Assets/Tests/PlayModeTests/Scripts/PauseMenu/ViaKeyboard.cs
```diff
@@ -1,4 +1,5 @@
 using System.Collections;
+using UnityEngine;
```
ViMaSter: Nit: I think this `using`-directive should be safe to remove, as there have only been whitespace-changes to this file.
```suggestion
```

ViMaSter (commented):
This is... really nice:
- less verbose testing code
- more humble objects <sup>(I need to remember that term...)</sup>
- more abstractions of game logic away from Unity
- even a percent more of code coverage

I think overall this is a change moving us in the right direction!  

Also a lot of scary big changes (now I know how y'all felt when I updated the `ActionDecoder`! : D), but the more I think about them, the more I like 'em.

A few open questions and points for discussions I'd like to see some form of resolution for, before giving approval, but the overall architecture change is something I definitely can get behind and both see the need for and gains in doing it. 

Not technically tied to the changes themselves, but the PR summary is also super concise! Used it quite often to go back and look at the overarching goals of this PR, which helped me understand some of the technical decisions made.

This shouldn't come as a surprise, but great work, as usual! 👍

unity-ggjj/Assets/Scripts/Interfaces/IAudioController.cs
PhoebeMitchell: You mean like `Playsfx`? I think the S should be a capital because it's PascalCase

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayer.cs
PhoebeMitchell: The player can press C button at any time, and it's only supposed to press the witness at certain times. It's intended functionality, so not sure throwing an exception is right.

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayer.cs
PhoebeMitchell: Agreed. Couldn't think of a better name so just went with StoryPlayer. StoryPlayer was supposed to be NarrativeScriptPlayer but I wanted that name to be the part you see in Unity. How about NarrativeScriptPlayerComponent, as its the component that holds NarrativeScriptPlayers?

unity-ggjj/Assets/Scripts/Interfaces/IAudioController.cs
ViMaSter: I actually really meant all uppercase for abbreviations. : D  
This would be in line with [Unity's own naming convention](https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnGUI.html).

Generally I would expect things like `RenderUI()`, `RefreshAIController()` or `UnlockIAPContent()`, so to match that I think I would use `PlaySFX()`.  
Alternatively, we could also use the full word instead, like `PlaySoundEffect()`.

PhoebeMitchell: We've mostly been going by the C# conventions described here
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions
PlaySfx would be appropriate according to that.

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayer.cs
ViMaSter: Ah, I see. This class is directly called by a Unity event, so there's no layer of abstraction left to catch this.  
In that case could we change this to `AttemptPressWitness()`/`TryPessWitness()`? That way I'd expect that this method may not actually do anything if certain conditions aren't met.

unity-ggjj/Assets/Scripts/Interfaces/IAudioController.cs
ViMaSter: > A special case is made for two-letter acronyms in which both letters are capitalized, as shown in the following identifier: `IOStream`

I bet that someone at Microsoft accidentally pushed `System.IO` to a release version and then had to explain why it's the only abbreviation that's all capitalized. : D  
But fair enough, that works for me then! (Even if I find the 2-letter-exception to be a bit odd.)

unity-ggjj/Assets/Scripts/BGScenes/SceneController.cs
notexplosive: 🎵 boop boop boop boop! 🎵  it's the Function Names Boy coming at you with another diatribe about identifier naming.

If a function name is ambiguous that ambiguity often comes from the name being dishonest (does not describe the behavior) or incomplete (only describes a subset of its behavior).

The first step towards a great name is something something honest and complete, that just spells out the exact behavior of the function (this gets verbose real fast).

`SetWaitingToFalseAndContinue` is an honest and complete name for this function. It tells you all the things its doing.

Then we can get a bit more domain-aware.

"Set waiting to false" feels like "FinishWaiting" -> `FinishWaitingAndContinue` like a terse-but-still-honest name.

Not saying we have to pick the name I chose, I'm just trying to spread my "naming as a process" philosophy.

🎵 boop boop boop boop 🎵

unity-ggjj/Assets/Scripts/BGScenes/SceneController.cs
```diff
@@ -364,4 +333,14 @@ public void ReloadScene()
...
+    {
+        _narrativeGameState.NarrativeScriptPlayer.Waiting = false;
+        _narrativeGameState.NarrativeScriptPlayer.Continue();
+    }
```
notexplosive: I spotted some [feature envy](https://refactoring.guru/smells/feature-envy)! 👀 

I think NarrativeScriptPlayer should own this function. My only evidence is when I see

```cs
class A{
   B b;

   void SomeFunction(){
      b.DoThing();
      b.DoOtherThing();
   }
}
```

That means there's feature envy, either `DoThing` and `DoOtherThing` should be moved to A, or (more likely) `SomeFunction` should be moved to B.

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayer.cs
notexplosive: Chiming in: If we're gonna continue the pattern of having `HumbleThing` -> `UnityThing` it would be nice to have some convention around it. I think `-Component` suffix is good! It differentiates the raw names (so unity doesn't cry) while clarifying what's mapped to what.

I think I like `-Component` a little better than `-Adapter` because Component is already an understood noun in the Unity world where Adapter could mean a lot of things... idk if that's a compelling reason but that's just how my brain is.

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayer.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    public StoryPlayer StoryPlayer { private get; set; }
+    public INarrativeScript ActiveNarrativeScript => StoryPlayer.ActiveNarrativeScript;
+
+    public GameMode GameMode
+    {
+        get => StoryPlayer.GameMode;
+        set => StoryPlayer.GameMode = value;
```
notexplosive: If StoryPlayer is publicly `get`able then should we have this proxy-property? (same for Waiting below)

(I write code like this too so I'm wondering out loud if it's good)

More generally: Shouldn't anything that wants to talk to the StoryPlayer just talk to it directly rather than talking to this Component as a proxy? I notice basically everything in this component is just forwarding to the StoryPlayer. Is there something I'm missing?

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlaylist/NarrativeScriptPlaylist.cs
```diff
@@ -51,6 +52,18 @@ public NarrativeScript GetRandomFailureScript()
...
+        var failureScript = FailureScripts[Random.Range(0, FailureScripts.Count)];
+        failureScript.Reset();
+        return failureScript;
+    }
+
+    /// <summary>
+    /// Returns all the narrative scripts in this playlist
+    /// Includes FailureStates and GameOverScript
+    /// </summary>
+    /// <returns>A collection of all narrative scripts in the playlist</returns>
```
notexplosive, line 64: nit: Would it make sense to move this summary to the interface? I know that might leave the illusion of a method missing it's summary but I think that would ultimately make it show up in more places... right?

unity-ggjj/Assets/Scripts/SceneLoading/FadeToImageTransition.cs
```diff
@@ -33,7 +33,7 @@ private void Start()
     {
         if (_shouldFadeInOnAwake)
         {
-            _imageFader.StartFade(1, 0, _fadeTime, _onTransitionInComplete);
+            _imageFader.StartFade(1, 0, _fadeTime, () => _onTransitionInComplete.Invoke());
```
notexplosive: How is this different than the old behavior? Doesn't calling an Action do the same thing as `Invoke`ing it?

Also did you mean to say `?.Invoke()`?

notexplosive (approved):
i did it

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayer.cs
ViMaSter: > If StoryPlayer is publicly getable then should we have this proxy-property? (same for Waiting below)

Just as an aside: `StoryPlayer` (currently) is publicly `set`able, but only privately `get`able. 👀
https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/b5ac0c73f180667897c44e4cf675e3edcac60271/unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayer.cs#L5
Given that, I think it just serves as an adapter which only exposes methods and properties that really need to be used by Unity-internal components or rendered in the Editor; anything else gets hidden.

Although I would like having these "Unity adapter classes" only expose the underlying object and its public interface (and thereby have "non-Unity classes" and the Unity Editor share the same level of interface visibility) and it also would be an even simpler approach, I'm not sure if that's feasible for all classes we'll have.

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayer.cs
PhoebeMitchell: ~~Is there a reason for this? Personally I feel it's cleaner for things to just access this property.~~

Removed all the accessor stuff from this class. Only stuff required by Unity remains.

unity-ggjj/Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
PhoebeMitchell: If we remove it, the game will essentially softlock. Maybe we should program in some kind of error screen if the catch block is reached. I think this should be a separate issue though. It's been kind of useful to have this feature tbh as it means i can put actions in narrative scripts which haven't been added yet, though that's admittedly less relevant now.

unity-ggjj/Assets/Scripts/NarrativeScript/StoryPlayer.cs
PhoebeMitchell: I specifically moved it to its own method to make it clear this section is all about handling what happens if the story cannot continue. I think it reads easier for the Continue method to just check each thing in turn and return if any of them are true.

unity-ggjj/Assets/Scripts/UI/PenaltyManager.cs
PhoebeMitchell: _penaltyObject.Count doesn't decrement until the object is destroyed. Now PenaltiesLeft decrements instantly. I was running into a bug with using the Count method, though can't remember where exactly.

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/Visibility/ViaKeyboard.cs
PhoebeMitchell: StoryProgresser is just a tool to progress the story. I think I prefer having evidence menu  and cross examination tests in their own namespaces as opposed to having StoryProgresser namespace.

unity-ggjj/Assets/Tests/PlayModeTests/Scripts/AppearingDialogueController/GetDelay.cs
PhoebeMitchell: Changing Awake to Start causes a softlock if AppearingDialougeController decides to call its start method before NarrativeGameState. I think I prefer expecting an error here, as opposed to setting the script excecution order.

unity-ggjj/Assets/Scripts/Interfaces/IEvidenceController.cs
PhoebeMitchell: The action for AddToCourtRecord is ADD_RECORD. I'm hesitant to change it as that will break everyone's scripts. Would changing AddToCourtRecord to AddRecord be a compromise?

PhoebeMitchell: It should be just the ones that remove things which use strings. I personally don't see the point of converting a string to an ActorData/Evidence, and then using that to remove an object, when the original string can just be used instead.

unity-ggjj/Assets/Scripts/Game/Game.cs
PhoebeMitchell: I'm not entirely sure I love it either, but I think I prefer it to events, and every component having 100 references. Unity is a very "everything is available from everywhere" kind of workflow anyway, with being able to assign references in the inspector and find methods, so I'm not sure it makes sense to fight it. That said, if you have more ideas on how to better enforce a clean architecture feel free to express them.

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/StoryPlayer.cs
PhoebeMitchell: There isn't actually way to stop sub-stories. I've removed the section about being stopped.

unity-ggjj/Assets/Scripts/TextDecoder/Preloading/IObjectStorage.cs
PhoebeMitchell: It's needed to allow "as", which is needed because UnityEngine.Object can't be cast to T

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayer.cs
PhoebeMitchell: Scripts now access NarrativeScriptPlayer methods through a single interface. NarrativeScriptPlayerComponent now only contains things required by Unity.

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlaylist/NarrativeScriptPlaylist.cs
PhoebeMitchell: I think at this point the summaries are mostly a visual aid so they're more useful being inside classes.
Further discussion on this here: https://github.com/Studio-Lovelies/GG-JointJustice-Unity/discussions/166

unity-ggjj/Assets/Scripts/SceneLoading/FadeToImageTransition.cs
PhoebeMitchell: StartFade was changed to use a C# Action, so this had to be changed as they are UnityEvent objects, which need to be called using Invoke. As they are UnityEvents and not C# event Actions, "?" just checks if they are null which they should never be.

That said I realised the lambda here is unneeded so that's been removed.

unity-ggjj/Assets/Scripts/BGScenes/BGSceneList.cs
PhoebeMitchell: Now I think about it is this playlist feature still needed? If we have a script loading action that should just be used to load the next script. Were you thinking the script preloading would preload all future scripts? There could be a lot of scripts, and we'd have to account for accidental infinite loops.

unity-ggjj/Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
ViMaSter: So I just tested it this out and uncaught exceptions actually... don't crash the game. That's uhhh... news to me. 👀💦  
**TIL:** As Unity apps are compile to native code and not [CLI apps](https://en.wikipedia.org/wiki/Common_Language_Infrastructure), uncaught exceptions just... vanish.

This was actually my main reason for using exceptions over logs for any state that might happen during development but is unexpected during regular gameplay: I expected the game to simply end at that point. And that's actually not the case with Unity by default. Which I kind I just assumed; that's my bad!  
(Web Player builds also couldn't even "be killed" similarly to how `Application.Quit()` doesn't work there, so we would have needed at least some form of custom error handling for Web Player builds anyway, so I like your idea of adding some form of error handling.)  
Given that: Yes, we **absolutely** need some form of error handling.

I'll create a discussion for it, so we can discuss how we want this to look both from a code and UX perspective.  
(That all said: In my opinion there is a [fairly easy] way to do it, I also already have a rough idea for it, I just need to sketch it out and put up the thread for it.)

---

Keeping that in mind, I would still propose following the mental model that we treat uncaught exceptions as "fatal errors we can't recover from", even if we need to add a bit of custom code for that to apply. The advantage being, that this allows us to cover not only "happy paths" for our logic (by deliberately placing exceptions), making errors a bit more human readable (like the `ScriptParsingException` for example) and also allows us to handle them gracefully in our code (with things like `Assert.Throws(() => { ... })`) and think this `catch`-block can be removed.

The way I see it, if an action doesn't exist in the decoder (even if it might get added later on), it simply isn't implemented (yet), thereby also shouldn't be used in narrative scripts (yet) and the game cannot assume it can continue in a way we would expect it to, if it encounters a line it can't parse. (Imagine a piece evidence not being added, but the story still progressing to a cross examination where it would that piece is required to progress with the trial.)  
Plus, with https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/191 already integrated, it will even fail builds and throw errors in the editor, if a script with not-supported actions is present; I'd like to avoid treating the severity of specific errors differently during compile-time and run-time.

(I really hope this doesn't come across as pedantic. One of the last apps I worked on had no organized way of handling errors which made bugs nearly untraceable, so it's a subject I'm pretty passionate about. I've seen it cause a lot of headache and would avoid it if possible. 👀)

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/Visibility/ViaKeyboard.cs
ViMaSter: I also just noticed, that there's still some Keyboard input in some of the tests I mentioned. Given that, I actually agree with you: These namespaces still have a good reason to exist. Completely unironically, thanks for pointing it out and making me see this from a different perspective! 👍

unity-ggjj/Assets/Scripts/BGScenes/BGSceneList.cs
ViMaSter: I actually think that we need it to make the monoscene idea we had a reality, [I just expected it to "preload" narrative script resources only just before they start and unload them again when the next scene starts](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/discussions/170#:~:text=In%20my%20mind,to%20I.%201.).

unity-ggjj/Assets/Scripts/Interfaces/IEvidenceController.cs
ViMaSter: `AddEvidence` and `AddRecord` definitely work for me!  

I'm not too concerned about the nouns, just about having the method names consistent and use the same syntax and verbs! : D

ViMaSter: Hmm... Like I said, for me it just down to the same level of abstraction: Having to deduct the object from a `string` first so we can pass it to the `EvidenceController` in one method, but not requiring it and automatically doing it for another, feels inconsistent to me.

Using `string` everywhere would be fine for me too, it's just that the conversion needs to be done at some point anyway, so I'd rather have it be done **always outside** or **always inside** the `EvidenceController`.  
I can actually see arguments be made for both:
- **outside** it's keeping the interface domain-specific and using only complex data-types, delegating string parsing to whoever is using this interface
- **inside** it allows callers of methods in this interface to not have to worry how their string gets converted to a piece of evidence; it's the task of the class that is implementing `IEvidenceController` (I think I actually prefer this.)

unity-ggjj/Assets/Scripts/Game/Game.cs
ViMaSter: Haha, I'd love to. I'm just failing to pinpoint it to anything specific.

I'm pretty sure once we actually reach the point of implementing other game states, this class will come up again and at that point we'll also have a better understanding of what we want these classes to be. For now I'll attribute it to me doing premature optimization (which is the root of all evil! 👀) and leave it at that! : D

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/INarrativeScriptPlayer.cs
```diff
@@ -1,12 +1,14 @@
...
+    public GameMode GameMode { get; set; }
+    public bool Waiting { get; set; }
+    public bool CanPressWitness { get; }
+    public bool HasSubStory { get; }
+    
+    public void Continue(bool overridePrintingText = false);
+    public void HandleChoice(int choiceIndex);
+    public void PresentEvidence(ICourtRecordObject courtRecordObject);
+    public void StartSubStory(NarrativeScript narrativeScript);
+    public void SetWaitingToFalseAndContinue();
```
ViMaSter: Nit: _Superfluous `public`-modifier 2: Electric Boogaloo_
```suggestion
    INarrativeScript ActiveNarrativeScript { get; }
    GameMode GameMode { get; set; }
    bool Waiting { get; set; }
    bool CanPressWitness { get; }
    bool HasSubStory { get; }
    
    void Continue(bool overridePrintingText = false);
    void HandleChoice(int choiceIndex);
    void PresentEvidence(ICourtRecordObject courtRecordObject);
    void StartSubStory(NarrativeScript narrativeScript);
    void SetWaitingToFalseAndContinue();
```

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayer.cs
```diff
@@ -48,11 +45,11 @@ public bool CanPressWitness
                 return _subNarrativeScript.CanPressWitness;
             }
 
-            return IsAtChoice && GameMode == GameMode.CrossExamination && !_appearingDialogueController.IsPrintingText;
+            return IsAtChoice && GameMode == GameMode.CrossExamination && !_narrativeGameState.AppearingDialogueController.IsPrintingText && !Waiting;;
```
ViMaSter: Nit: Double semicolon.
```suggestion
            return IsAtChoice && GameMode == GameMode.CrossExamination && !_narrativeGameState.AppearingDialogueController.IsPrintingText && !Waiting;
```

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayerComponentComponent.cs
```diff
@@ -0,0 +1,40 @@
+using UnityEngine;
+
+public class NarrativeScriptPlayerComponentComponent : MonoBehaviour, INarrativeScriptPlayerComponent
```
ViMaSter: `ComponentComponent` and `INarrativeScriptPlayerComponent` seem redundant. Are those files actually required?

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/INarrativeScriptPlayerComponent.cs
```diff
@@ -0,0 +1,4 @@
+public interface INarrativeScriptPlayerComponent
```
ViMaSter, line 1: `ComponentComponent` and `INarrativeScriptPlayerComponent` seem redundant. Are those files actually required?

unity-ggjj/Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuTest.cs
```diff
@@ -27,7 +27,7 @@ public IEnumerator SetUp()
 
             StoryProgresser = new StoryProgresser();
             EvidenceController = Object.FindObjectOfType<EvidenceController>();
-            DialogueController = Object.FindObjectOfType<global::NarrativeScriptPlayerComponent>();
+            DialogueController = Object.FindObjectOfType<global::NarrativeScriptPlayerComponentComponent>();
```
ViMaSter: `DialogueController` is outdated here and should probably be changed to `NarrativeScriptPlayerComponent` for all other occurrences (like comments or variable names) too.

ViMaSter (commented):
Thanks as always for even addressing the minor stuff! Had a look at the already submitted commits and added a few more tiny finding here and there; if you were already planning on fixing them, feel free to just resolve those comment threads!

Also appreciate you taking stances on things that you see differently! These reviews and discussions can get overwhelming quick, so thanks for not getting discouraged and always being verbose and understanding! The discourse is always appreciated and actually made me reconsider a few approaches. 👍 

---

Not directly tied to the latest changes, but the PR in general: The fact that the `Tests.PlayModeTests.Scenes.CrossExamination.ViaKeyboard.CanPresentEvidenceDuringExamination` test actually fails reliably in this PR worries me a bit. Initially I attributed it to GitHub Runner flakiness, but other PRs go through it well and I'm afraid we've actually introduced a bug without noticing it. Unfortunately also can't reproduce it locally though.

If I had to hazard a wild guess, maybe `TestTools.DoUntilStateIsReached()` actually fails silently (on GitHub runners) - literally zero proof for that however. I don't think I've written dedicated test cases for it, so we might be relying on something that's actually broken underneath. 🤔

Should it come down to this "blocking" the PR and we can't reproduce it locally, I'm totally okay with opening an Issue for it (so it's on the radar) and just moving forward with merging it.

unity-ggjj/Assets/Scripts/TextDecoder/DirectorActionDecoder.cs
ViMaSter: Split this up into two discussions actually: [One about the UX](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/discussions/198), the other about [the technical side of things](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/discussions/199).

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/INarrativeScriptPlayer.cs
PhoebeMitchell: One day I'll remember to remove these when copy pasting methods 👍

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayerComponentComponent.cs
PhoebeMitchell: INarrativeScriptPlayerComponent is to allow NarrativeScriptPlayerComponent to be mocked.

unity-ggjj/Assets/Scripts/BGScenes/BGSceneList.cs
PhoebeMitchell: Why is loading all the scripts and then putting them in a playlist required though, if we aren't using anything in those scripts until we reach them? If resources are only loaded when a new script starts (which I agree should be the case), then we can just load the next script when it's needed.

unity-ggjj/Assets/Scripts/Interfaces/IEvidenceController.cs
PhoebeMitchell: I think I prefer the object reference approach, as it allows for easier testing (you don't need to worry about populating the object storage), and it means every object doesn't require a reference to the object storage (even though most already do through NarrativeGameState).

And actually now I look, the Remove methods actually use the object instances themselves so I will change them to not use strings.

unity-ggjj/Assets/Scripts/BGScenes/BGSceneList.cs
ViMaSter: Moved this to https://github.com/Studio-Lovelies/GG-JointJustice-Unity/discussions/170#discussioncomment-2239369. It's not actually something worth blocking the PR for. 👀

unity-ggjj/Assets/Scripts/TextDecoder/Preloading/IObjectStorage.cs
ViMaSter: Makes sense! Given that I think adding it to `GetObjectsOfType<T>()` would actually be good then, as `.Cast<T>()` would simply throw [an `InvalidCastException`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.cast?view=net-6.0#:~:text=source%20is%20null.-,InvalidCastException,-An%20element%20in). With `where T : Object` in place, we'd receive the error already at compile-time, if we try to use primitive types with either method.

ViMaSter (approved):
Seems like [we're stuck with failed tests on GitHub until Unity Support gets back to me](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/issues/200). The only other possible connection I see is the scene loading, but I also haven't verified it yet and don't want to hold this PR up any longer if we have Unity Support taking a look at the report anyway.

Tests succeed reliably when run in the editor and all major points I had either have separate discussion threads now or are fixed, so I think everything looks as good as it can to me at this point. Thanks for addressing and discussing all these things and being patient with my approvals time and time again!

---

Btw., feel free to mark that remaining conversation I opened as resolved once you've read it! Opened up the two threads since I think that topic is worth its own few PRs and with them, there's no need to hold up the PR for it IMO. Just didn't want to close it myself, if you were using that thread as a bookmark or mental note and haven't had a chance to look at it yet!

PR #190, opened 2022-02-13
Feature/actor controller tests

PhoebeMitchell:
## Summary

Adds test for ActorController because i noticed there weren't any and I wanted them.
Also very slightly changes ActorController to allow the currently speaking actor to not be the active actor, allowing for the speaking animations of non active actors to be played.

## Testing instructions

Run the tests. Experience green ticks.

## Additional information

- `[ ]` Changes UI


unity-ggjj/Assets/Scripts/Actor/ActorController.cs
```diff
@@ -214,11 +218,10 @@ public void StartTalking()
             return;
         }
 
-        if (_activeActor.MatchesActorData(_currentSpeakingActor))
+        if (_currentSpeakingActor != null)
```
ViMaSter: Micro nit: All the other `if`s are already inverted, I think it would be nice to keep it consistent, even if it technically makes the method 2 lines longer. That way all conditions this method returns for are grouped and the logic executed if none of these apply follows after that.

```cs
if (_currentSpeakingActor == null)
{
    return
}

_currentSpeakingActor.SetTalking(true);
```

<sup>GitHub doesn't allow me to make it a suggestion, because only line 221 and 223 are changed. :(</sup>

unity-ggjj/Assets/Scripts/Actor/ActorController.cs
```diff
@@ -13,18 +13,18 @@ public class ActorController : MonoBehaviour, IActorController
     [Tooltip("Attach the NameBox here")]
     [SerializeField] private NameBox _nameBox;
```
ViMaSter, line 16: Since this mostly a polish PR, looking over it I've noticed some things I'd like to see changed in some of the affected files that actually aren't _directly_ part of your changes, so I was wondering if you a) agree with them and b) you'd want to address them in this PR already.

- There's `try/catch`-blocks for core functionality, where I'd rather completely remove the `catch` block and have the game actually throw an exception if expected data doesn't exist, than print something and like nothing happened:
    - `ActorController.SetActiveSpeaker()`
    - `ActorController.FindActorDataInInventory()`
    - `ActorController.AssignActorToSlot()`


- `ActorController` has a member called `_activeScene`. That is currently set by `ActorController.OnSceneChanged()` which is called by the `SceneController._onSceneChanged` UnityEvent. Rather than set the active scene via an event, could we add a reference to `[I]SceneController` and access it directly instead? I see two benefits here:
    - `ActorController` currently doesn't do anything on scene change. It only uses the name of the active scene in other methods when they get called, so there's no need to actively inform it about a scene change.
    - (From what I can see) `SceneController._onSceneChanged` is only used to update `ActorController`; if that requirement is gone, since `ActorController` can fetch the value directly (through `ISceneController`), we could remove the event entirely.

That said: Your PR, your call! If you agree with them, but feel this is out of scope, you can just create an issue for them so me (or someone else) can pick them up, next time we make any changes to those files and mark this thread as resolved!

ViMaSter (approved):
More tests! Always like to see that! Even if the coverage isn't increased dramatically, I'm always a proponent of having tests with as few dependencies as possible; so having these controllers covered on their own is a good step in that direction! 👀👌

notexplosive (approved):
Love to see it 💚

unity-ggjj/Assets/Scripts/Actor/ActorController.cs
PhoebeMitchell: Changed the catches to throw exceptions. Personally I think having the custom messages is useful so the try/catches are still there.

I think I'll implemented the direct reference to actor controller in the dialogue refactor PR if thats ok. That PR is more focused on cleaning up how different controllers communicate.

unity-ggjj/Assets/Scripts/Actor/ActorController.cs
PhoebeMitchell: Inverted the if statement, then made all the if statements into one if statement.

unity-ggjj/Assets/Scripts/Actor/ActorController.cs
```diff
@@ -192,7 +193,7 @@ public void SetActiveSpeaker(string actorName, SpeakingType speakingType)
             ActorData actorData = _dialogueController.ActiveNarrativeScript.ObjectStorage.GetObject<ActorData>(actorName);
             _nameBox.SetSpeaker(actorData, speakingType);
             CurrentSpeakingActorData = actorData;
-            _currentSpeakingActor = _actorDataToActor[actorData];
+            _currentSpeakingActor = _actorDataToActor.ContainsKey(actorData) ? _actorDataToActor[actorData] : null;
```
ViMaSter, line 196: If I'm reading this change correct, it sets the `_currentSpeakingActor` to `null` if it's not present in the `_actorDataToActor` Dictionary and quietly continues execution. Can you elaborate on the decision to return `null` here?

I am asking because if we are trying to set the currently active speaker to a specific actor and that actor cannot be found (for any reason, like it not being pre-loaded because it never occured in the `.ink`-file), I would actually expect this to throw an exception instead; if data we need isn't present, we've encountered a state the game cannot recover from. Would love to hear your thoughts on it!

---

Nit portion of the comment: I'm always a fan of adding a test cases covering a fixed bug, to ensure the newly implemented "expected behavior" (to fix whatever was unintended behavior previously and cause the bug) is reliably prevented from reoccuring moving forward, even if this piece of the code is touched again at some point in the future.

That said, it's a) a bug that happened before this got merged to `develop` b) fixing a bug that (I think) can only be present inside a test case directly interacting with this class which shouldn't happen in-game and c) not a user-reported bug, so it's definitely not super important to keep it from reoccuring. If that's reason enough for you not to add it, I totally can understand that! I just think that it (either the current logic or whatever change we might still add in this PR) would make for a great assertion inside a test case!

PhoebeMitchell: This change is because it's possible for an actor to speak even if they don't have a dedicated Actor object in the scene. So if the speaking actor doesn't have an Actor object they don't need to be animated when they are talking. Without the check it attempts to find the speaking actor in the scene and causes an exception.

PR #191, opened 2022-02-14
Feature/176 validate actions inside .ink files

### LINKED ISSUE #176: EditMode test suite OR on asset rebuild hook, which validates all Actions inside `.ink` files
## Description
Currently, if any changes are made to `.ink`-files or the ActionDecoder Actions, scenes have to be manually played through, to ensure all actions still follow the correct syntax.

For reference: When working on https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/154, I removed a parameter from the `SUBSTITUTE_EVIDENCE`-action:  
`&SUBSTITUTE_EVIDENCE:Bent_Coins,Stolen_Dinos_1` -> `&SUBSTITUTE_EVIDENCE:Bent_Coins`

When rebasing the branch onto `develop` after a few other PRs had been integrated, some `.ink` files were relocated. Even though when this branch was originally created all `.ink` files were updated, this resulted in new `.ink` files that were added and not yet updated. As they are newly created and not modified, Git also didn't highlight them as conflicting files during the rebase process.

## Acceptance Criteria

This story will be finished when **any of these two steps** automatically occur:
- (good:) every time the `EditMode` tests are ran, **every** action line inside **all .ink files** is validated against the expected syntax for that action
- (better:) every time AssetBundles are being rebuilt, Unity also verifies, that all `.ink`-scripts contain only valid action lines and pushes errors to the Unity console, if they don't follow the expected syntax ([Unity's Scriptable Build Pipeline](https://docs.unity3d.com/Packages/com.unity.scriptablebuildpipeline@1.7/api/index.html) or [scripted importers](https://docs.unity3d.com/Manual/AssetDatabaseRefreshing.html#:~:text=Scripted%20importers%3A%20Custom%20importers%20that%20extend%20Unity%E2%80%99s%20import%20capabilities.%20These%20are%20written%20in%20C%23.) should allow to listen for rebuilds and attach custom validation steps for specific files)

PhoebeMitchell:
## Summary

An attempt at #176. When an Ink file is compiled after being modified it is validated, and any errors found on actions lines are reported in the console. Also adds a menu item which allows actions in every Ink file in the project to be validated.

Screenshot 2022-02-14 at 09 06 29

## Testing instructions

Open an Ink file with errors (e.g., ActorTestInk.ink), modify it, and check that errors are reported in the Unity console when it is recompiled. Select Assets/Validate Narrative Scripts, and see that the console is filled with errors from all our broken narrative scripts.

## Additional information

- `[ ]` Changes UI
  
closes #176


unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -33,6 +33,12 @@ public class ActionDecoder
     ///         4. Invoking the method with the parsed parameters
     /// </remarks>
```
ViMaSter, line 32: I think this comment doesn't fully apply to this method after this change:  
Steps 1-3 actually live inside `GetMethod()` now, as the invocation is only done when `OnNewActionLine()` is called.

Maybe this comment could be split so the part describing the line-parsing logic is part of `GetMethod()` and the invocation is in `OnNewActionLine()`?

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -33,6 +33,12 @@ public class ActionDecoder
...
     public void OnNewActionLine(string actionLine)
+    {
+        var method = GetMethod(actionLine);
+        method.MethodInfo.Invoke(this, method.ParsedMethodParameters.ToArray());
+    }
+
+    public static Method GetMethod(string actionLine)
```
ViMaSter: From what I can see, with this small refactoring, the naming of these methods could actually be made a bit more descriptive:
- I think something like `InvocationDetails` is more descriptive than `Method`, as it not only contains the method, but parameters to invoke it with too.
- Because of that I would also change `GetMethod()` to `GenerateInvocationDetails()`:  
    I'd use `Generate` instead of `Get` because I try to limit the usage of the verb `get` to logic that simply returns a pre-existing value. If there's any form of computation or calculation involved that exceeds "dereferencing something", I personally prefer terms like `generate` or `calculate`. That way, the name of the method describes the underlying level of complexity.
- `OnNewActionLine()` could also be changed to `InvokeMatchingMethod()`. I only kept `OnNewActionLine` since handling new action lines was all the method did and I also didn't want to come in changing namings while we weren't even sure if this kind of implementation made sense in the first place.  
   While `On...()` explains **why** the method exists, it doesn't actually explain **what** this method really does; which previously was generating and invoking a method based on an action line and now is taking the return value of `GetMethod()`/`GenerateInvocationDetails()` and invoking the method with the parameters contained in it.

As usual, always open to any other ideas on it, if you maybe have some alternatives too!

unity-ggjj/Assets/Scripts/TextDecoder/DialogueController.cs
```diff
@@ -12,7 +12,7 @@ public enum GameMode
 
 public class DialogueController : MonoBehaviour, IDialogueController
 {
-    private const char ACTION_TOKEN = '&';
+    public const char ACTION_TOKEN = '&';
```
ViMaSter, line 15: Nit: I actually think it would make sense to move this to `ActionDecoder`; I see a few benefits with that: 
1. instead of going `actionLine.Substring(1,` inside `ActionDecoder.GetMethod()`, we can actually assert that the first character is the `ACTION_TOKEN`, remove it if it is, and throw a `System.FormatException` if it is not
2. a test-case that the `ActionDecoder` fails with said `System.FormatException` if a line that doesn't start with `&` is passed inside it could be added (because if this ever occurs during normal game execution, the current error message will not be all to descriptive)
3. less references to `*Controller` code we're trying to refactor

Let me know if that makes sense to you too!

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -100,26 +106,23 @@ public void OnNewActionLine(string actionLine)
...
+                throw new ArgumentException($"TextDecoder.Parser for type {methodParameter.ParameterType} has no constructor without parameters");
             }
 
             // Find the 'Parse' method on that parser
             var parseMethod = parser.GetMethod("Parse");
             if (parseMethod == null)
             {
-                Debug.LogError($"TextDecoder.Parser for type {methodParameter.ParameterType} has no 'Parse' method");
-                return;
+                throw new MissingMethodException($"TextDecoder.Parser for type {methodParameter.ParameterType} has no 'Parse' method");
```
ViMaSter, line 131: I... could've sworn this already existed on `develop`. But according to the diff it didn't? 👀💦  
Well, good catch in any case! 👍

unity-ggjj/Assets/Editor/Ink/Validation.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// <summary>
+        /// Called when an Ink file is compiled after editing it
+        /// </summary>
+        private static void OnPostprocessAllAssets (string[] importedAssets, string[] deletedAssets, string[] movedAssets, string[] movedFromAssetPaths) {
+            foreach (var importedAssetPath in importedAssets) {
+                if (InkEditorUtils.IsInkFile(importedAssetPath))
+                {
+                    InkFile inkFile = InkLibrary.GetInkFileWithPath(importedAssetPath);
+                    ValidateNarrativeScript(inkFile);
+                }
```
ViMaSter: Nit: 
image
: D
```suggestion
                if (!InkEditorUtils.IsInkFile(importedAssetPath))
                {
                    continue;
                }
                InkFile inkFile = InkLibrary.GetInkFileWithPath(importedAssetPath);
                ValidateNarrativeScript(inkFile);
```
It doesn't make much of a difference at the moment, but since we're already at 3 levels of indentation without the `if`-conditional, I like limiting any more to simple flow instructions.

unity-ggjj/Assets/Editor/Ink/Validation.cs
```diff
@@ -0,0 +1,57 @@
...
+                {
+                    try
+                    {
+                        ActionDecoder.GetMethod(lines[i]);
+                    }
+                    catch (Exception exception)
+                    {
+                        Debug.LogError($"Error on line {i + 1} of {inkFile}: {exception.Message}");
+                    }
+                }
```
ViMaSter: Nit: Similarly to the comment above, I find that reducing nesting makes this easier to follow. (Especially with the `try/catch`-block adding another level of indentation.
```suggestion
                if (string.IsNullOrEmpty(lines[i]))
                {
                    continue;
                }
                if (lines[i][0] != DialogueController.ACTION_TOKEN)
                {
                    continue;
                }
                try
                {
                    ActionDecoder.GetMethod(lines[i]);
                }
                catch (Exception exception)
                {
                    Debug.LogError($"Error on line {i + 1} of {inkFile}: {exception.Message}");
                }
```
<sup>If you still want to group the `if`-conditional, I'm okay with that too. I personally find multiple simple conditionals easier to read than one complex one, but like in the other PR, if you feel like grouping them works, I'm okay with that too. 👀<sup>

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -142,8 +145,10 @@ public void OnNewActionLine(string actionLine)
             parsedMethodParameters.Add(methodParameters[suppliedParameterCount].DefaultValue);
         }
 
-        // Call the method
-        method.Invoke(this, parsedMethodParameters.ToArray());
+        Method method;
+        method.MethodInfo = methodInfo;
+        method.ParsedMethodParameters = parsedMethodParameters;
+        return method;
```
ViMaSter: Nit: This could be made a bit more concise using an object initializer:
```suggestion
        return new Method {
            MethodInfo = methodInfo,
            ParsedMethodParameters = parsedMethodParameters
        };
```
<sup>I know using initializers and assignments results in different machine code in C++, but I'm actually not sure if the C# compiler actually optimizes it to the same IL code... 🤔 I just find that syntax to be a bit more readable.</sup>

ViMaSter (approved):
It doesn't happen often, that I write tickets and they
<ol type="a">
  <li>get implemented at all</li>
  <li>exactly as I meant them to</li>
  <li>in not even two weeks</li>
</ol>

This is _**so**_ nice.

Got some questions about namings and syntax, but from an architectural perspective this is... 100% what I had in mind. Super well done!

unity-ggjj/Assets/Scripts/TextDecoder/DialogueController.cs
PhoebeMitchell: Waaay ahead of you; this is already done in #189 :D I can change it in this PR too. I assume that won't create some kind of git merge error..?

ViMaSter: > Waaay ahead of you; this is already done in https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/189 :D

Ahh, perfect then! <sup>Should actually start reviewing that one... The 52 changed `.cs` files were always kinda intimidating. 👀</sup>

> I can change it in this PR too. I assume that won't create some kind of git merge error..?

Just checked it, no need to integrate it since it's in that one already!  
And yes, I wouldn't expect there to be any sort of merge conflict. If you were to integrate that change here and it gets to merged to `develop` before #189, the next time you rebase or merge `develop` into `#189` that "change" would simply be removed, since it's identical to the upstream branch (`develop` in that case).

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
PhoebeMitchell: Oddly enough Rider didn't automatically suggest this like it usually does, so assumed it wasn't possible

notexplosive (approved):
Eggcelent 🥚

ViMaSter (approved):

PR #196, opened 2022-02-16
Correct list of valid "SceneAssetNames" inside auto-generated documentation

### LINKED ISSUE #193: [BUG] Doc Generator Generates the wrong "Scenes" constants
## Describe the bug

In the generated docs we list the scenes in `constants.md`, it looks like this.
image

This is.... not right? It should list all possible BGScenes...? Right?

ViMaSter:
## Summary
Resolves #193.

- [before](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/develop/docs/constants.md#sceneassetname)
- [after](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/bug/193-scene-constants/docs/constants.md#sceneassetname)

## Additional information

- `[ ]` Changes UI
  - closes #193

### DIFF
--- a/docs/constants.md
+++ b/docs/constants.md
@@ -117,9 +117,18 @@
   - TutorialBoysTragicallyGenericReminiscence
 
 ## SceneAssetName
-  - BaseScene
-  - MainMenu
-  - Splash
+  - Anime
+  - TMPHAssistant
+  - TMPHCourt
+  - TMPHDefense
+  - TMPHGalaxyBrain
+  - TMPHJudge
+  - TMPHLobby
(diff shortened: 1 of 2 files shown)


PhoebeMitchell (approved):
Thanks for the quick fix 👍

notexplosive (approved):
I tried to fix this myself but then I realized I had no idea where this code actually lived. I didn't think to look in the docstring of the SCENE function. (I was looking at the docgen code and was like "wait... there's nothing here?!")

So maybe some design feedback there? I'm glad it's fixed though! tyty!

Issue #173, opened 2022-02-01
Create a `Demo` Assembly containing every asset (scenes, scripts, textures) only used for manual testing


ViMaSter:
## Description

As code can be created before assets like art, music or narrative scripts are finalized, we are currently creating automated tests and/or scenes which can be run to test new functionality by hand. Some of these include (but aren't limited to) screen shaking, save files, audio playback or UI elements like the evidence (court record) menu.

These files currently live inside the same Assembly that is ran by end-users, even though they are inaccessible via a normal playthrough.

Completing this should result in various developer-targeted UX improvements:
- reduce compilation time 
- reduce binary size of the final game
- referencing placeholder assets inside the game Assembly, will actually prevent the game from being built due to reference errors

## Acceptance Criteria

This story will be finished when every file that is **exclusively** used to demonstrate new functionality and **not required in the final game binary** is moved to a new Assembly named `Demo`.

This Issue includes https://github.com/Studio-Lovelies/GG-JointJustice-Unity/issues/118.

notexplosive:
Emphasizing the need for this: Ana almost put the script he was working on in `TestScripts`

ViMaSter:
Quick update on this: I played around with Assembly Definitions and noticed that they actually only affect C# scripts.  
<sup>(Which should've been obvious to me from the name **Assembly**, but hey, better late than never.)</sup>  
It should be possible to fulfill the acceptance criteria however using [`Addressables`](https://unity.com/how-to/simplify-your-content-management-addressables); the downside being, from what I've seen playing around with them, that will require a fair bit of custom code:

Unity (without Addressable) offers no simple platform-specific Asset management. Ideally, we'd just have this menu from the "Assembly Reference" inspector window, but for any kind of Resource:
<img src="https://user-images.githubusercontent.com/1689033/158480547-d8be333b-7415-4634-90c5-d49551f1527e.png" height=400 />

[...]

PR #206, opened 2022-03-12
Feature/loading scene with narrative script

### LINKED ISSUE #204: Loading scene with narrative script
## Description
When we load the game scene the narrative script needs to be set.
We have agreed to use a find method to do this.

Steps:
- Load the scene additively
- Find the narrative script playlist with a Find method
- Set the narrative script and initialise it
- Unload last the scene
- Start the script
...

## Acceptance Criteria
This story will be finished when a narrative script can be started from another scene.

PhoebeMitchell:
## Summary

Adds functionality to specify a narrative script when loading scenes with the SceneLoader.

## Testing instructions

Run MainMenu. Click the start button. Do you go to the game scene? Does the script start playing? Try changing the script on the start button's scene loader component. Run the tests.

## Additional information

- `[ ]` Changes UI
  
closes #204


notexplosive (approved):

unity-ggjj/Assets/Tests/PlayModeTests/Tools/TestTools.cs
```diff
@@ -57,5 +58,21 @@ public static IEnumerator DoUntilStateIsReached(Func<IEnumerator> action, Func<b
...
             return FindInactiveInScene<T>().SingleOrDefault(obj => obj.name == name);
         }
+
+        /// <summary>
+        /// Sets a private field on an object via reflection
+        /// </summary>
+        /// <param name="targetObject">The object with the field to set</param>
+        /// <param name="fieldName">The name of the field to set</param>
+        /// <param name="value">The value to set the field to</param>
+        public static void SetField(object targetObject, string fieldName, object value)
```
ViMaSter, line 68: Thanks for wrapping this; it's complex code, so no need to repeat it over and over again and the TestTools are the perfect place for it. 👍

Nothing really actionable but as I recently stumbled across it and it feels relevant to me:

Unity actually has the ability to set values of **only editor-exposed private fields/properties** via code. Make a `SerializedObject` out of the object to be modified, call `.FindProperty()` which returns a [`SerializedProperty`](https://docs.unity3d.com/ScriptReference/SerializedProperty.html) and then underlying values can be set with the property matching the type of the value:
```cs
Transform coolTransform = gameObject.transform;
SerializedObject serializedTransform = new SerializedObject(coolTransform);
serializedTransform.FindProperty("position").vector3Value = Vector3.zero;
```

Compared to C# Reflaction, this has the benefit of never exposing a `private` member variable that isn't accessible via the editor. (Limiting test code to things a human could do, seems like a good thing to me.)

However (and the reason this isn't really actionable), until Unity 2022.1 **it isn't possible to access values of `SerializedObject` as `object`** (Unity 2022.1 introduced `boxedValue`) like so:
```cs
Transform coolTransform = gameObject.transform;
SerializedObject serializedTransform = new SerializedObject(coolTransform);
serializedTransform.FindProperty("position").boxedValue = Vector3.zero;
```

Again, nothing actionable, but maybe worth an entry on the not-yet-existing "list of reasons we might want to upgrade Unity in the future".

unity-ggjj/Assets/Tests/PlayModeTests/Tools/InputTestTools.cs
```diff
@@ -114,11 +114,21 @@ public IEnumerator WaitForBehaviourActiveAndEnabled(Behaviour behaviour, ButtonC
...
         }
-
+        
         /// <summary>
         /// Sets the mouse to a position and clicks.
         /// </summary>
         /// <param name="position">The position to click at.</param>
+        public IEnumerator ClickAtPosition(Vector2 position)
+        {
+            yield return SetMousePosition(position);
```
ViMaSter: Nit: I think the comment and parameter name for this method could be a bit more verbose:
```cs
        /// Sets the mouse to specific pixel and clicks.
        /// </summary>
        /// <param name="screenSpacePosition">The position in pixels to click at.</param>
        public IEnumerator ClickAtScreenSpacePosition(Vector2 screenSpacePosition)
        {
            yield return SetMousePosition(screenSpacePosition);
```
(Haven't made this a GitHub Suggestion, as all the references would need to be updated too.)

unity-ggjj/Assets/Tests/PlayModeTests/Tools/InputTestTools.cs
```diff
@@ -114,11 +114,21 @@ public IEnumerator WaitForBehaviourActiveAndEnabled(Behaviour behaviour, ButtonC
...
+            yield return PressForFrame(Mouse.leftButton);
+        }
+
+        /// <summary>
+        /// Sets the mouse to a position in world space and clicks.
+        /// </summary>
+        /// <param name="position">The position to click at.</param>
         public IEnumerator ClickAtPositionWorldSpace(Vector2 position)
         {
             yield return SetMousePositionWorldSpace(position);
```
ViMaSter: Nit: I think with the introduction of `ClickAtPosition` the comment and parameter name for this method could be a bit more verbose:

```cs
        /// <param name="worldSpacePosition">The position to click at in world space coordinates.</param>
        public IEnumerator ClickAtWorldSpacePosition(Vector2 worldSpacePosition)
        {
            yield return SetMousePositionWorldSpace(worldSpacePosition);
```

(Haven't made this a GitHub Suggestion, as all the references would need to be updated too.)

unity-ggjj/Assets/Tests/PlayModeTests/Scripts/SceneLoaderTests.cs
```diff
@@ -0,0 +1,25 @@
...
+using UnityEngine;
+using UnityEngine.SceneManagement;
+using UnityEngine.TestTools;
+
+namespace Tests.PlayModeTests.Scripts
+{
+    public class SceneLoaderTests
+    {
+        [UnityTest]
+        public IEnumerator GameSceneCanBeLoadedWithNarrativeScript()
```
ViMaSter, line 14: More script tests! Always love seeing more things we can reliably test in isolation! 👍

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/MainMenu/ViaMouse.cs
```diff
@@ -99,22 +99,12 @@ public IEnumerator CanEnterAndCloseTwoSubMenusWithCloseAllButton()
...
             
-            RectTransform startGameButton = mainMenu.gameObject.GetComponentsInChildren<RectTransform>().First(menuItem => menuItem.gameObject.name == "NewGameButton");
-
-            yield return _inputTestTools.SetMousePosition(startGameButton.position + startGameButton.localScale * 0.5f);
-            Assert.False(sceneManagerAPIStub.loadedScenes.Contains("Transition - Test Scene"));
-            yield return _inputTestTools.PressForFrame(Mouse.leftButton);
-            Assert.True(sceneManagerAPIStub.loadedScenes.Contains("Transition - Test Scene"));
-
-            SceneManagerAPI.overrideAPI = null;
+            yield return TestTools.WaitForState(() => SceneManager.GetActiveScene().name == "Game");
```
ViMaSter, line 107: It can be so simple... Super nice optimization, well spotted! 👍

unity-ggjj/Assets/Scripts/GameState/NarrativeGameState.cs
```diff
@@ -29,7 +29,20 @@ public class NarrativeGameState : MonoBehaviour, INarrativeGameState
 
     private void Start()
     {
-        _narrativeScriptPlaylist.InitializeNarrativeScripts();
+        // If the narrative script is not null then we are debugging and should start the game automatically
```
ViMaSter: This is a super-duper big nit, to the point where I'm not sure if it makes sense to do it _just_ yet because it would require a lot of boilerplate code we don't have just yet.

If this "debug-only" code-block does grow in the future, I think introducing a "DebugStartGameState" in its stead might be useful:  
It would be a fairly simple class that only takes the narrative script we want to start from and then mirrors whatever behavior the core game loop would do when transitioning from the main menu to the `NarrativeGameState`:

```cs
// Note: This is pseudo-code
class DebugStartGameState
{
    public NarrativeScript NarrativeScriptToLaunch; // replaces `NarrativeScriptPlaylist.NarrativeScript`; settable in the editor
    private Start()
    {
        UnityEngine.LoadScene("NarrativeGameScene");
        Find<NarrativeScriptPlaylist>().NarrativeScript = NarrativeScriptToLaunch;
        Find<NarrativeGameState>().StartGame();
    }
}
```

I see a few benefits here:
1.`NarrativeGameState` would behave identical both in quick-script-launch debug scenes and the core game loop, it's just that we enter it differently (either via the `DebugStartGameState` or selecting a chapter in the main menu)
2. We would relieve both `NarrativeScriptPlaylist` and `NarrativeGameState` of debug-specific code
3. (If https://github.com/Studio-Lovelies/GG-JointJustice-Unity/issues/173 is implemented by the time we do it) `DebugStartGameState` could be made part of the `Demo` Assembly and would prevent the game from building, if we ever do accidentally end up using it inside the core game loop
4. It would force us to think about how we want to implement GameState transitions (both how we pass data to a new state and destroy the state we were in previously and all its associated data)

ViMaSter (approved):
...this actually is the last piece in the "refactoring existing code to load scenes from narrative scripts" puzzle we were shooting for, isn't it?! So happy to see this realized!  
<sup>(Especially since I mostly just talked about it and you did all the work; object preloading and everything! 👀💦)</sup>

...oh and there's refactored code with less nesting.  
...and cleaned up tests.  
It's a moths PR alright.

Really enjoy doing PRs for code like this! 👍

unity-ggjj/Assets/Scripts/GameState/NarrativeGameState.cs
PhoebeMitchell: Split the code into a new script "DebugGameStarter". When this component is assigned a narrative script it will start the game with that narrative script.

Also removed the narrative script field from NarrativeScriptPlaylist, which allowed the removal of the component altogether. It is now a regular C# object called NarrativeScriptStorage which lives on the NarrativeGameState component.

unity-ggjj/Assets/Tests/PlayModeTests/Tools/TestTools.cs
PhoebeMitchell: Interesting. I've seen you use this in tests before. 2022 might not be LTS for a while though. 2021 isn't even LTS yet :p

ViMaSter (approved):

PR #210, opened 2022-03-16
Refactor/monoscene playmode tests

### LINKED ISSUE #118: Move test narrative scripts out of resources folder
## Description
All test narrative scripts should be moved to their own folder, outside of the "Resources" folder. This is because everything in the resources folder is included in the final build of the game and we do not want to include the test scripts.

## Acceptance Criteria
This story will be finished when all the test scripts have been moved to their own folder, outside of the "Resources" folder.

### LINKED ISSUE #185: [BUG] Pressing enter with pause menu open continues story playback
## Describe the bug
Opening the pause menu and pressing enter causes the next line of dialogue to play. If the next line of dialogue is the &PRESENT_EVIDENCE action the pause menu will not close.

...

## Steps To Reproduce

- Open Inky-TestScene
- Run EvidenceAddingTestScript
- Open the pause menu (ESC)
- Press enter to resume game
- Witness how the menu closes and the next line of dialogue plays.
- Open the pause menu again
- Press enter
- The court record menu opens, but the pause menu doesn't close
- The game can only be progressed by using the mouse to press the evidence to present

## Expected behavior
Having the menu open should pause the game. You should not be able to continue while in the pause menu.

## Screenshots
bug

PhoebeMitchell:
## Summary

This PR refactors the tests to use the narrative script loading implemented in #206.
All the play mode tests now all use the game scene.
All test ink scripts have been moved to the PlayModeTests folder.
Unused ink scripts have been deleted (if I've deleted something you want kept feel free to speak up)
Also fixes the bug where pressing enter in the pause menu progresses the story (and another one where the EventSystem would not work on scene load due to two of them being active).
Removes the standard inclusion prefab, now everything uses one scene.

## Testing instructions

Run the tests. Do they still work? They should all use the "Game" scene now.

## Additional information

- `[ ]` Changes UI
  
closes #185 
closes #118


unity-ggjj/Assets/Scripts/SceneLoading/SceneLoader.cs
```diff
@@ -90,6 +93,9 @@ private IEnumerator LoadSceneCoroutine()
         SceneManager.UnloadSceneAsync(SceneManager.GetActiveScene());
     }
 
+    /// <summary>
+    /// Assigns the assigned narrative script to the NarrativeScriptStorage and starts the game
```
ViMaSter: Nit: At first glance "assigning an assigned script" was a bit difficult to understand. Would something like this work?
```suggestion
    /// Passes the narrative script of this SceneLoader instance to the NarrativeScriptStorage and starts the game
```

notexplosive (approved):
1️⃣  M o n o s c e n e 1️⃣

Looks good-- but I also haven't really worked on the end to end tests so I don't have much to say about them.

unity-ggjj/Assets/Scripts/SceneLoading/SceneLoader.cs
```diff
@@ -39,6 +40,11 @@ public void LoadScene(string sceneName)
             throw new Exception("Scene failed to load");
         }
 
+        if (EventSystem.current != null)
+        {
+            EventSystem.current.gameObject.SetActive(false);
+        }
+
```
ViMaSter: Nit: Rather than just disabling the GameObject that contains the current event system, could we instead either `Destroy()` it or (even better) run `UnloadSceneAsync()` with the current scene to completely clean up remaining GameObjects?

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/Visibility/ViaKeyboard.cs
```diff
@@ -3,16 +3,17 @@
...
 using UnityEngine.TestTools;
-using Object = UnityEngine.Object;
 
 namespace Tests.PlayModeTests.Scenes.VisibilityTest
 {
     public class ViaKeyboard
     {
         [UnityTest]
-        [ReloadScene("Assets/Scenes/TestScenes/Visibility - TestScene.unity")]
+        [ReloadScene("Game")]
```
ViMaSter: Nit: With `SceneManager.LoadScene("Game");` being used in other tests now (which I actually like!), could we replace this attribute entirely? (Would be even more code we could remove. 👀)

unity-ggjj/Assets/Tests/PlayModeTests/Scripts/AudioController/AudioControllerTests.cs
```diff
@@ -17,14 +18,14 @@ public class AudioControllerTests
         private const string MUSIC_PATH = "Audio/Music/";
 
         [UnityTest]
-        [ReloadScene("Assets/Scenes/TestScenes/BlankScene.unity")]
         public IEnumerator AudioController_PlaySong_FadesBetweenSongs()
         {
+            yield return EditorSceneManager.LoadSceneAsyncInPlayMode("Assets/Scenes/TestScenes/BlankScene.unity", new LoadSceneParameters());
```
ViMaSter: Nit: Are there any pre-setup GameObjects or Components that are required for these tests to work? I'm wondering, if it would be possible, to loop with `SceneManager.sceneCount` and just unload all scenes instead. (That way we could also delete the blank scene.)

unity-ggjj/Assets/Tests/PlayModeTests/Scripts/EvidenceMenu/EvidenceMenuTest.cs
```diff
@@ -1,5 +1,6 @@
 using System.Collections;
 using System.Linq;
+using NUnit.Framework.Internal;
```
ViMaSter: Nit: I think this namespace should be safe to delete?

unity-ggjj/Assets/Tests/PlayModeTests/Scripts/FullScreenAnimationTests.cs
```diff
@@ -1,5 +1,6 @@
 using System.Collections;
 using NUnit.Framework;
+using NUnit.Framework.Internal;
```
ViMaSter: Nit: I think this namespace should be safe to delete?

unity-ggjj/Assets/Tests/PlayModeTests/Tools/TestTools.cs
```diff
@@ -74,5 +74,19 @@ public static void SetField(object targetObject, string fieldName, object value)
...
+        /// </summary>
+        /// <param name="narrativeScriptName">The path of the narrative script to load</param>
+        public static void StartGame(string narrativeScriptName)
+        {
+            var textAsset = AssetDatabase.LoadAssetAtPath<TextAsset>($"Assets/Tests/PlayModeTests/TestScripts/{narrativeScriptName}.json");
+            Assert.IsNotNull(textAsset);
+            var narrativeGameState = Object.FindObjectOfType<NarrativeGameState>();
+            narrativeGameState.NarrativeScriptStorage.NarrativeScript = new NarrativeScript(textAsset);
+            narrativeGameState.StartGame();
+        }
```
ViMaSter, line 90: Nothing actionable here, just food for thought for future refactorings:

From a high-level perspective  `NarrativeScriptPlayerComponent.LoadScript()` and `TestTools.StartGame()` serve the same purpose; making the game run a specific narrative script. Maybe it's possible to align these two in the future? I could see this being part of the `Game` Assembly and covered by both an explicit test and replacing the need for `TestTools.StartGame()`.

That all aside: Making this a separate method is actually an abstraction that helps keep the tests readable, so it's a good idea to add it. 👍

ViMaSter (approved):
One of my earliest experiences in game dev was one of my colleagues checking how much code he removed from the repository every few weeks. His reasoning being, that keeping the feature set identical (or even adding features) while removing code was always a good thing, as long as people still approved the changes for being readable.

This is... well that. Just in PR form. Suuuuper nice clean-up change. 👍

unity-ggjj/Assets/Scripts/SceneLoading/SceneLoader.cs
PhoebeMitchell: Changed to destroy. This implementation requires both scenes to be active at the same time in order to pass the data from one to the other, so UnloadSceneAsync would not work as far as I know.

unity-ggjj/Assets/Tests/PlayModeTests/Scripts/AudioController/AudioControllerTests.cs
PhoebeMitchell: If you put an invalid path as as the path it appears to create a new empty scene using the path as the scene name, so I've changed it to "TestScene" and removed the blank scene.

unity-ggjj/Assets/Tests/PlayModeTests/Tools/TestTools.cs
PhoebeMitchell: NarrativeScriptPlayerComponent.LoadScript() uses Resources and this script uses AssetDatabase (so test narrative scripts can be outside of Resources and don't need to be included in the final build) so the loading implementation itself would need to be abstracted. I think that would be good, but as you say, something for a future refactor.

ViMaSter (approved):

PR #211, opened 2022-03-18
Split code coverage report into EditMode, PlayMode Scripts and PlayMode Scenes

ViMaSter:
## Summary
This change **does not change game code**.

Right now tests are split into...
- [EditMode tests](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/tree/develop/unity-ggjj/Assets/Tests/EditModeTests)
  - treated as **unit tests** for code that **have a dependency on `UnityEditor`** or **no dependency on `UnityEngine`**
- [PlayMode tests `/Scripts`](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/tree/develop/unity-ggjj/Assets/Tests/PlayModeTests/Scripts)
  - treated as **unit tests** for code that **have no dependency on `UnityEditor`** AND **have a dependency on `UnityEngine`**
- [PlayMode tests `/Scenes`](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/tree/develop/unity-ggjj/Assets/Tests/PlayModeTests/Scenes)
  - roughly treated as **integration tests**

However, our current coverage reports only display the sum of all coverages, not in which types of tests certain parts of code are covered.
### Before
All tests are grouped under the "automated" flag:
image

### After
All tests will continue to have the `automated` flag (to retain historic coverage data) but will additionally be grouped by
- `EditMode`
- `PlayModeScripts`
- `PlayModeScenes`
image
It will also be possible to browse coverage for a file for only a specific (group of) flags on CodeCov.io and lines in the code viewer highlight which lines are covered (green) or not covered (red) based on the selected flags.

https://user-images.githubusercontent.com/1689033/158991289-98891a5f-2c00-4b4c-b28b-674e011d66da.mp4

## Testing instructions
This is [how a report will look like](https://codecov.io/gh/Studio-Lovelies/GG-JointJustice-Unity/tree/5c38ebdda411cd5658def125a498e4ef9d0908af/unity-ggjj).
Here is a link to [the file with the dropdown selector](https://app.codecov.io/gh/Studio-Lovelies/GG-JointJustice-Unity/blob/5c38ebdda411cd5658def125a498e4ef9d0908af/unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs).
The CodeCov reporter will comment on this PR after the builds are finished.

To test this for yourself:
1. Make a PR onto this branch (`feature/split-coverage-reports`)
2. Wait for builds to finish
3. Check out the comment and link insides it created by the CodeCov bot

## Additional information

- `[ ]` Changes UI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -103,6 +103,7 @@ jobs:
           projectPath: unity-ggjj/
           githubToken: ${{ secrets.GITHUB_TOKEN }}
           testMode: EditMode
+          checkName: Test Results - EditMode
           customParameters: -debugCodeOptimization -enableCodeCoverage -coverageResultsPath ./coverage-results -coverageOptions generateAdditionalMetrics;assemblyFilters:+GG-JointJustice
 
       - name: Upload XML report to Codecov
@@ -118,7 +119,7 @@ jobs:
         uses: actions/upload-artifact@v2
         if: ${{ failure() || success() }}
         with:
-          name: Test results
+          name: Test Results - EditMode
           path: artifacts
 
   unityTestPlayModeScripts:
@@ -155,6 +156,7 @@ jobs:
           projectPath: unity-ggjj/
           githubToken: ${{ secrets.GITHUB_TOKEN }}
           testMode: PlayMode
+          checkName: Test Results - PlayMode Scripts
           customParameters: -debugCodeOptimization -enableCodeCoverage -coverageResultsPath ./coverage-results -coverageOptions generateAdditionalMetrics;assemblyFilters:+GG-JointJustice
 
       - name: Upload XML report to Codecov
@@ -170,7 +172,7 @@ jobs:
         uses: actions/upload-artifact@v2
         if: ${{ failure() || success() }}
         with:
-          name: Test results
+          name: Test Results - PlayMode Scripts
           path: artifacts
 
   unityTestPlayModeScenes:
@@ -207,6 +209,7 @@ jobs:
           projectPath: unity-ggjj/
           githubToken: ${{ secrets.GITHUB_TOKEN }}
           testMode: PlayMode
+          checkName: Test Results - PlayMode Scenes
           customParameters: -debugCodeOptimization -enableCodeCoverage -coverageResultsPath ./coverage-results -coverageOptions generateAdditionalMetrics;assemblyFilters:+GG-JointJustice
 
       - name: Upload XML report to Codecov
@@ -222,7 +225,7 @@ jobs:
         uses: actions/upload-artifact@v2
         if: ${{ failure() || success() }}
         with:
-          name: Test results
+          name: Test Results - PlayMode Scenes
           path: artifacts
 
   unityBuild:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,8 +72,8 @@ jobs:
           name: ${{ steps.getManualLicenseFile.outputs.filePath }}
           path: ${{ steps.getManualLicenseFile.outputs.filePath }}
 
-  unityTest:
-    name: Run PlayMode and EditorMode tests for Unity project
+  unityTestEditMode: 
+    name: Run EditMode tests for Unity project
     needs: [checkLicense]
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -90,7 +90,114 @@ jobs:
...
+          name: EditMode
+          flags: EditMode
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: unity-ggjj/coverage-results/**/*.xml
+
+      - name: Store test artifacts
+        uses: actions/upload-artifact@v2
+        if: ${{ failure() || success() }}
+        with:
+          name: Test results
```
PhoebeMitchell: I noticed not all the results are showing. There should be ~390 I think (?)
Screenshot 2022-03-18 at 12 28 03

Two failures are being reported (which I assume is the same "first test fails" problem we already had) but only one test is shown to be failing and the playmode results file only shows the output of the script tests.
Screenshot 2022-03-18 at 12 28 22

Maybe theres a conflict between the names of the output files?

.github/workflows/build.yml
```diff
@@ -99,14 +206,15 @@ jobs:
...
           githubToken: ${{ secrets.GITHUB_TOKEN }}
+          testMode: PlayMode
           customParameters: -debugCodeOptimization -enableCodeCoverage -coverageResultsPath ./coverage-results -coverageOptions generateAdditionalMetrics;assemblyFilters:+GG-JointJustice
 
       - name: Upload XML report to Codecov
         uses: codecov/codecov-action@v2
         if: ${{ failure() || success() }}
         with:
-          name: EditMode+PlayMode
-          flags: automated
```
PhoebeMitchell, line 109: The automated flag still seems to be showing in the codecov report. Would it be possible to remove it?
Screenshot 2022-03-18 at 12 29 17

PhoebeMitchell (commented):
Cool stuff! I'm slowly beginning to get the hang of this yml stuff. Just a couple of things I noticed below.

.github/workflows/build.yml
ViMaSter: Oh, good catch, I think that's exactly it. Pretty sure I'll just have to append some prefixes to these files before we move them to the artifacts, as they'll overwrite each other.

.github/workflows/build.yml
ViMaSter: I would expect this to be gone in future PRs (when both the base commit and branch commit only have `EditMode`, `PlayModeScenes` and `PlayModeScripts` flags).  
The `automated` flag probably comes from the commit this PR is based on (which is `develop`, or rather ff24f3756c96f2b63f02e502fe06b38e0ab27fce) and should make this occur only this cnce.

That said, I'll create a test PR to verify this before I merge it in, just to be safe. Maybe we'll actually have to purge the coverage history for the project for it to be fully gone (which I don't really want to do, unless it's absolutely required).

PhoebeMitchell: It's not a huge deal. I can live with a question mark in the report :D

ViMaSter: Just tried it out [in my fork](https://github.com/ViMaSter/GG-JointJustice-Unity/pull/20#issuecomment-1072728390):
image
As suspected, it's because of the base commit having the `automated` flag; after this PR we're clear!

.github/workflows/build.yml
ViMaSter: Turns out, it's actually not the file names. : D

The `game-ci/unity-test-runner` creates the `Test Results` tab itself (it's not GitHub automatically parsing the XML files) and if it's run multiple times without changing the tab name (the property is called [`checkName`](https://game.ci/docs/github/test-runner#checkname)), it overwrites itself every time.  
Since both tasks, running the tests and creating the tab on GitHub is part of the same action, we can't merge them afterwards, but similarly to the coverage, we can have individual tabs instead:
image

This actually also confirms a suspicion: With `-batchMode`, it's always the first run PlayMode test that fails.  
...because now there's two. 👀💦

PhoebeMitchell (approved):
🛹

notexplosive (approved):

PR #212, opened 2022-03-20
Added WebGL build to GitHub Actions

### LINKED ISSUE #172: Add WebGL to automated build process
## Description

As browsers are also a potential target platform, we should have WebGL builds added to the [automated build pipeline](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/develop/.github/workflows/build.yml).

## Acceptance Criteria

This story will be finished when [runs of game build generations](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/actions/workflows/build.yml) also create a .zip-file for WebGL-versions of our game.

ViMaSter:
## Summary
This change **does not change game code**.

This PR adds **WebGL builds** to the GitHub Actions which are automatically run.

Note: WebGL builds **usually take around 2x** the amount of time of native builds, as Unity first builds a native build to then convert it to WebAssembly. **The time until test results are visible on GitHub does not change**, the "failed workflow" email notification is only sent **once all jobs are completed** however.

image

## Testing instructions
1. Download the `WebGL-*` artifact of [any successful run of this branch](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/actions/runs/2012409460)
2. **IMPORTANT:** Open this folder via HTTP by either:
    1. uploading this folder to your own webhost
    2. use [this temporary version hosted on my personal webspace](https://vimaster.de/dump/Game%20Grumps%20-%20Joint%20Justice/)
3. Give it around a minute to load (this heavily varies based on download and processing speeds)
4. Verify the game runs as expected

## Additional information

- `[ ]` Changes UI
  - closes #172


notexplosive (approved):
pre-approved so it'll be ready when it's time

unity-ggjj/ProjectSettings/ProjectSettings.asset
```diff
@@ -557,7 +557,7 @@ PlayerSettings:
   webGLTemplate: APPLICATION:Default
```
PhoebeMitchell, line 557: Could this be set to minimal?

.github/workflows/build.yml
```diff
@@ -240,7 +240,8 @@ jobs:
           { unityPlatform: "StandaloneOSX",       outputName: "macOS"       },
```
PhoebeMitchell, line 240: Unrelated to this line. Could the dimensions be set to a 16:9 aspect ratio?

PhoebeMitchell (approved):
Looks good. Couple things that would be nice to address below while we're here.

unity-ggjj/ProjectSettings/ProjectSettings.asset
ViMaSter: I actually tried that, but inside the minimal template, the game will not even have a loading bar. Since the game needs around a minute to start up initially, I decided to just go with the default packaging for now, since staring at a white page for 60 seconds feels like nothing is happening (and there's no indication of anything happening aside from the Chrome DevTools. Would you be okay with that or should we just go with the default then? 

We can also customize it and add our own template later down the line and maybe have some of the UI artists do a custom template. It's literally just an HTML file, so if we have a Photoshop/InDesign/toolofyourchoice-file of the layout, we can always spruce it up.

.github/workflows/build.yml
ViMaSter: Whoops, that was actually part of the branch but I think I removed it when I removed the Studio Lovelies website template I was tinkering with. Set it to 1280x720 now!

unity-ggjj/ProjectSettings/ProjectSettings.asset
PhoebeMitchell: Huh, weird. I encountered a problem like this where WebGL games were just a black screen and fixed it by disabling compression. Seems like it's already disabled though. Afiak the loading bar should show even if the minimal template is selected (at least it does for my other projects), and at the risk of being like "it works fine on my machine": it works fine on my machine 😅 
Screenshot 2022-03-21 at 20 07 22
If it doesn't work though, it's fine as it is 👍 Merge it

PR #213, opened 2022-03-21
Feature/FADE_SONG action and PLAY_SONG transition time

PhoebeMitchell:
## Summary

Adds a FADE_SONG action to fade out from a song without fading into a new one.
Adds an optional transitionTime parameter to PLAY_SONG, that allows specifying a transition time from an Ink script. Default time is 0.

## Testing instructions

Run the new tests and/or make a new Ink script containing:
&PLAY_SONG:aBoyAndHisTrial,0
&FADE_SONG:2

This will play a song instantly, and fade it out over 2 seconds.

## Additional information

- `[ ]` Changes UI


unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -150,11 +150,12 @@ protected override void PLAY_SFX(SfxAssetName sfx)
 
     /// <summary>Plays the provided song. Stops the current one. Loops infinitely.</summary>
     /// <param name="songName" validFiles="Assets/Resources/Audio/Music/*.mp3">Filename of a song</param>
+    /// <param name="optional_transitionTime">(Optional) The time taken to transition between songs</param>
     /// <example>&amp;PLAY_SONG:TurnaboutGrumpsters</example>
     /// <category>Audio</category>
-    protected override void PLAY_SONG(SongAssetName songName)
+    protected override void PLAY_SONG(SongAssetName songName, float optional_transitionTime = 0)
```
ViMaSter, line 156: Nit: Could this be changed to `float? optional_transitionTime`? That way, the automatically generated tests for the `ActionDecoder` will generate tests for calling this method with and without the second parameter.

(Just noticed there's an inconsistency where `SET_POSE` has no optional parameter, but `PLAY_EMOTION` does, in case you had used `SET_POSE` as a reference point. Check out the `EditMode` tests for `PLAY_EMOTION` in the editor, to see what this change would result in.)

Can't make this a GitHub suggestion as line 157 hasn't changed and GitHub is being weird about it. :(

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -166,6 +167,18 @@ private void STOP_SONG()
...
         OnActionDone?.Invoke();
     }
+
+    /// <summary>
+    /// Fade out the currently playing song over a given time
+    /// </summary>
+    /// <param name="time">The time taken to fade out</param>
+    /// <example>&amp;FADE_SONG:2</example>
+    /// <category>Audio</category>
+    private void FADE_SONG(float time)
```
ViMaSter: Nit: Could we change this and the new method to be named `FADE_IN_SONG` and `FADE_OUT_SONG` instead? "Fade" could imply both a fade in and out, so I think I'd prefer being a bit more verbose here.

unity-ggjj/Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -19,10 +19,7 @@ public class AudioController : MonoBehaviour, IAudioController
     [Tooltip("SFX Volume level set by player")]
     [Range(0f, 1f)]
     [SerializeField] private float _settingsSfxVolume = 0.5f;
-
-    [Tooltip("Total duration of fade out + fade in")]
-    [Range(0f, 4f)]
-    [SerializeField] private float _transitionDuration = 2f;
```
ViMaSter, line 25: Really like this new and unified interface of music playback methods. 👍

Refactoring things while working on a related feature; "_Always leave the ~~campground~~ codebase cleaner than you found it._" par excellence!

ViMaSter (approved):
Moar features, moar coverage, moar docs. Nicely done. I could get used to this. 👍

unity-ggjj/Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -121,6 +130,15 @@ public void StopSong()
         }
     }
 
+    /// <summary>
+    /// Fade out the currently playing song over a given time
+    /// </summary>
+    /// <param name="time">The time taken to fade out</param>
+    public void FadeSong(float time)
```
notexplosive: The function names are a little awkward here. It looks like you have two functions that want to be named `FadeOutSong` maybe something like:

- `FadeOutSong` should become `FadeOutSongCoroutine`. Helps clarify it's a coroutine and shouldn't be called without a `StartCoroutine` around it.
- `FadeSong` should become `FadeOutSong` to make its name more informative. Maybe `StartFadeOutSong` because it initiates a coroutine?

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
notexplosive: I think `PLAY_SONG` can keep its name because its purpose is more generalized (it plays a song and happens to let you choose how to fade it). But I do agree that `FADE_SONG` feels like it's missing a word. `FADE_OUT_SONG` feels good to me.

notexplosive (approved):
🎸 ⚡

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
PhoebeMitchell: Chose FADE_SONG because it was called FADE_MUSIC in the lua build. However I think I agree so will change it.

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
PhoebeMitchell: Not sure I follow. Adding a ? seems to generate the same tests except they don't work because there currently isn't a ValidData entry for System.nullable.

On a similar topic, Rider seems to be very against me adding a ? for reference types. It automatically changes it to [CanBeNull]. I'm not sure what the difference is, but it doesn't have the warning that ? gives. Are the ?s necessary? Most of the tests under RunCommandWithOptionalParameters don't use them.
Screenshot 2022-03-22 at 02 17 24

Also would be make sense to change "command" to "commands" in RunCommandWithOptionalParameters, as it appears to be running all the commands with optional parameters?

unity-ggjj/Assets/Scripts/AudioController/AudioController.cs
PhoebeMitchell: Changed to FadeOutSong

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
ViMaSter: Welcome to: I forgot how my own code works. 🙃 The optional parameter tests are actually generated by specifying the default value, not `?`. You're absolutely right, sorry for the incorrect suggestion.

`!` and `?` are effectively a shorthand for `[NotNull]` and `[CanBeNull]`, so they're simply annotations to help the compiler prevent us from doing things we don't want to do.

Regarding that annotation: [Before C# 8. reference types (like `ActorAssetName`) can be "null" by default](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references#:~:text=Prior%20to%20C%23%208.0%2C%20all%20reference%20types%20were%20nullable.) and with C# this has changed. Now, if we have code that allows something to be null, that file needs to start with `#nullable` to not throw a warning of us using an annotation that - implicitly - is _technically_ not allowed to be set.

Long story short: Mea culpa, carry on! 👀

---

> Also would be make sense to change "command" to "commands" in RunCommandWithOptionalParameters, as it appears to be running all the commands with optional parameters?

That's... a fun one. I would make the argument for keeping the singular, as that method itself (and that's what the name should describe) only ever tests one command. It's just that it's run multiple times with different parameters. Not married to it though, so feel free to change it!

PhoebeMitchell: I found the naming kinda confusing because I expected it to run all the action with with optional parameters with their optional parameters, but was confused why why it was also running them without the parameters. I expected it to run the same actions without optional parameters in RunValidCommandWithoutOptionalParameters but that is actually just the test for all actions without optional parameters. It makes sense as a method name but not so much as a test name.

That said, I understand it now, and don't care enough to change it atm, so will leave it, but maybe it's something to bare in mind.

PR #214, opened 2022-03-21
Bugfix/ink script validation ignores variables

PhoebeMitchell:
## Summary

Narrative script validation currently just reads files line by line and checks if the actions give an error. This does not take into account Ink variables though. An action such as `&PAN_TO_POSITION:1,{panTime}` would give an error because {panTime} is not a number. This PR changes the method of validation so that Ink scripts are compiled before being read so that variables are taken into account.

## Testing instructions

On develop try making a narrative script like:
```
&PAN_TO_POSITION:1,{panTime}
VAR panTime = 2
```
This should give an error when validating.

Now try the same thing on this PR's branch. Now a success message should appear.

## Additional information

- `[ ]` Changes UI


notexplosive (approved):
I wasn't aware our narrative script format supported variables.... neat!

unity-ggjj/Assets/Editor/Ink/Validation.cs
```diff
@@ -29,33 +31,45 @@ public class Validation : AssetPostprocessor
         [UnityEditor.MenuItem("Assets/Validate Narrative Scripts", false, 202)]
         private static void ValidateAllNarrativeScripts()
         {
-            InkLibrary.GetMasterInkFiles().ToList().ForEach(ValidateNarrativeScript);
+            var noErrors = true;
+            foreach (var unused in InkLibrary.GetMasterInkFiles().ToList().Where(inkFile => !ValidateNarrativeScript(inkFile)))
```
ViMaSter: Nit: Since this variable is unused, it could actually be eliminated by replacing `Where()` with `All()` and assigning it directly to the `noErrors` variable:
```cs
var noErrors = InkLibrary.GetMasterInkFiles().ToList().All(ValidateNarrativeScript));
```

On that subject, I think code duplication could be reduced a bit, by having dedicated methods for finding and reporting found errors and then just calling these from the Asset Processor and Menu Item instead; both of these processes do the same thing, it's just that the list of files to validate varies from manual validation and asset import. 
 
[I could see something like this work](https://gist.github.com/ViMaSter/12adf7470362ecdd3c9891da73f67167), but it might be premature optimization, given this only occurs twice (and [three is my magic number](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) for anything refactoring-specific). Feel free to just use whatever parts of it you think would make the most sense, I'm happy with any (or none) of them! (I just find LINQ to be incredibly nifty. 👀)

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScript.cs
```diff
@@ -58,10 +58,10 @@ private void ReadScript(Story story, IActionDecoder actionDecoder)
     {
         var lines = new List<string>();
         
-        ReadContent(story.mainContentContainer.content, lines);
-        ReadContent(story.mainContentContainer.namedOnlyContent?.Values.ToList(), lines);
+        ReadContent(story.mainContentContainer.content, lines, story);
+        ReadContent(story.mainContentContainer.namedOnlyContent?.Values.ToList(), lines, story);
 
-        var actions = new HashSet<string>(lines.Where(line => line[0] == '&'));
+        var actions = lines.Where(line => line[0] == '&').Distinct();
```
ViMaSter, line 64: Today I learned: https://stackoverflow.com/a/6298720/1088653

Since I tend to write code for people first and compilers second (we're not chasing any benchmark numbers), this is a great change! 👍

ViMaSter (approved):
This kind of validation code is the part of "implement a HTML parser" CS homework I get, understand the need for, but _absolutely_ despise writing myself. Thank you so much for doing it! : D

Like I said before, making use of functionality libraries provide, instead of writing our own code is always a good thing <sup>(shoutout to my credits implementation 👀💦)</sup> and usually just involves a bit of integration work to be done right; well spotted and great implementation. 👍

unity-ggjj/Assets/Editor/Ink/Validation.cs
PhoebeMitchell: Ooh I like. Knew there had to be a more elegant way of doing this, but sleep deprived me didn't care enough x)
Implemented your code.

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScript.cs
PhoebeMitchell: Could have sworn I'd already changed this to Distinct in the PR I initially added this line (funny how that keeps happening 🤔 )
As you say, HashSet works but I figured it's not immediately clear what the intention is.

ViMaSter (approved):

PR #220, opened 2022-03-24
Fix PlayMode tests on GitHub Actions using class derivatives

### LINKED ISSUE #200: [BUG][BLOCKED BY UNITY SUPPORT] First PlayMode test case run fails when using `-batchmode`
**tl;dr:** The first (few) PlayMode test(s) will fail on GitHub until we get a response from Unity Support.

## Describe the bug

Introduced by https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/189/, the first test case that utilizes [`InputTestFixture`](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.0/api/UnityEngine.InputSystem.InputTestFixture.html) ([`Tests.PlayModeTests.Tools.InputTestTools`](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/274d3fce2cef8d03cac3a62ffd9339ae5e9b403a/unity-ggjj/Assets/Tests/PlayModeTests/Tools/InputTestTools.cs) in our project) fails to succeed when run headless using `-batchmode`.

## Steps To Reproduce

- Checkout 4d74b50c29d95ce96699524889358c860d912201 or any commit after it
- Run all `PlayMode` tests
- All tests should succeed
- Close all Unity editor instances
- Run `"[PATH_TO_UNITY_EDITOR]/Unity.exe" -runTests -batchmode -projectPath "[PATH_TO_THIS_REPO]/unity-ggjj" -testResults results.xml -testPlatform PlayMode` in a command line
  - for any non-Windows OS: remove the `.exe`-suffix
  - replace `[PATH_TO_UNITY_EDITOR]` and `[PATH_TO_THIS_REPO]` with both the absolute path to your Unity Editor folder and your local clone of this repository respectively
- Wait for the `Unity` process to exit 
- Check the generated `results.xml` that was generated inside your current working directory
- (At least, but possibly more than) one test should fail. This test/these tests should be (one of the) first test(s) visible inside the `Test Runner` window inside the Unity Editor

## Expected behavior

Running tests via the `Test Runner` window by hand and using `-batchmode` should yield the same test results.

## Additional context

Opened up a ticket with Unity support. Can't post the ticket link though, because that link has both the ID and some random secret in it, which gives full access to all tickets tied to my account. 👍

---

From what I could gather by doing a few changes [on a branch on my fork of our repo](https://github.com/vimaster/GG-JointJustice-Unity/tree/refactor/dialogue-controller-2-fix), I seem to actually have implemented the `InputTestFixture` incorrectly, so this might be related to it:

[Based on the official (and pretty sparse) documentation](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.0/api/UnityEngine.InputSystem.InputTestFixture.html#:~:text=Can%20be%20derived%20from%20or%20simply%20instantiated%20from%20another%20test%20fixture.), there's two ways to use `InputTestFixture`:
- Either [having the class that contains the test cases derive from it](https://github.com/Unity-Technologies/InputSystem/blob/develop/Assets/Tests/InputSystem/Plugins/DeviceSimulatorTests.cs#L15) **OR**
- (this is what we're doing at the moment) instantiating it inside the class containing the tests

Questionable interface design aside, instantiating it in a class [requires us to also call the `Setup()` and `TearDown()`](https://forum.unity.com/threads/input-when-the-game-does-not-have-focus-for-integration-tests.991142/#post-6437402) which we currently are not doing.  
Doing this actually fixes two issues we've encountered in the past:
1. running the tests multiple times resulting in "more than 255 virtual keyboards" breaking all input and requiring Unity to be restarted

2. any sort of mouse or keyboard input disrupting test execution, requiring the Editor to have Window focus and no keyboard or mouse input to be made while tests are run

From my understanding, `Setup()` and `TearDown()` instruct Unity to either switch to or away from a context inside the `InputTestFixture` instance. Doing so, detaches the engine from any physical devices currently plugged in, so tests are actually running in isolation with only the devices present inside the `InputTestFixture`.

That said, I was unable to get mouse tests working that way. I'm sure the way I've set up these tests is probably also not the "recommended approach", resulting in this issue. More importantly, however, it also makes this bug even worse, as the Editor wouldn't just report failed tests but actually freeze and not respond generating no test results at all, as it did a few PRs ago.

ViMaSter:
## Summary
This PR fixes the `PlayMode` tests on GitHub Actions (and when running in `-batchmode`).

**Resolving this issue is possible in two different approaches, which produce the same result with differing code.  
Check out the other implementation (#221) and leave feedback on the approach you would prefer and why.**

Before:
image
:(

After: 
image
:)

The Unity Input System allows automated tests to create and use "fake" devices. This can be achieved in two different ways:
- either (**as in this PR**) each class that contains tests, derives from [`InputTestFixture`](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.0/api/UnityEngine.InputSystem.InputTestFixture.html) directly and creates devices as needed or [a derivative thereof that sets up a predefined collection of devices](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/a83b32fe2695c893231cf21b0ca238357954f61f/unity-ggjj/Assets/Tests/PlayModeTests/Tools/InputTestTools.cs#L14)
- or (**as in #221**) each class that contains tests, creates an instance of [`InputTestFixture`](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.0/api/UnityEngine.InputSystem.InputTestFixture.html) and creates devices as needed or creates [a derivative thereof that sets up a predefined collection of devices](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/a83b32fe2695c893231cf21b0ca238357954f61f/unity-ggjj/Assets/Tests/PlayModeTests/Tools/InputTestTools.cs#L14) but then [**explicitly needs to use `[SetUp]` and `[TearDown]` methods**](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/6b435436bffc6e4a6ffd171c6e4888960667e49e/unity-ggjj/Assets/Tests/PlayModeTests/Scenes/MainMenu/ViaKeyboard.cs#L13-L27)

As mentioned, both approaches achieve the same goal:  
Calling `InputTestFixture.SetUp()` saves all real devices from Unity and disconnects them, so **only** fake devices exist and tests can run in isolation.  
This is required to prevent oddities caused by real input devices:

- OS-specific things like no input events if the Editor isn't focused
- mouse positions being overwritten by the real mouse cursor
- running out of devices (if fake devices are added and the 255 device threshold is crossed)
- removing real devices and the editor being unable to send any input to the game until a device is fully disconnected and reconnected

`InputTestFixture.TearDown()` re-attaches all previously known real devices and allows the user to interact with the game as normal.

**In this implementation, this is done automatically. As all classes containing input tests [derive from InputTestTools (and thereby `InputTestFixture`)](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/a83b32fe2695c893231cf21b0ca238357954f61f/unity-ggjj/Assets/Tests/PlayModeTests/Tools/InputTestTools.cs#L14), `SetUp` and `TearDown` are called automatically by the NUnit runner.**

Additionally, `StoryProgresser` and `InputTestTools` **cannot exist in parallel**:  
As both derive from `InputTestFixture` calling `SetUp` on one, disconnects all devices of the other. Instead:
- `StoryProgresser` now derives from `InputTestTools`
- classes that use **both**, derive from `StoryProgresser`
- classes that **only need `InputTestTools`**, derive from `InputTestTools`

## Testing instructions
Follow the [`Steps to Reproduce`](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/issues/200#:~:text=headless%20using%20%2Dbatchmode.-,Steps%20To%20Reproduce,-Checkout%204d74b50%20or) of Issue #200.

## Additional information

- `[ ]` Changes UI
  - closes #200
  - #221


unity-ggjj/Assets/Tests/PlayModeTests/Scenes/MainMenu/ViaKeyboard.cs
```diff
@@ -9,17 +9,19 @@
...
+        public IEnumerator UnitySetUp()
         {
             yield return SceneManager.LoadSceneAsync("MainMenu");
         }
-        
+        [UnityTearDown]
+        public IEnumerator UnityTearDown()
+        {
+            yield return null;
+        }
```
PhoebeMitchell: What does this do? Seems to do nothing, but I never know with Unity.

unity-ggjj/Assets/Tests/PlayModeTests/Tools/InputTestTools.cs
```diff
@@ -13,8 +13,24 @@ namespace Tests.PlayModeTests.Tools
...
+            InputSystem.RemoveDevice(Mouse);
+            InputSystem.RemoveDevice(Keyboard);
+
+            base.TearDown();
+        }
+
+        public Keyboard Keyboard;
+        public Mouse Mouse;
 
         private EditorWindow _gameViewWindow;
```
PhoebeMitchell: Nit: Move fields to be above methods

unity-ggjj/Assets/Tests/PlayModeTests/Tools/InputTestTools.cs
```diff
@@ -13,8 +13,24 @@ namespace Tests.PlayModeTests.Tools
...
+        public override void TearDown()
+        {
+            InputSystem.RemoveDevice(Mouse);
+            InputSystem.RemoveDevice(Keyboard);
+
+            base.TearDown();
+        }
+
+        public Keyboard Keyboard;
+        public Mouse Mouse;
```
PhoebeMitchell: Nit: Make public fields properties. Also can they be protected?

unity-ggjj/Assets/Tests/PlayModeTests/Tools/InputTestTools.cs
```diff
@@ -13,8 +13,24 @@ namespace Tests.PlayModeTests.Tools
     /// </summary>
     public class InputTestTools : InputTestFixture
```
PhoebeMitchell: nit: Maybe changed InputTestTools to just InputTest? I would say everything that inherits it would be an input test, not an input test tools.

unity-ggjj/Assets/Tests/PlayModeTests/Tools/StoryProgresser.cs
```diff
@@ -3,24 +3,16 @@
...
-        
-        public StoryProgresser()
-        {
-            _appearingDialogueController = Object.FindObjectOfType<AppearingDialogueController>();
-        }
-        
         /// <summary>
         /// Holds the X Key until an AppearingDialogueController is not printing text
         /// </summary>
         public IEnumerator ProgressStory()
```
PhoebeMitchell: Would it make sense to just make this method be part of InputTestTools instead of its own class?

unity-ggjj/Assets/Tests/PlayModeTests/Scripts/ActorControllerTests.cs
```diff
@@ -10,30 +10,43 @@
...
         private global::ActorController _actorController;
         private Animator _witnessAnimator;
 
+        public override void Setup()
+        {
+        }
+
+        public override void TearDown()
+        {
+        }
```
PhoebeMitchell, line 25: Do we not want to call the base Setup and TearDown methods in this test?

unity-ggjj/Assets/Tests/PlayModeTests/Scripts/ActorControllerTests.cs
```diff
@@ -10,30 +10,43 @@
...
+            yield return ProgressStory();
             _witnessAnimator = GameObject.Find("Witness_Actor").GetComponent<Actor>().GetComponent<Animator>();
             AssertIsNotTalking(_witnessAnimator);
         }
 
+        [UnityTearDown]
+        public void UnityTearDown()
+        {
+            base.TearDown();
+        }
```
PhoebeMitchell, line 47: Is calling the TearDown method not done automatically when the test ends?

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/CrossExamination/ViaKeyboard.cs
```diff
@@ -9,31 +9,26 @@
```
PhoebeMitchell, line 8: Seems one of the tests is being flaky again.
Screenshot 2022-03-24 at 21 36 34
Passes when played on its own, but fails when playing all the tests.
The timing is very precise. Maybe it should be changed to use WaitForState, checking if the volume is 0.

PhoebeMitchell, line 8: If tests are cancelled then all input in Unity remains disabled. Do you know if there's a way to re-enable it?

PhoebeMitchell (approved):
Finally, green ticks across the board 👌 😩 
I think I prefer this implementation to the instance implementation. Saves every test class having to call setup and teardown

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/MainMenu/ViaKeyboard.cs
ViMaSter: > What does this do?

Absolutely nothing. I first made a working version of the `instances`-branch, so that's an oversight - fixed it in 1556855eced9412b9437f0be70e15ee4600670d3. Thanks for pointing it out!

unity-ggjj/Assets/Tests/PlayModeTests/Scripts/ActorControllerTests.cs
ViMaSter: Exactly, yes! The reason for that is in line 39: `yield return ProgressStory();`.
For reference, [this is the execution order of events inside the Unity Test Framework.](https://docs.unity3d.com/Packages/com.unity.test-framework@1.1/manual/reference-unitysetup-and-unityteardown.html#:~:text=return%20new%20ExitPlayMode()%3B%0A%20%20%20%20%7D%0A%7D-,Execution%20order,-UnitySetUp%20and%20UnityTearDown)
As you can see, `UnitySetUp` gets called before `SetUp`.

As we already want to use `StoryProgresser.ProgressStory();` inside `UnitySetUp`, we cannot wait until `SetUp()` is called implicitly. Doing so, will cause a `NullReferenceException`, as the `Keyboard` and `Mouse` inside `StoryProgresser` aren't initialized, when the `UnitySetUp` is called.

unity-ggjj/Assets/Tests/PlayModeTests/Scripts/ActorControllerTests.cs
ViMaSter: This relates to your comment above. As we overloaded `TearDown` to do nothing inside this derivative (so that the `SetUp` method is only called once - by hand inside `UnityStartUp` rather than later by the Unity Test Framework), I moved the TearDown into `UnityTearDown` as well, just to keep consistency inside this class.

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/CrossExamination/ViaKeyboard.cs
ViMaSter: I just realized I could actually close this with "Well it doesn't on GitHub so clearly that's only your problem now." now that the tests are working on GitHub again, since that's the whole reason we wanted to have those tests running on GitHub again. : P

Definitely going to look at it though. Pretty sure if it happens to you, it's only a matter of time before it happens on the Runners as well, and since it's a "close to 0 but not really zero" kinda thing, your solution seems reasonable.

ViMaSter: [Official forum post about this](https://forum.unity.com/threads/playmode-test-clean-up-error-or-how-to-disable-real-inputs.886549/#:~:text=Basically%20Input%20System%20mock%20is%20still%20operating%20and%20it%20does%20discover%20devices%20anymore.%20A%20simple%20way%20to%20reset%20it%20could%20do%20the%20trick.%20For%20now%20the%20only%20fix%20is%20to%20relaunch%20the%20editor.).  
Last post on Sep 9, 2020. :(

I can actually reproduce this and tried to see if any of the TearDown methods get called: image
Nothing. Since the runtime where the fake Inputs are set up is stopped (rather than gracefully shut down), I don't think there's any way for code to react and reinstate the real devices when a run is cancelled.

I tried to see if there's a way to add a button to the Editor that finds all or our own custom `InputTestFixtures` and calls `.TearDown()` on them, but they're already destroyed. And playing with some of the internals of `InputSystem` (which is what [`InputTestFixture.TearDown()` interacts with under the hook](https://github.com/Unity-Technologies/InputSystem/blob/develop/Packages/com.unity.inputsystem/Tests/TestFixture/InputTestFixture.cs#L156-L185)) only resulted in more exceptions or no fixes so far.

The quickest way is reloading the Editor probably, since I don't see any "proper" or easy way to reload the InputSystem at the moment.

I think we literally just jumped into the next non-resolvable bug. : }

ViMaSter: Fun fact: This actually works for me right now (both individually and in batches), however, just refactoring the tests actually did not. 👀  


Wild guess: Since `NormalizedVolume -= Time.deltaTime / seconds;` uses `Time.deltaTime` which is `float`, I suppose the value can either over- or undershoot at random times. 

I've moved from a "remove (unprecise) `deltaTime` from volume" to "set to volume based on (precise) interpolated time" approach and that seems to have did the trick.
Let me know if [the new test syntax](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/220/commits/d29edebe8af888e701a6ac3ffea6975e1b758ad8#diff-c183e3c921da802a1d9e842f539bb009528abff43ee4c1e52c5b8cc629c7ca13L91-R99) and [new fader logic](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/220/commits/d29edebe8af888e701a6ac3ffea6975e1b758ad8#diff-737d90868b744ec23694e0cd2e3f1871b88f3f355302816d330278d386e88cb5L52-L59) both work on your machine and were what you had in mind!

unity-ggjj/Assets/Tests/PlayModeTests/Scripts/ActorControllerTests.cs
PhoebeMitchell: Ooh yeah, I encountered this problem before. Thanks for the explanation.
Not a huge fan of the way it's fixed here but I've got nothing better to suggest. Seems like a huge oversight on Unity's behalf tbh.

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/CrossExamination/ViaKeyboard.cs
PhoebeMitchell: Now AudioController_PlaySong_FadesBetweenSongs appears to be broken 😅 
Screenshot 2022-03-25 at 13 08 00
Edit: just seen your discord message haha

PhoebeMitchell: Well... at least you got it working without disabling the fast enter playmode settings (which I couldn't do 😭 )
~NotExplosive MonoGame engine when?~

ViMaSter: So here's an idea: Rather than testing `FadeToNewSong` which directly interacts with `AudioSource`, could we just cover `MusicFader` instead? I see two upsides here:
 - This could be done in `EditMode` (faster execution)
 - We'd reduce flakiness: Rather than waiting for X seconds and _hoping_ it was enough time for Unity to update all MonoBehaviors, we can set `Time.time` by hand (at least I remember that being a thing), wait for exactly one frame and then verify that `NormalizedVolume` is updated accordingly.

The obvious downside being, that `AudioController` itself loses coverage. Curious to hear everyone's thoughts on it.

ViMaSter: Fixed the current tests and things appear to be stable? Let me know if the tests work for you locally now too!

<sup>And apparently I need to get a MacBook since that always matches what happens on the GitHub Runners...</sup>

PhoebeMitchell: Tests pass on my machine 👍
Regarding testing MusicFader, if all the volume changing logic happens there and audiocontroller just handles the time part, that should be good.
Though I do like that the actions themselves are tested with the play mode tests.

ViMaSter: Whoo! \o/
And a fair point, absolutely. At the end of the day, more coverage is more coverage and I was looking for good excuses not to deal with Unitys flakiness when it comes to runtime tests. Things seem to work now though, so thank you for not giving in and making me properly fix the tests. : D

ViMaSter: Anything I should look into here before marking this as resolved or should we just create a new Issue for it to keep track of it, I'll bug Unity Support again and we just have to keep in mind not to stop test runs halfway through for the time being?

(Just to make sure I don't close an open thread again... : D)

PhoebeMitchell: Now that I think about it, I had to disable the fast enter playmode settings because of this exact problem. The input gets reset on a domain reload, which currently isn't done in our project.
ProjectSettings > Editor
Screenshot 2022-03-26 at 14 45 14
I think the domain is also reloaded after editing a script, which is a bit of an unintuitive workaround, but seems to fix the problem for me. If there was a way to force a domain reload that would be good, but haven't found one yet.
Maybe disable the fast enter playmode settings so it doesn't cause problems for anybody, and I can reenable them locally when I working on stuff and disable them for PRs, then this can be resolved.

ViMaSter:
Closed in favor of https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/221. (AudioController test changes of this PR have been integrated into that PR before merging.)

PR #221, opened 2022-03-24
Fix PlayMode tests on GitHub Actions using class instances

### LINKED ISSUE #200: [BUG][BLOCKED BY UNITY SUPPORT] First PlayMode test case run fails when using `-batchmode`

ViMaSter:
## Summary
This PR fixes the `PlayMode` tests on GitHub Actions (and when running in `-batchmode`).

**Resolving this issue is possible in two different approaches, which produce the same result with differing code.  
Check out the other implementation (#220) and leave feedback on the approach you would prefer and why.**

Before:
image
:(

After: 
image
:)

The Unity Input System allows automated tests to create and use "fake" devices. This can be achieved in two different ways:
- either, **as in #220**, each class that contains tests, derives from [`InputTestFixture`](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.0/api/UnityEngine.InputSystem.InputTestFixture.html) directly and creates devices as needed or [a derivative thereof that sets up a predefined collection of devices](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/b3e6b55b31d0cfa414c1cc45cdf6816d1f75798e/unity-ggjj/Assets/Tests/PlayModeTests/Tools/InputTestTools.cs#L14-L22)
- or **as in this PR**, each class that contains tests, creates an instance of [`InputTestFixture`](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.0/api/UnityEngine.InputSystem.InputTestFixture.html) and creates devices as needed or creates [a derivative thereof that sets up a predefined collection of devices](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/b3e6b55b31d0cfa414c1cc45cdf6816d1f75798e/unity-ggjj/Assets/Tests/PlayModeTests/Tools/InputTestTools.cs#L14-L22) but then [**explicitly needs to use `[SetUp]` and `[TearDown]` methods**](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/6b435436bffc6e4a6ffd171c6e4888960667e49e/unity-ggjj/Assets/Tests/PlayModeTests/Scenes/MainMenu/ViaKeyboard.cs#L13-L27)

As mentioned, both approaches achieve the same goal:  
Calling `InputTestFixture.SetUp()` saves all real devices from Unity and disconnects them, so **only** fake devices exist and tests can run in isolation.  
This is required to prevent oddities caused by real input devices:

- OS-specific things like no input events if the Editor isn't focused
- mouse positions being overwritten by the real mouse cursor
- running out of devices (if fake devices are added and the 255 device threshold is crossed)
- removing real devices and the editor being unable to send any input to the game until a device is fully disconnected and reconnected

`InputTestFixture.TearDown()` re-attaches all previously known real devices and allows the user to interact with the game as normal.

**In this implementation, this is done manually. Each classes containing input tests [explicitly creates `[SetUp]` and `[TearDown]` methods where the respective methods on InputTestTools (and thereby `InputTestFixture`) are called](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/6b435436bffc6e4a6ffd171c6e4888960667e49e/unity-ggjj/Assets/Tests/PlayModeTests/Scenes/MainMenu/ViaKeyboard.cs#L13-L27).**

Additionally, `StoryProgresser` and `InputTestTools` **cannot exist in parallel**:  
As both derive from `InputTestFixture` calling `SetUp()` on one, disconnects all devices of the other. Instead:
- `StoryProgresser` now derives from `InputTestTools`
- classes that use **both**, only have a `StoryProgresser` instance
- classes that **only need `InputTestTools`**, only have an `InputTestTools` instance

## Testing instructions
Follow the [`Steps to Reproduce`](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/issues/200#:~:text=headless%20using%20%2Dbatchmode.-,Steps%20To%20Reproduce,-Checkout%204d74b50%20or) of Issue #200.

## Additional information

- `[ ]` Changes UI
  - closes #200
  - #220

### DIFF
--- a/unity-ggjj/Assets/Scripts/AudioController/MusicFader.cs
+++ b/unity-ggjj/Assets/Scripts/AudioController/MusicFader.cs
@@ -41,6 +41,7 @@ public class MusicFader
     /// <returns>WaitUntil object that returns control to the coroutine when it's finished</returns>
     public WaitUntil FadeOut(float seconds = 1f)
     {
+        var startTime = Time.time;
         return new WaitUntil(() =>
         {
             if (seconds <= 0)
@@ -49,17 +50,17 @@ public class MusicFader
                 return true;
             }
 
-            NormalizedVolume -= Time.deltaTime / seconds;
+            NormalizedVolume = Mathf.Lerp(1f, 0f, (Time.time - startTime) / seconds);
 
-            if (NormalizedVolume <= 0)
-            {
-                NormalizedVolume = 0;
-                return true;
-            }
-            else
+            // not enough time has passed
+            if (Time.time - startTime <= seconds)
             {
                 return false;
             }
+
+            NormalizedVolume = 0;
+            return true;
+
         });
     }
 }
--- a/unity-ggjj/Assets/Tests/PlayModeTests/Scenes/CrossExamination/ViaKeyboard.cs
+++ b/unity-ggjj/Assets/Tests/PlayModeTests/Scenes/CrossExamination/ViaKeyboard.cs
@@ -2,7 +2,6 @@
 using NUnit.Framework;
 using Tests.PlayModeTests.Tools;
 using UnityEngine;
-using UnityEngine.InputSystem;
 using UnityEngine.SceneManagement;
 using UnityEngine.TestTools;
 using Object = UnityEngine.Object;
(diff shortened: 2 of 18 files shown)


PhoebeMitchell (dismissed):

PhoebeMitchell (commented):
Wait I approved by accident how do I undo

notexplosive (approved):

lyndsiWilliams (approved):

PR #232, opened 2022-03-27
bugfix/223-First-evidence-item-is-immediately-selected-when-presenting-evidence

### LINKED ISSUE #223: [BUG] First evidence item is immediately selected when presenting evidence
## Describe the bug
When the PRESENT_EVIDENCE action is used the evidence menu opens and the first item is immediately selected. This does not close the menu, causing a soft lock.

## Steps To Reproduce
Create and run an Ink script containing the following:
```
test script

&ADD_EVIDENCE:AttorneysBadge
&SCENE:TMPHLobby
&ADD_FAILURE_SCRIPT:TMPH_FAIL_1

-> Present("BentCoins") ->

end of test

=== Present(presentedObject)
    &PRESENT_EVIDENCE
    + [Wrong]
        -> Present(presentedObject)
    + [{presentedObject}]
        &MODE:Dialogue
        ->->
```

Run this script and witness how when the evidence menu opens a failure state script starts playing straight away.
Change BentCoins to AttorneysBadge and "end of test" will show, but the evidence menu will still be open.

## Expected behavior
Evidence should not be immediately selected. Evidence menu should close when evidence is selected.

PhoebeMitchell:
## Summary

Adds a one frame delay when the evidence menu is open before buttons can be selected.
This fixes buttons being selected on the same frame the menu is opened.

## Testing instructions

Open AddRecordTest.ink
Remove the line `&REMOVE_EVIDENCE:Bent_Coins`
Play the script in the Game scene.
When prompted to present evidence you should be able to present the evidence by pressing X or Enter.
The text box should say correct after you do this and NOT before.

## Additional information

- `[ ]` Changes UI
  
closes #223

### DIFF
--- a/unity-ggjj/Assets/Scripts/Evidence/EvidenceMenuItem.cs
+++ b/unity-ggjj/Assets/Scripts/Evidence/EvidenceMenuItem.cs
@@ -1,3 +1,4 @@
+using System.Collections;
 using UnityEngine;
 using UnityEngine.EventSystems;
 using UnityEngine.UI;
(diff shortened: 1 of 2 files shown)


unity-ggjj/Assets/Scripts/Evidence/EvidenceMenuItem.cs
```diff
@@ -10,7 +11,38 @@ public class EvidenceMenuItem : MonoBehaviour, ISelectHandler
...
+    private Selectable Selectable
+    {
+        get
+        {
+            if (_selectable == null)
+            {
+                _selectable = GetComponent<Selectable>();
+            }
+
+            return _selectable;
```
lyndsiWilliams:
```suggestion
            _selectable == null ? _selectable = GetComponent<Selectable>() : _selectable;
```
Would this work as a ternary?

lyndsiWilliams (approved):
Just a nit, otherwise this looks great and works great locally! All tests pass, even that pesky mouse test 🐭

ViMaSter (approved):
_Let's add "offset being enabled and being interactive" to the list of "things I would've expected the UI system to already take care of implicitly"..._ (Does that count as a head-desk moment? 🤔)

Simple workaround and comment describing the need for it, which is always appreciated for things that "look weird but are required". 👍

PR #238, opened 2022-03-29
Change `Evidence` to `EvidenceData` to be in line with `Actor`/`ActorData`

ViMaSter:
**This is a proactive proposal PR, so I'm happy to close it if we don't agree on it!**

Opening it up to discussion or creating an Issue before seeing the code changes for it didn't really feel worth it, given the small size of the change.

## Summary

Writing [the word list for this project](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/wiki/Word-List), I realized [our definitions of Actors and Evidence are not aligned](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/wiki/Word-List#:~:text=Actor,Evidence%20object%20in%20Unity.%22), but realized, they might as well:
- `Actor`: Thing we instantiate in the game
- `ActorData`: ScriptableObject of data for Actors

...which would dictate that...

- `Evidence`: Thing we instantiate in the game
- `EvidenceData`: ScriptableObject of data for Evidence

However:
- we don't instantiate Evidence, only UI renderers for it (so there's technically no need for an `Evidence` class)
- what we should call `EvidenceData` is actually called `Evidence` right now

This PR changes `Evidence` to `EvidenceData`, unless we're referring to "a piece of evidence" from a gameplay-perspetive inside the comments.

## Testing instructions
1. Check if there are any missed instances of `Evidence` that should be changed and aren't part of this PR
2. Run the game and make sure scenes run as expected (I used the `Game` scene with the `RossCoolX` story.)

## Additional information

- `[ ]` Changes UI


notexplosive (approved):
For refactor PRs like this there's a rather large ratio of "obvious thing you probably used automated tooling to do 100 times" and "meaningful change done by a human." 

What can help this is if you do the automated steps as singular commits and clearly mark them. 

For instance if I saw a commit message like: `Rename Evidence to EvidenceData via automated refactoring` (or something like that, doesn't need to be that verbose)

I can just skim that commit and say "yep, I trust Rider/VS did the right thing" then read the rest of the changes commit-by-commit because they're likely just tiny cleanup things that came after (which appears to be the case here).

unity-ggjj/Assets/Resources/Evidence/AttorneysBadge1.asset
```diff
@@ -9,7 +9,7 @@ MonoBehaviour:
   m_GameObject: {fileID: 0}
```
PhoebeMitchell: I didn't realise we still had these test evidences. They should be safe to delete at this point.

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -245,17 +245,17 @@ protected override void SCENE(SceneAssetName sceneName)
     }
     
     /// <summary>Shows the given evidence on the screen in the given position.</summary>
-    /// <param name="item" validFiles="Assets/Resources/Evidence/*.asset">Name of item to show</param>
+    /// <param name="evidenceName" validFiles="Assets/Resources/Evidence/*.asset">Name of the piece of evidence to show</param>
     /// <param name="itemPos">`Left`, `Right` or `Middle`</param>
     /// <example>&amp;SHOW_ITEM:Switch,Left</example>
     /// <category>Scene</category>
-    protected override void SHOW_ITEM(EvidenceAssetName item, ItemDisplayPosition itemPos)
+    protected override void SHOW_ITEM(EvidenceAssetName evidenceName, ItemDisplayPosition itemPos)
```
PhoebeMitchell: Technically actors can also be shown... but then again I wouldn't classify an actor as an item either 🤔

PhoebeMitchell (approved):
Looks good, but seems something broke in the docs

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
ViMaSter: > Technically actors can also be shown... but then again I wouldn't classify an actor as an item either 🤔

That's actually a good point! I just looked at the parameter type and updated the parameter named based on it, but looking into the methods it calls, there's actually already support for actors (except the documentation also only just lists Evidence, as it uses `EvidenceAssetName`).

How would you feel about changing `EvidenceAssetName` and `ActorAssetName` to derive from a new subclass called `CourtRecordItemName`, use that subclass here and change the parameter name to `courtRecordItemName`?

PhoebeMitchell: Sounds good. In the code ActorData and Evidence implement ICourtRecordObject so maybe that could maybe be updated to ICourtRecordItem to be consistent?

How would this work with the  "validFiles" option? Can it point to multiple folders?

ViMaSter: > Sounds good. In the code ActorData and Evidence implement ICourtRecordObject so maybe that could maybe be updated to ICourtRecordItem to be consistent?

Nice! Updating `ICourtRecordObject` to `ICourtRecordItem` sounds like a great idea actually! Gonna make that part of the PR too.

> How would this work with the "validFiles" option? Can it point to multiple folders?

- Right now:
  - No support for anything like that.  
- As a first pass in this PR:
  - I'll change the docs generator, so it recursively searches subdirectories too, instead of limiting itself to the exact folder and filename mask. That way if we specify `Assets/Resources/*.asset` for `CourtRecordItemName`, it can replace `*` with `Actors/Arin` or `Evidence/Switch` (as Actors and Evidence are only `*.assets` we have in `Resources` at the moment).  
    [Under the hood](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/develop/docs_generator/XMLDocParser.cs#L118) it uses [`DirectoryInfo.EnumerateFiles()`](https://docs.microsoft.com/en-us/dotnet/api/system.io.directoryinfo.enumeratefiles?redirectedfrom=MSDN&view=net-6.0#overloads), so it's just a matter of adding one parameter to make it recursive.
- If we do come across a need for multiple folders:
  - We can always extend the generator, we'll just have to agree on a syntax that feels the most intuitive first. Felt that might be scope creep for this PR though, so I'd go for the recursive search first, unless you feel like there's an easier way to handle it for now.

---

Oh, also fixed the issue with the docs generator. Good catch!

PhoebeMitchell: Yes, do the simpler one for now 👍

ViMaSter: > How would this work with the "validFiles" option? Can it point to multiple folders?

Hey, remember my great workaround solution? Guess what? If I just always search recursively, incorrect data shows up for the `NarrativeScriptAssetName`s, as [the `Failures` and `Templates` folder](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/tree/448dab498b1b71a33d1f9487ce930c44406f7bb2/unity-ggjj/Assets/Resources/InkDialogueScripts) would be crawled too:  
image

Borrowed a page from `.gitignore` files and resolved this by supporting `**` now:
`Assets/Resources/*.asset` (single `*`) -> only files in `Assets/Resources` ending with `.asset`
`Assets/Resources/**.asset` (double `*`) -> any file ending with `.asset` in `Assets/Resources` or subdirectories thereof

Based [on the diff](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/238/commits/3ed810849edbb34278582f87d5f3f14a8f6a9643) this seems to have worked. 👀  
Feel free to mark this as resolved, if it all looks good to you too!

PhoebeMitchell: Ooh, and that would allow narrative scripts to be stored in separate files too I think?

Only problem is they're merged into one category in the docs and I love how neat and organised it currently is :D, but I think having the information there and it looking not so good is better than having missing information. This is good 👍

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -245,13 +245,13 @@ protected override void SCENE(SceneAssetName sceneName)
     }
     
     /// <summary>Shows the given evidence on the screen in the given position.</summary>
-    /// <param name="evidenceName" validFiles="Assets/Resources/Evidence/*.asset">Name of the piece of evidence to show</param>
+    /// <param name="itemName" validFiles="Assets/Resources/**.asset">Name of the piece of evidence to show</param>
```
PhoebeMitchell: This description could be updated now this is for all items

PhoebeMitchell (approved):

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
ViMaSter: Ah, good catch! Updated it to
```xml
<summary>Shows a piece of evidence or actor on the screen in the given position.</summary>
<param name="itemName" validFiles="Assets/Resources/**.asset">Name of a piece of evidence or actor to show</param>
```

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
ViMaSter: > Ooh, and that would allow narrative scripts to be stored in separate files too I think?

Technically yes. However, the current solution requires **filenames to be unique across directories**:  
We're effectively merging `/Actors` and `/Evidence` and referring to items by only their filename (`Arin` or `BentCoins`).  
If we had an `Arin` Actor and `Arin` Evidence, one would override the other. 

The alternative would require us to always specify a sub-directory (`Actor/Arin` or `Evidence/BentCoins`), but then your outlined scenario of having a `Story` folder, pointing the game at it, and being able to reference all files both in the root and any subdirectories (`Case1/Intro`, `Case1/CrossExamination/RossCoolX`) would become true.

> Only problem is they're merged into one category in the docs and I love how neat and organised it currently is :D, but I think having the information there and it looking not so good is better than having missing information. This is good 👍

If we were to implement that second scenario, we could group them by subfolder inside the docs:

---

## NarrativeScriptAssetName

### Case1
- Case1/Intro
### Case1/CrossExamination
- Case1/CrossExamination/RossCoolX
### Case2
- Case2/AboutAlmonds
- Case2/SeriousBeezness
- Case2/ToBeContinued

ViMaSter:
@notexplosive  I completely missed your review on my last PR cause of GMail grouping topics and every link immediately scrolling down to the threads from moths. My bad! :eyes:

> For refactor PRs like this there's a rather large ratio of "obvious thing you probably used automated tooling to do 100 times" and "meaningful change done by a human."
> 
> What can help this is if you do the automated steps as singular commits and clearly mark them.

I can definitely try that! As I mentioned before, I'm used to just used to everyone just looking at the total changes, rather than the individual steps. The reasoning for the changes is usually easier to find in the PR description, rather than deductible from the commit history and (putting myself in your shoes) if there was a discussion about changes before I got my initial review in, I just view those conversations and then move to the current diff. I see where you're coming from though; being used to `aaa`, `fix bug`, `refactor` commit messages, that just never crossed my mind.
 
I'll use that as an excuse to try [semantic commits](https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716#semantic-commit-messages) which I always wanted to try out and seem to overlap with your idea of "flagging" certain types of commits, so thanks for the suggestion!  

As an example, would something like an `auto:`-prefix for those "tool-assisted commits" work for you?
- current commit message: `Changed Evidence to EvidenceData to be in line with Actor/ActorData`
- using semantic commits: `auto(Evidence): renamed Evidence to EvidenceData`

> For instance if I saw a commit message like: `Rename Evidence to EvidenceData via automated refactoring` (or something like that, doesn't need to be that verbose)
> 
> I can just skim that commit and say "yep, I trust Rider/VS did the right thing" then read the rest of the changes commit-by-commit because they're likely just tiny cleanup things that came after (which appears to be the case here).

Funny story: In this exact instance, ReSharper actually goofed up:
image

After just running a rename-operation on `Evidence`, all comments ended up as 
```xml
/// <param name="evidenceDatae evidence to add.</param>
```
for some reason. In an instance like that, would you prefer it to just be committed-as-is (at risk of either the build or things like the docs generator breaking, but having split automated from manual change), or should there at least be an attempt of getting things to a "valid" state, before creating a commit (but still flag that as an automated commit, even though there's a few manual tweaks)?

notexplosive:
@ViMaSter (because github won't let me reply directly)
`auto:` prefix works great! Exactly what I was talking about.

> for some reason. In an instance like that, would you prefer it to just be committed-as-is (at risk of either the build or things like the docs generator breaking, but having split automated from manual change), or should there at least be an attempt of getting things to a "valid" state, before creating a commit (but still flag that as an automated commit, even though there's a few manual tweaks)?

I think the automated commit happens alone, then a tiny commit for your follow-up changes. But I think this only really works for tightly scoped refactors like this where you could feasibly read it commit-by-commit and make sense of it.

PR #241, opened 2022-03-30
Feature/Case 1 stories and tests

### LINKED ISSUE #209: Tutorial Narrative Scripts
## Description
We have narrative scripts for the tutorial case in [here](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/tree/feature/turnabout-1-narrative-scripts)

They need to be updated:
- Dialogue and cross examination scripts should be merged now we have the &MODE action to swap between dialogue and cross examination mode
- They need to be functional i.e. you can play through them with no errors
- Any commented actions need to be uncommented as they should have been implemented by now
- The LOAD_SCRIPT action needs to be added to the end of each script to load the next script
- Any underscores or spaces need to be removed from the script names

## Acceptance Criteria
This story will be finished when the narrative scripts are updated and functional.

PhoebeMitchell:
## Summary

Adds the stories for case 1.
Fixes a few minor bugs so stories can be played.
Adds a test that automatically plays through each story. To ensure the tests do not take forever the test script sets AppearingDialogueController's AppearInstantly property to true. To view the tests as they would be played in game comment out line 55 of the `Case1Tests` script.

## Testing instructions

Play the stories manually (They are in Resources>InkDialogueScripts/Case1)
Run the tests. Check if they pass.

## Additional information

- `[ ]` Changes UI
  
closes #209


unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayerComponent.cs
```diff
@@ -45,16 +45,20 @@ public void TryPressWitness()
     }
 
     /// <summary>
-    /// Loads a narrative script, ending the current narrative script
-    /// and continuing the beginning of the loaded script
+    /// Loads a narrative script using a given narrative script name,
+    /// ending the current narrative script and
+    /// continuing the beginning of the loaded script
```
ViMaSter: Nit: "Continuing the beginning" reads a bit weird (since it's the start, it can't really continue), though I understand what the comment is trying to communicate. Would this work as an alternative?
```suggestion
    /// starting the newly loaded script
```

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayerComponent.cs
```diff
@@ -45,16 +45,20 @@ public void TryPressWitness()
...
     /// </summary>
     /// <param name="narrativeScriptName">The name of the narrative script to load</param>
     public void LoadScript(string narrativeScriptName)
     {
         var narrativeScriptText = Resources.Load<TextAsset>($"InkDialogueScripts/{narrativeScriptName}");
-        NarrativeScriptPlayer.ActiveNarrativeScript = new NarrativeScript(narrativeScriptText);
+        Debug.Assert(narrativeScriptName != null, $"Failed to load narrative script with name {narrativeScriptName}");
```
ViMaSter: I think this should be `narrativeScript>Text< != null ` rather than `narrativeScript>Name< != null`?  

Also, when displaying file-related error messages, I'm always a fan of printing the absolute path. With Unity we should probably also mention that it's a [Resource](https://docs.unity3d.com/ScriptReference/Resources.html) we're trying to load, that way a user knows where the file should be.
```suggestion
        Debug.Assert(narrativeScriptText != null, $"Failed to load narrative script resource at 'InkDialogueScripts/{narrativeScriptName}'");
```

unity-ggjj/Assets/Scripts/TextDecoder/Preloading/ObjectPreloader.cs
```diff
@@ -99,7 +99,7 @@ private void LoadEvidence(string evidenceName)
         }
         catch (NullReferenceException exception)
         {
-            Debug.LogWarning($"{exception.GetType().Name}: Object at path {path} could not be loaded");
+            throw new NullReferenceException($"{exception.GetType().Name}: Object at path {path} could not be loaded");
```
ViMaSter: Nit: Not passing **the exception we're capturing** into **the new exception we're throwing**, means we discard the stack of the original exception.
  
The "deepest" line we'll see in the console with the current version of the code is  
`unity-ggjj/Assets/Scripts/TextDecoder/Preloading/ObjectPreloader.cs:102`
However with...
```suggestion
            throw new NullReferenceException($"{exception.GetType().Name}: Object at path {path} could not be loaded", exception);
```
...we'll see that this exception is thrown because of something in the `try` block. For example:  
`unity-ggjj/Assets/Scripts/TextDecoder/Preloading/ObjectPreloader.cs:94` or maybe even  
`unity-ggjj/Assets/Scripts/TextDecoder/Preloading/ObjectStorage.cs:58` if the `.Contains()` would be able to throw.

I just realized this a few days ago, so that's why I've haven't been pointing this out in previous PRs. 👀

unity-ggjj/Assets/Scripts/TextDecoder/Preloading/ObjectStorage.cs
```diff
@@ -35,7 +35,14 @@ public void Add(Object obj)
...
     public T GetObject<T>(string objectName) where T : class
     {
-        return _objects[objectName] as T;
+        try
+        {
+            return _objects[objectName] as T;
+        }
+        catch (KeyNotFoundException exception)
+        {
+            throw new KeyNotFoundException(objectName);
```
ViMaSter: Nit: [See earlier comment](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/241#discussion_r838903904).
```suggestion
            throw new KeyNotFoundException(objectName, exception);
```

unity-ggjj/Assets/Tests/PlayModeTests/Tools/StoryProgresser.cs
```diff
@@ -15,5 +19,82 @@ public IEnumerator ProgressStory()
...
+                        var courtRecordObjects = narrativeGameState.ObjectStorage.GetObjectsOfType<ICourtRecordObject>().ToList();
+
+                        if (courtRecordObjects.Any(courtRecordObject => courtRecordObject is ActorData && courtRecordObject.InstanceName == evidenceName))
+                        {
+                            yield return PressForFrame(Keyboard.cKey);
+                        }
+                        else if (courtRecordObjects.All(courtRecordObject => courtRecordObject.InstanceName != evidenceName))
+                        {
+                            yield break;
+                        }
```
ViMaSter: If I'm reading this correctly, we simply stop the coroutine, if no court record item the player is in possession of matches the name of our requested evidence.  
If that's the case, I would suggest, we instead throw an exception. That way, a test fails, if we're attempting to select evidence we don't have inside the court record currently.

unity-ggjj/Assets/Tests/PlayModeTests/Tools/StoryProgresser.cs
```diff
@@ -15,5 +19,82 @@ public IEnumerator ProgressStory()
...
+                        yield return SelectEvidence(evidenceName);
+                    }
+                    yield return choiceIndex switch
+                    {
+                        0 => PressForFrame(Keyboard.xKey),
+                        1 => PressForFrame(Keyboard.cKey),
+                        2 =>  SelectEvidence(evidenceName),
+                        _ => throw new ArgumentException($"Choice index can only be 0, 1, or 2 in GameMode {gameMode}")
+                    };
```
ViMaSter: Similar to you (rightfully!) questioning the use of string dependencies over hard dependencies, I try to avoid ["magic numbers" (numbers with implicit context)](https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants). : D

I see two ways to resolve this:
1. [as mentioned in this comment](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/241#discussion_r838940166), I would generally split up this method into two "cross-examination choice" and "dialogue choice" methods; then you could use an enum which would be as strongly typed as possible
2. define constants for these actions and use them whenever you call this in a cross-examination context; this would be a bit looser typing but better than pure numbers

```cs
private const int CONTINUE_STORY = 0;
private const int PRESS = 1;
private const int SELECT_EVIDENCE = 2;
```

unity-ggjj/Assets/Tests/PlayModeTests/Tools/StoryProgresser.cs
```diff
@@ -15,5 +19,82 @@ public IEnumerator ProgressStory()
...
+                    if (evidenceMenu != null && evidenceMenu.CanPresentEvidence)
+                    {
+                        var narrativeGameState = Object.FindObjectOfType<NarrativeGameState>();
+                        var courtRecordObjects = narrativeGameState.ObjectStorage.GetObjectsOfType<ICourtRecordObject>().ToList();
+
+                        if (courtRecordObjects.Any(courtRecordObject => courtRecordObject is ActorData && courtRecordObject.InstanceName == evidenceName))
+                        {
+                            yield return PressForFrame(Keyboard.cKey);
+                        }
+                        else if (courtRecordObjects.All(courtRecordObject => courtRecordObject.InstanceName != evidenceName))
```
ViMaSter: Nit: These LINQ expressions are quite a mouthful! : D  
I think this is one of the few cases, where I would suggest decorating lines with comments, as I also don't see any way of expressing them differently in code:
```suggestion
                        // when trying to select an actor as evidence, we first need to flip to the Actor Court record page
                        if (courtRecordObjects.Any(courtRecordObject => courtRecordObject is ActorData && courtRecordObject.InstanceName == evidenceName))
                        {
                            yield return PressForFrame(Keyboard.cKey);
                        }
                        // if none of the court records we have match the evidenceName, we exit this coroutine
                        else if (courtRecordObjects.All(courtRecordObject => courtRecordObject.InstanceName != evidenceName))
```

unity-ggjj/Assets/Tests/PlayModeTests/Tools/StoryProgresser.cs
```diff
@@ -15,5 +19,82 @@ public IEnumerator ProgressStory()
...
             Release(Keyboard.xKey);
         }
+
+        /// <summary>
+        /// Selects a choice and presses the x key to progress the story
+        /// </summary>
+        /// <param name="choiceIndex">The index of the choice to select</param>
+        /// <param name="gameMode">The GameMode the game is currently using</param>
+        /// <param name="evidenceName">The name of any evidence to present (if any)</param>
+        public IEnumerator SelectChoice(int choiceIndex, GameMode gameMode, EvidenceAssetName evidenceName)
```
ViMaSter: Nit: I appreciate the versatility of this method, but I think the legibility of it suffers a bit too much for it. I would suggest splitting this into a method used during cross-examinations and another for dialogue choices.

[Let me know what you think about this alternative](https://gist.github.com/ViMaSter/196d1eeea7f6048127b6aada5ffe6237) - I think it also solves the "magic number" problem I mentioned below a bit cleaner, while the asserts ensure that the wrong method is prevented from being called at the wrong time.  
Just whipped this up, since I think it's easier to talk about tangible code than just trying to articulate the changes and then having you do them by hand. Feel free to dismiss it or take parts of it that seem useful to you, would love to hear your feedback on it or if there's something I forgot to consider here!

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/NarrativeScripts/Case1Tests.cs
```diff
@@ -0,0 +1,159 @@
...
+        {
+            yield return SceneManager.LoadSceneAsync("Game");
+            _appearingDialogueController = Object.FindObjectOfType<AppearingDialogueController>();
+            _narrativeGameState = Object.FindObjectOfType<NarrativeGameState>();
+        }
+
+        [UnityTest]
+        [Timeout(1000000)]
+        [TestCaseSource(nameof(NarrativeScripts))]
+        public IEnumerator NarrativeScriptsRunWithNoErrors(TextAsset narrativeScriptText)
```
ViMaSter, line 40: Is that... a fully automated playthrough of our game? Passing on GitHub?!  This. Rocks.  
Seriously. 

[In 2014 I read a blog post about how Platinum Games had an automated bug checker for Bayonettta 2](https://www.platinumgames.com/official-blog/article/6968) that just kinda plays the game to make sure nothing's just randomly broken. 8 years later, I still haven't seen anything to that scale ever again.  
...and you kinda just did exactly that for our game.

Seeing someone write this kind of code in a production app that's not just "a DirectX render test library they wrote over the weekend", but actually has multiple developers working on it, would make me hire them. This is nuts and is going to be so insanely useful moving foward. Absolutely stunning work here. 👍

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/NarrativeScripts/Case1Tests.cs
```diff
@@ -0,0 +1,159 @@
...
+using UnityEngine.TestTools;
+using Object = UnityEngine.Object;
+
+namespace Tests.PlayModeTests.Scenes.NarrativeScripts
+{
+    public class Case1Tests
+    {
+        private static IEnumerable<TestCaseData> NarrativeScripts => Resources
+            .LoadAll<TextAsset>("InkDialogueScripts/Case1").OrderBy(textAsset => textAsset.name.Split('-')[1].PadLeft(2, '0'))
+                .Select(narrativeScript => new TestCaseData(narrativeScript).SetName(narrativeScript.name).Returns(null));
```
ViMaSter, line 19: TIL about `NUnit.Framework.TestCaseData`. This is crazy handy. 👍

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/NarrativeScripts/Case1Tests.cs
```diff
@@ -0,0 +1,159 @@
...
+        [UnitySetUp]
+        public IEnumerator SetUp()
+        {
+            yield return SceneManager.LoadSceneAsync("Game");
+            _appearingDialogueController = Object.FindObjectOfType<AppearingDialogueController>();
+            _narrativeGameState = Object.FindObjectOfType<NarrativeGameState>();
+        }
+
+        [UnityTest]
+        [Timeout(1000000)]
```
ViMaSter: Nit: With big time ranges like that, I prefer to write them in multiplications going from biggest to smallest unit:
```suggestion
        [Timeout(5 * 60 * 1000)]
```
This would limit the test to 5 minutes, while being a bit more readable.  
1,000,000 milliseconds is around 16.6666 minutes, but [based on the test run of this PR](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/pull/241/checks?check_run_id=5760419670), the longest scene (`1-13-TrialEnd`) only took around 3 minutes, so I figured 5 minutes might be an okay middleground?

---

One thing to keep in mind for the future: [The Unity Test Runner unfortunately doesn't support parallel runs](https://forum.unity.com/threads/parallel-tests.1072166/#post-7019491), so maybe once we have the second case covered automatically, we want to start looking into running a test job _per_ case.  
Imagine `PlayMode Scripts`, `PlayMode Scenes`, `PlayMode Case 1`, `PlayMode Case 2`, ...  
That way, the time our PRs are stuck in "did I break the tests? "-limbo is kept short, if two people gave their approval and all that was fixed was a typo.

<sup>(It's 2022 and we're forced to run sequential tests.)</sup>

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/NarrativeScripts/Case1Tests.cs
```diff
@@ -0,0 +1,159 @@
...
+                    {
+                        break;
+                    }
+                }
+
+                if (_narrativeScript.Story.canContinue)
+                {
+                    yield return ProgressStory();
+                }
+                else if (_narrativeScript.Story.currentChoices.Count > 0)
```
ViMaSter: Nit: Would it be possible to eliminate a few (or maybe even all) of the `else if`s by `continue;` or `break;`ing in the `if` that comes before it?

I personally find `else if` difficult to keep track of, as they implicitly require the inverse of the first `if` to be true. If that first `if` has a `continue` block however, I can completely forget about that first `if`'s conditional and just look at the second conditional in isolation.

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/NarrativeScripts/Case1Tests.cs
```diff
@@ -0,0 +1,159 @@
...
+                    {
+                        // If all choices visited, continue as normal
+                        yield return storyProgresser.SelectChoice(0, _narrativeScriptPlayer.GameMode, null);
+                    }
+                }
+                // No choices and cannot continue means the end of a story
+                else
+                {
+                    break;
+                }
```
ViMaSter: Nit: Similarly to the `else if` at the beginning of this method, would it be possible to hoist these "simple" conditionals up by inverting the conditional and then `continue;` or `break;` in them? (There's one pretty simple `else` at the top of this `while`-loop as well.)

Doing so, should reduce nesting a bit, and - especially as this is pretty complex code already - allow the reader to "pop conditionals of their 'so how did the processor get here again?' stack" by just quickly evaluating a conditional, reading the code that's acting on it and then moving past it.

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/NarrativeScripts/Case1Tests.cs
```diff
@@ -0,0 +1,159 @@
...
+            var storyProgresser = new StoryProgresser();
+            storyProgresser.Setup();
+
+            while (true)
+            {
+                _appearingDialogueController.AppearInstantly = true;
+                
+                if (NarrativeScriptHasChanged(_narrativeScript))
+                {
+                    // Are there choices remaining to explore?
```
ViMaSter: Nit: This method is complex to follow, but I'm not sure if there's much that can be done about it. Going against my usual "comments can lie, code doesn't" approach, I find these inline comments before each conditional to be incredibly helpful.  
I think I would actually suggest doing them for each conditional in this file because - admittedly - it's difficult to keep track of where inside this logic loop I currently am while reading it. : D

Curious to hear your thoughts about it!

ViMaSter (approved):
So this is abso-~~I so wanna swear on GitHub right now~~-lutely insane?!

- It's the complete first case
- it **actually** runs
- there's 83%+ total code coverage now
- THE GAME CAN PLAY ON ITS OWN
- **WITHOUT FAILURE**
- **_EVEN ON OUR GITHUB RUNNERS!_**

This is incredible.

I think there are a few technical things we can do differently (wouldn't be a PR from me otherwise, would it?), but honestly, simply to not make this absolute merge hell, I think this should go in sooner than later and just be a reference of "things we can address in another PR".

Here's a crazy proposal:
1. Merge this in
2. Make a PR that starts the game from `1-1 Intro` when you click "New Game" and get that merged
3. Get that build out to QA to get them to write gameplay bugs
4. In the meantime, fix all the technical bits and nits

Opinions?

---

First bugs I encountered are two tiny visual bugs.
1. Dan's "lean" sprite wasn't showing for a few seconds the first time it appeared in `1-2` while running in the editor.
2. Afterwards - during the line of the screenshot - Dan was gone and didn't appear even after a few minutes of waiting:
image

unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayerComponent.cs
PhoebeMitchell: It's supposed to be "continuing _at_ the beginning". Does it make more sense without the missing word?

unity-ggjj/Assets/Scripts/TextDecoder/Preloading/ObjectPreloader.cs
PhoebeMitchell: Weirdly this is something I do habitually in Java, but not in C# apparently. Good catch!

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/NarrativeScripts/Case1Tests.cs
PhoebeMitchell: Ooh, good tip for representing time.
And yeah I could see tests getting very long, so splitting them up more would be good.

unity-ggjj/Assets/Tests/PlayModeTests/Tools/StoryProgresser.cs
PhoebeMitchell: This is actually because the evidence to present is passed as the choice text. So for choices where the text isn't an evidence item we want to ignore them. Added a comment to make this clearer (and remind myself because I forgot what this was for 😅)

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/NarrativeScripts/Case1Tests.cs
PhoebeMitchell: Yeah, I realised whoever was reviewing this wasn't going to have a great time. Sometimes I find it can be difficult to deduce the reasoning behind test scripts so happy to continue writing comments.

unity-ggjj/Assets/Tests/PlayModeTests/Tools/StoryProgresser.cs
PhoebeMitchell: Admittedly this was kind of a Frankenstein's monster of a test that I just added stuff to when I ran into a problem playing the scripts. Building on your suggested code I've gone through and split it up into distinct methods that are called from Case1Tests. Wasn't fully happy with how this code looked so thanks for giving me a starting point in going about cleaning it up 👍 Let me know how I did

unity-ggjj/Assets/Resources/InkDialogueScripts/Case1/1-2-PreTrial.ink
```diff
@@ -0,0 +1,541 @@
...
+&DIALOGUE_SPEED:0.06
+&AUTO_SKIP:True
+I'm wha---
+&AUTO_SKIP:False
+
+&SHAKE_SCREEN:0.25,0.2
+&PLAY_SFX:smack
+&SET_POSE:Angry
+&DIALOGUE_SPEED:0.02
+BELIEVEABLE!
```
notexplosive: Missing a `&SPEAK:Dan` here. Arin says BELIEVEABLE in game (also isn't it BELIEVABLE?)

unity-ggjj/Assets/Resources/InkDialogueScripts/Case1/1-2-PreTrial.ink
```diff
@@ -0,0 +1,541 @@
...
+
+&PLAY_SFX:evidenceDing
+&ADD_EVIDENCE:Plumber_Invoice
+&SHOW_ITEM:Plumber_Invoice,Left
+&DIALOGUE_SPEED:0.06
+&NARRATE
+<align="center"><color={lightBlue}>Plumber Invoice has been added to the Court Record.
+&PLAY_SFX:evidenceShoop
+&HIDE_ITEM
+&WAIT:0.1
```
notexplosive: Design thoughts for the future:
Do we need to do this _every_ time we get evidence? Can we make some of these implied? This seems like a lot of "Narrative Code" that needs to be copy-pasted.

```
&NARRATE_LIGHT_BLUE  // ... or something to imply it's the omniscient gameplay narrator
Plumber Invoice has been added to the Court Record.
&ADD_EVIDENCE:Plumber_Invoice,Left // implies, ding, shoop, and wait 0.1 seconds.
```

I guess this has the added benefit that we can silently add evidence. But maybe that could be a separate `&ADD_EVIDENCE_SILENTLY` or something

unity-ggjj/Assets/Resources/InkDialogueScripts/Case1/1-3-TrialStart.ink
```diff
@@ -0,0 +1,302 @@
...
+<- COURT_TMPH
+
+&ADD_RECORD:Arin
+&ADD_RECORD:Dan
+&ADD_RECORD:Jory
+
+&ADD_EVIDENCE:AttorneysBadge
+&ADD_EVIDENCE:PlumberInvoice
+&ADD_EVIDENCE:Switch
+&ADD_EVIDENCE:Jory_Srs_Letter
```
notexplosive, line 14: ~~Does &LOAD_SCRIPT remember your already obtained evidence? I wonder if this is why I had duplicates?~~

We talked about this a bit on discord. This causes a bug, but removing it means we can't access scenes individually. Definitely need some design thought.

unity-ggjj/Assets/Scripts/TextDecoder/Preloading/ObjectPreloader.cs
```diff
@@ -99,7 +99,7 @@ private void LoadEvidence(string evidenceName)
         }
         catch (NullReferenceException exception)
         {
-            Debug.LogWarning($"{exception.GetType().Name}: Object at path {path} could not be loaded");
+            throw new NullReferenceException($"{exception.GetType().Name}: Object at path {path} could not be loaded", exception);
```
notexplosive: Feels a bit weird to catch a NullReferenceException and then throw a new NullReferenceException.

Maybe this should be its own exception type? Not blocking but something to think about.

Maybe we could have a generic `GameplayException` or `NarrativeException` for exceptions that happen while executing a narrative script?

unity-ggjj/Assets/Scripts/TextDecoder/Preloading/ObjectStorage.cs
```diff
@@ -35,7 +35,14 @@ public void Add(Object obj)
...
     public T GetObject<T>(string objectName) where T : class
     {
-        return _objects[objectName] as T;
+        try
+        {
+            return _objects[objectName] as T;
+        }
+        catch (KeyNotFoundException exception)
+        {
+            throw new KeyNotFoundException(objectName, exception);
```
notexplosive: ditto here with the exception

notexplosive (approved):
I didn't finish playing through Case 1 but I did read through the code.

First off WOOHHOOO! GO TEAM!! Way to go!!
Thank you moths especially for converting Case 1 over to the new format. This is really incredible.

image

I had duplicates of evidence in this scene (right before getting the livestream recording added to evidence). We talked about this on discord and this might be a design problem to sort out.

Dan was also invisible basically the whole game (he was there at the intro but that's about it)

image

unity-ggjj/Assets/Resources/InkDialogueScripts/Case1/1-2-PreTrial.ink
PhoebeMitchell: This could probably be converted to an Ink ~~function~~ knot which might end up looking something like `<‑ AddEvidence("PlumberInvoice")`

unity-ggjj/Assets/Resources/InkDialogueScripts/Case1/1-3-TrialStart.ink
PhoebeMitchell: Will leave this as is for now as we have an issue for it https://github.com/Studio-Lovelies/GG-JointJustice-Unity/issues/243

unity-ggjj/Assets/Resources/InkDialogueScripts/Case1/1-2-PreTrial.ink
PhoebeMitchell: Changed this into an ink knot which can be diverted to

PR #245, opened 2022-04-02
ActorSlots are now referenced to by name (`Witness`) rather than a one-based index (`1`) inside Ink Stories

### LINKED ISSUE #178: ActorSlots of scenes are referenced by name and use position of GameObject of Actor
## Description

Some scenes in our games support multiple actors in different locations on screen. For this we utilize `Actor Positions`.
Right now, these are
- created inside Unity for each scene and the offset values have to be set up by hand with pixel values
  image
- referenced inside `.ink`-files by their index (i.e. setting Ross as Witness in the scene of the screenshot above, would work by calling `&SET_ACTOR_POSITION:2,Ross` (as the index starts at 1 inside `.ink`-files)

## Acceptance Criteria

This story will be finished when
- manually entering a pixel offset is no longer required, as the position of the GameObject of the `Actor` of a slot is used instead
- the name of the GameObject of the `Actor` used as a slot is used instead of the "index+1":
  - `&SET_ACTOR_POSITION:2,Ross` would become `&SET_ACTOR_POSITION:Witness_Actor,Ross`

ViMaSter:
**Note: This PR uses [semantic commit messages](https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716) for each commit.** (thx @notexplosive!)

## Summary
This branch implements #178 inside 690c15d4b08ad6790bb8a888197f30c3704f7a98.  
This change does **not** impact the game from a player's perspective in any way.
 
Out of convenience, 6a96a805dcbd071fa2c843f5c115d85e25b07111 also addresses various inconsistencies inside files I had open after implementing the feature; they are not directly tied to the feature and isolated to that commit.

### Before
```
&SET_ACTOR_POSITION:1,Arin
&JUMP_TO_POSITION:2
```

### After
```
&JUMP_TO_POSITION:Witness
```

## Testing instructions
1. Run any Story that was modified in this PR
2. Ensure the scene operates identically to before this PR

## Additional information

- `[ ]` Changes UI
  - closes #178


docs/methods/Dialogue.md
```diff
@@ -70,7 +70,7 @@ Examples:
 
 ⏲ Instant
 
-Hides the dialogue textbox until the next line of dialogue.
+Hides the dialogue text-box until the next line of dialogue.
```
lyndsiWilliams: Nit: I think this is fine as textbox

docs/methods/Dialogue.md
```diff
@@ -79,7 +79,7 @@ Examples:
 
 ⏳ Waits for completion
 
-Plays an "Objection!" animation and soundeffect for the specified actor.
+Plays an "Objection!" animation and sound-effect for the specified actor.
```
lyndsiWilliams: Nit: Could this be `sound effect` instead? Noun + noun doesn't get a hyphen.

docs/methods/Dialogue.md
```diff
@@ -88,7 +88,7 @@ Examples:
 
 ⏳ Waits for completion
 
-Plays a "Take that!" animation and soundeffect for the specified actor.
+Plays a "Take that!" animation and sound-effect for the specified actor.
```
lyndsiWilliams: Nit: Same as above, could this be `sound effect` instead?

docs/methods/Dialogue.md
```diff
@@ -97,7 +97,7 @@ Examples:
 
 ⏳ Waits for completion
 
-Plays a "Hold it!" animation and soundeffect for the specified actor.
+Plays a "Hold it!" animation and sound-effect for the specified actor.
```
lyndsiWilliams: Nit: Same as above, could this be `sound effect` instead?

unity-ggjj/Assets/Scripts/AudioController/AudioController.cs
```diff
@@ -67,22 +66,11 @@ public void PlaySong(AudioClip song, float transitionTime)
...
-    public void PlaySfx(string sfx)
-    {
-        AudioClip soundEffectClip = _narrativeGameState.ObjectStorage.GetObject<AudioClip>(sfx);
-        PlaySfx(soundEffectClip);
-    }
+    private AudioSource CreateAudioSource(string gameObjectName) => new GameObject(gameObjectName) {
+            transform = {
+                parent = transform
+            }
+        }.AddComponent<AudioSource>();
```
lyndsiWilliams:
```suggestion
    private AudioSource CreateAudioSource(string gameObjectName) => new GameObject(gameObjectName) {
            transform = { parent = transform }
        }.AddComponent<AudioSource>();
```
Nit: I can't remember if we had a code style to specify brackets in these situations, and this could definitely be the JS-lover in me, but this seems more readable on one line IMO. What do you think, would that work or is that too not-C#-y?

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -75,7 +75,7 @@ private void APPEAR_INSTANTLY()
         OnActionDone?.Invoke();
     }
 
-    /// <summary>Hides the dialogue textbox until the next line of dialogue.</summary>
+    /// <summary>Hides the dialogue text-box until the next line of dialogue.</summary>
```
lyndsiWilliams: Yet another text nit: I think this is fine as textbox

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -319,7 +319,7 @@ private void WAIT(float seconds)
         NarrativeGameState.SceneController.Wait(seconds);
     }
 
-    /// <summary>Plays an "Objection!" animation and soundeffect for the specified actor.</summary>
+    /// <summary>Plays an "Objection!" animation and sound-effect for the specified actor.</summary>
```
lyndsiWilliams: Sprinkling text nits all over this PR: Can this be `sound effect`?

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -328,7 +328,7 @@ private void OBJECTION(ActorAssetName actorName)
         SHOUT(actorName, "Objection", true);
     }
 
-    /// <summary>Plays a "Take that!" animation and soundeffect for the specified actor.</summary>
+    /// <summary>Plays a "Take that!" animation and sound-effect for the specified actor.</summary>
```
lyndsiWilliams: Here's another one: Can this be `sound effect`?

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -337,7 +337,7 @@ private void TAKE_THAT(ActorAssetName actorName)
         SHOUT(actorName, "TakeThat", true);
     }
 
-    /// <summary>Plays a "Hold it!" animation and soundeffect for the specified actor.</summary>
+    /// <summary>Plays a "Hold it!" animation and sound-effect for the specified actor.</summary>
```
lyndsiWilliams: Last one I promise: Can this be `sound effect`?

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoderBase.cs
```diff
@@ -32,50 +32,45 @@ public void InvokeMatchingMethod(string actionLine)
...
         var optionalParameters = methodParameters.Count(parameter => parameter.IsOptional);
         if (parameters.Length < (methodParameters.Length - optionalParameters) || parameters.Length > (methodParameters.Length))
         {
-            throw new TextDecoder.Parser.ScriptParsingException($"'{action}' requires {(optionalParameters == 0 ? "exactly" : "between")} {(optionalParameters == 0 ? methodParameters.Length.ToString() : $"{methodParameters.Length-optionalParameters} and {methodParameters.Length}")} parameters (has {parameters.Length} instead)");
+            throw new ScriptParsingException($"'{action}' requires {(optionalParameters == 0 ? "exactly" : "between")} {(optionalParameters == 0 ? methodParameters.Length.ToString() : $"{methodParameters.Length-optionalParameters} and {methodParameters.Length}")} parameters (has {parameters.Length} instead)");
         }
 
         var parsedMethodParameters = new List<object>();
         // For each supplied parameter of that action...
         for (var index = 0; index < parameters.Length; index++)
```
lyndsiWilliams, line 72: I know you didn't write this bit and this could definitely be saved for a code cleanup ticket or something but could [foreach](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/arrays/using-foreach-with-arrays) be used here?

lyndsiWilliams (approved):
I've got _a bunch_ of nits here but nothing merge-blocking! Functionally the code looks great and it all works great locally, all tests passed ✅

unity-ggjj/Assets/Scripts/AudioController/AudioController.cs
PhoebeMitchell: Personally I feel this reads a little confusingly with a lambda. At first glance I thought you were passing the lambda as an argument. Would prefer the following:
```suggestion
    private AudioSource CreateAudioSource(string gameObjectName)
    {
        return new GameObject(gameObjectName)
        {
            transform =
            {
                parent = transform
            }
        }.AddComponent<AudioSource>();
    }
```

unity-ggjj/Assets/Scripts/BGScenes/BGScene.cs
```diff
@@ -1,89 +1,82 @@
...
-    [Tooltip("List of potential actor slots in the scene that can be filled with ActorData and an associated camera position")]
-    [SerializeField] private List<ActorSlot> _actorSlots;
-
-    private ActorSlot _currentActorSlot;
+    [Serializable]
+    public class ActorSlot
+    {
+        public Actor AttachedActor { get; set; }
+        public Vector2Int Position { get; set; }
+    }
```
PhoebeMitchell, line 18: Could this class have its own file?

unity-ggjj/Assets/Scripts/BGScenes/BGScene.cs
```diff
@@ -1,89 +1,82 @@
...
-using UnityEngine.Serialization;
 
 public class BGScene : MonoBehaviour
 {
-    [Tooltip("List of potential actor slots in the scene that can be filled with ActorData and an associated camera position")]
-    [SerializeField] private List<ActorSlot> _actorSlots;
-
-    private ActorSlot _currentActorSlot;
+    [Serializable]
+    public class ActorSlot
```
PhoebeMitchell, line 14: Could this class be a struct instead? Seems to only be storing data.

unity-ggjj/Assets/Scripts/BGScenes/BGScene.cs
```diff
@@ -1,89 +1,82 @@
...
     /// Sets the active actor to the first actor in the scene.
     /// </summary>
     private void Awake()
     {
-        if (SupportsActorSlots())
+        static string NormalizeActorName(Actor actor) => actor.name.Replace("_Actor", "");
+        static ActorSlot GenerateSlotDataForActor(Actor actor)
+        {
+            var position = actor.GetComponent<Transform>().localPosition * 100;
+            return new ActorSlot()
```
PhoebeMitchell: nit: () is redundant

unity-ggjj/Assets/Scripts/BGScenes/BGScene.cs
```diff
@@ -1,89 +1,82 @@
...
-    }
 
-    /// <summary>
-    /// Sees whether the current scene has support for actor slots
-    /// </summary>
-    /// <returns>whether this bg-scene has support for actor slots</returns>
-    public bool SupportsActorSlots()
-    {
-        return _actorSlots.Count > 0;
+        CurrentActorSlot = _actorSlotBySlotName.First().Value;
```
PhoebeMitchell: Simplify this to `_actorSlotBySlotName.FirstOrDefault().Value;` without the if statement?

unity-ggjj/Assets/Scripts/BGScenes/SceneController.cs
```diff
@@ -223,48 +223,36 @@ private bool HasFullScreenAnimationPlayer()
...
+        _narrativeGameState.ActorController.SetActiveActorObject(_activeScene.CurrentActorSlot.AttachedActor);
+        PanCamera(seconds, _activeScene.CurrentActorSlot.Position);
     }
 
     /// <summary>
     /// Jump cuts to the target sub position, if the bg-scene has sub positions.
     /// </summary>
-    /// <param name="oneBasedSlotIndex">Target actor slot, 1 based.</param>
-    public void JumpToActorSlot(int oneBasedSlotIndex)
+    /// <param name="slotName">Name of an actor slot of the currently active scene</param>
```
PhoebeMitchell: Nit: I think "Name of an actor slot _in_ the currently active scene" reads better

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -483,13 +484,13 @@ private void PLAY_EMOTION(ActorPoseAssetName poseName, ActorAssetName? optional_
     }
 
     /// <summary>Sets the target sub-position of the current bg-scene to have the target actor.</summary>
-    /// <param name="oneBasedSlotIndex">Whole number representing the target sub-position of the currently active scene</param>
+    /// <param name="slotName">Name of an actor slot of the currently active scene</param>
```
PhoebeMitchell: Same again, "in the currently active scene"

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoderBase.cs
```diff
@@ -161,15 +151,49 @@ public bool IsAction(string line)
...
-    protected abstract void PLAY_SFX(SfxAssetName sfx);
-    protected abstract void PLAY_SONG(SongAssetName songName, float optional_transitionTime = 0);
-    protected abstract void SCENE(SceneAssetName sceneName);
-    protected abstract void SHOW_ITEM(CourtRecordItemName itemName, ItemDisplayPosition itemPos);
-    protected abstract void ACTOR(ActorAssetName actorName);
-    protected abstract void SPEAK(ActorAssetName actorName);
-    protected abstract void SPEAK_UNKNOWN(ActorAssetName actorName);
-    protected abstract void THINK(ActorAssetName actorName);
-    protected abstract void SET_ACTOR_POSITION(int oneBasedSlotIndex, ActorAssetName actorName);
+    protected virtual void ADD_EVIDENCE(EvidenceAssetName evidenceName)
```
PhoebeMitchell, line 154: What's the reason for making all these virtual? "NotImplementedException" implies these should be implemented at some point but really they just serve the same purpose as an interface.

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/MainMenu/ViaKeyboard.cs
```diff
@@ -11,13 +11,13 @@ namespace Tests.PlayModeTests.Scenes.MainMenu
     public class ViaKeyboard
     {
         private readonly InputTestTools _inputTestTools = new InputTestTools();
-        private Keyboard Keyboard;
+        private Keyboard _keyboard;
```
PhoebeMitchell: Could this be a property instead?
`private Keyboard Keyboard => _inputTestTools.Keyboard`

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/MainMenu/ViaMouse.cs
```diff
@@ -12,13 +12,13 @@ namespace Tests.PlayModeTests.Scenes.MainMenu
     public class ViaMouse
     {
         private readonly InputTestTools _inputTestTools = new InputTestTools();
-        private Mouse Mouse;
+        private Mouse _mouse;
```
PhoebeMitchell: Same, here. Could be a property.

PhoebeMitchell (approved):

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoderBase.cs
ViMaSter: I actually did write that haha. I think it just got extracted it into this `ActionDecoderBase` a few commits ago. : D

I much prefer `foreach` or even `.Select` over regular `for` loops too, it's just that in this case [I've used the index explicitly](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/dde716cad305e8646f8aefe81e0467ee8d022ba7/unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoderBase.cs#L91) later down the line to generate "more human-readable error messages".

If you write something line `&SHOW_ITEM:Bent_Coins;Left` in our Ink Stories but accidentally write `&SHOW_ITEM:Bent_Coins;>K<eft` instead, the line above generates this error:
`'Keft is incorrect as parameter #2 (itemPos) for action 'SHOW_ITEM': Cannot convert 'Keft' into an ItemDisplayPosition (valid values include: 'Left', 'Right', 'Middle')`

To make the hint `parameter #2` work, the index is used. I think it's useful, but I also rarely write stories in Ink, so I'm not sure how useful that is; if you feel it's redundant, I'm happy to remove it!

unity-ggjj/Assets/Scripts/AudioController/AudioController.cs
ViMaSter: Actually just updated it based on @lyndsiWilliams comments before I saw your review! 🙈

[Is this new format](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/feature/178-actor-slots-by-name/unity-ggjj/Assets/Scripts/AudioController/AudioController.cs#L69-L72) an okay middle-ground for you or would you like to still see some changes to it?

unity-ggjj/Assets/Scripts/BGScenes/BGScene.cs
ViMaSter: It actually was before. : D  
And I actually do agree with the "only data? go with struct!" sentiment.

My reasoning was, that structs can't be `null`, unless we explicitly mark it as such and  `CurrentActorSlot` has to be `null` at startup, as we can only determine its value, once we know if the scene supports multiple actor slots or not.

Pretty open to both however; either making `ActorSlot` a `struct` and `CurrentActorSlot` a `?`/`nullable` or keeping it a class.  
Ultimately, which one would feel more natural to you? I'll just go with that then!

unity-ggjj/Assets/Scripts/AudioController/AudioController.cs
PhoebeMitchell: probably could just be one-lined with a lambda in the new format. It's just having the open parenthesis at the end of the line threw me off originally because my brain immediately went to "method starts here"

unity-ggjj/Assets/Scripts/BGScenes/BGScene.cs
PhoebeMitchell: Keep it a class 👍

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoderBase.cs
ViMaSter: For game code, that definitely still holds true, both before and after this PR, but I added a few new tests in this PR, that are a bit easier to set up because of this. [[1]](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/420bc43896364e9a4ea8e7bbb388ebe5d55b5904/unity-ggjj/Assets/Tests/EditModeTests/TextDecoder/ActionDecoderTests/UnsupportedParameterTypeTests.cs#L9-L15) [[2]](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/420bc43896364e9a4ea8e7bbb388ebe5d55b5904/unity-ggjj/Assets/Tests/EditModeTests/TextDecoder/ActionDecoderTests/ImplementationTests.cs#L60-L62) These were added to increase coverage and actually depend on these methods not being implemented. (They verify, that the correct exceptions are thrown for incorrectly set up Parsers or ActionDecoder derivatives.)

While working on that, I noticed, that every class that derived from `ActionDecoderBase` had to add a block of `NotImplementedExceptions` for every method, just to get code to compile. By adding a base implementation that throws these exceptions these derivatives automatically compile, it's just that if specific behaviors are triggered without being overridden beforehand, they now raise exceptions implicitly, rather than having to explicitly state it in each derivative.

Definitely open to other ideas on how to handle this without causing too much code duplication, if there's a better way to go about this!

unity-ggjj/Assets/Scripts/AudioController/AudioController.cs
ViMaSter: Going to cc @lyndsiWilliams here, just feel free to vote on a format, I'll just configure my IDE to use whichever. : D
```cs
/// 1
private AudioSource CreateAudioSource(string gameObjectName)
{
    return new GameObject(gameObjectName)
    {
        transform =
        {
            parent = transform
        }
    }.AddComponent<AudioSource>();
}


/// 2
private AudioSource CreateAudioSource(string gameObjectName) => new GameObject(gameObjectName) { transform = { parent = transform } }.AddComponent<AudioSource>(); }

/// 3
private AudioSource CreateAudioSource(string gameObjectName)
{
    return new GameObject(gameObjectName) { transform = { parent = transform } }.AddComponent<AudioSource>();
}
```

unity-ggjj/Assets/Scripts/BGScenes/BGScene.cs
ViMaSter: That's a good idea generally. The issue here is, that right now it's part of the `BGScene` class, as it is only used there. Unless we make `BGSCene` a `partial` class, I don't think we can both move move it to a separate file and let it be part of that class?

An alternative would be to use a namespace instead of making it a sub-class - since we actually never talked about namespaces, I opened up https://github.com/Studio-Lovelies/GG-JointJustice-Unity/discussions/250. (Spoiler: I'm a big fan.)
Which approach would you prefer? Or did you have another alternative in mind all together? If we go for the namespaces, I'd actually suggest that we make that part of whatever PR we will have that resolves #173 instead.

PhoebeMitchell: So the idea is to limit its usage elsewhere? That's fair enough 👍 

Side note: does it still need to be serializable? Doesn't seem to be used in Unity Editor at all.

ViMaSter: > So the idea is to limit its usage elsewhere? That's fair enough 👍

Exactly! I don't think there's a way to prevent instances of it from being created outside of `BGScene` (a private constructor would also prevent `BGScene` from doing so), but at least you're forced to write `new BGScene.ActorSlot()` and that might make you question whether or not that is actually what you should be doing. Going to add a comment for it too!

> Side note: does it still need to be serializable? Doesn't seem to be used in Unity Editor at all.

Good catch, it actually doesn't anymore, since it's just a data class that gets populated based on the "slot GameObject"'s name and position!

docs/methods/Scene.md
```diff
@@ -103,7 +103,7 @@ Examples:
 
 ## JUMP_TO_POSITION
 Values: 
-  - Whole number representing the target sub-position of the currently active scene
+  - Name of an actor slot in the currently active scene
```
notexplosive, line 106: Maybe "name of the actor"? Is the concept of a "slot" even relevant anymore?

docs/methods/Scene.md
```diff
@@ -114,7 +114,7 @@ Examples:
 
 ## PAN_TO_POSITION
 Values: 
-  - Whole number representing the target sub-position of the currently active scene
+  - Name of an actor slot in the currently active scene
```
notexplosive, line 117: ditto here

unity-ggjj/Assets/Scripts/Actor/ActorController.cs
```diff
@@ -29,17 +28,20 @@ public void OnSceneChanged(BGScene newScene)
     }
 
     /// <summary>
-    /// Set the target GameObject to be considered the active actor to manipulate when actions are triggered in  the dialogue script
+    /// Set the target GameObject to be considered the active actor to manipulate when actions are triggered in the dialogue script
```
notexplosive, line 31: Commit notation nit:
I like `Various ReSharper Cleanups` because you're clarifying that I can basically skip this commit. Although I think it would be even better if I could identify what cleanup tools you used.

It would be helpful if you either:
- listed the commands you ran in the commit message (ie: `Format Solution, Make things readonly, Use var keyword` etc)
- Or did each cleanup step as its own commit. (this has the added benefit of reverting just one cleanup step if it accidentally caused a bug)

(not blocking!! just food for thought, please do not rewrite your history for this)

unity-ggjj/Assets/Scripts/BGScenes/BGScene.cs
notexplosive: (unresolving to get my 2cents in)

Instead of null, you could have a static instance of ActorSlot that represents the EmptyActorSlot. Or-- (a little more heavyweight) you could have an IActorSlot interface which is implemented by ActorSlot and EmptyActorSlot (a whole type just for the empty actor slot).

That way you get the benefit of a "null" without the drawback of crashing when you run `nullThing.LiterallyAnything()`

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoderBase.cs
notexplosive: I wish foreach would give you the value _and_ the index.

Lua has a concept of `ipairs` where you can say `for i, v in ipairs(list)` and it creates a for loop where `i` is the index and `v` is the iterated value. I wish other languages had that.

notexplosive (approved):

docs/methods/Scene.md
ViMaSter: > Is the concept of a "slot" even relevant anymore?

Unless I'm misunderstanding something, I think it still is? The actor slot is something like `Witness`, `Defense`, `Prosecution`, the actor would be `Arin`, `Dan` or `Ross`. (So far that's only used in the `TMPHCourt` scene however.)

unity-ggjj/Assets/Scripts/BGScenes/BGScene.cs
ViMaSter: Hmm... I think that having a slot being `null` and crashing if it's not set, is actually what I'd expect out of a not-set `ActorSlot`. At least no use-case for a "gracefully handled empty slot" comes to my mind in our current game, but let me know if I'm missing something here!

That said, generally, it's a pattern I see used far too rarely as well; the idea of useful pre-defined static constants to compare against (if they're an expected state of the system), is always good thing (thinking of things like `Vector3.zero`, `Quaternion.identity`, `Envionment.NewLine`, `Mathf.Epsilon`). I'm just not sure how it could be applied for `ActorSlots`.

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoderBase.cs
ViMaSter: I believe you could (ab)use a `.Select` LINQ expression to achieve that behavior. I just personally don't find it very readable:
```cs
foreach (var (item, index) in myValues.Select((value, i) => (value, i)))
{
    Console.WriteLine($"{index}: {item}");
}
```
I think I'd prefer to follow a simple "no `for`, unless you need the index" rule, especially since LINQ is usually considered [pretty slow for most of its jobs pre .NET 5](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-5/) (warning: **lots** of text; but also [**lots** of pretty descriptive numbers like these](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-5/#:~:text=Array.Sort(_array)%3B%0A%20%20%20%20%7D%0A%7D-,Type,Ratio,-DoubleSorting)].  
Most of the time, I find the runtime performance to be a worthwhile compromise over code readability. In this case, it seems a bit overkill to make a for loop prettier to me.

That said, if you both agree that we should use the LINQ expression, I'm okay with changing it, as performance isn't too big of a concern for us anyway!

unity-ggjj/Assets/Scripts/BGScenes/BGScene.cs
notexplosive: I think I was just seeing "I want this struct to be a class because I want nullability" and thought the [null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern) would come to the rescue.

Maybe not in this case though, feel free to re-resolve

docs/methods/Scene.md
notexplosive: Gotcha, thanks for clarifying!

unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoderBase.cs
notexplosive: I do not have a strong opinion

unity-ggjj/Assets/Scripts/AudioController/AudioController.cs
PhoebeMitchell: Any of these are fine for me

ViMaSter: Went with `1` for now, just so this PR doesn't dangle around for too long. : D
Always open for a [Discussions thread](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/discussions) about it, If it's something we want to solidify further or change later down the line!

PR #251, opened 2022-04-04
Feature/234 &LOAD_SCENE action

### LINKED ISSUE #234: &LOAD_SCENE action
## Description
An action needs to be added to ~start the credits.~ load other scenes, for example the credits scene.
It should be something like ~`&CREDITS`~ `&LOAD_SCENE:SceneName` where `SceneName` is the name of the scene to load e.g., `Credits`

## Acceptance Criteria
Scenes can be loaded via an action.

PhoebeMitchell:
## Summary

Adds a &LOAD_SCENE action so scenes like "Credits" can be loaded via Ink story.

## Testing instructions

Run the new test.
Play the Ink story "SceneLoadingTest"
Adds "&LOAD_SCENE:Credits" to the end of case 1

## Additional information

- `[ ]` Changes UI
  
closes #234


unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoder.cs
```diff
@@ -577,6 +578,17 @@ private void ADD_FAILURE_SCRIPT(FailureScriptAssetName failureScriptName)
         NarrativeGameState.NarrativeScriptStorage.AddFailureScript(failureScriptName);
         OnActionDone?.Invoke();
     }
+
+    /// <summary>
+    /// Loads a Unity scene
+    /// </summary>
+    /// <param name="sceneName" validFiles="Assets/Scenes/*.unity">The name of the scene to load</param>
+    /// <example>&amp;LOAD_SCENE:Credits</example>
+    /// <category>Script Loading</category>
```
ViMaSter, line 587: Who needs to look at users playing the game, when you can just look at other developers using a feature you've built!

So awesome to see the doc generator scaling well and being intuitive to extend. 👍

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/CreditsTests.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class CreditsTests
+    {
+        [UnityTest]
+        public IEnumerator CreditsCanBeLoadedViaAction()
+        {
+            yield return SceneManager.LoadSceneAsync("Game");
+            var gameState = Object.FindObjectOfType<NarrativeGameState>();
+            var actionDecoder = Object.FindObjectOfType<ActionDecoderComponent>();
+            actionDecoder.Decoder.NarrativeGameState = gameState;
+            actionDecoder.OnNewActionLine("&LOAD_SCENE:Credits\n");
```
ViMaSter, line 20: Nit: Just to verify the SceneLoader actually doing its job, I would suggest asserting, that we're not actually starting in the `Credits` scene, and only are there, after the line has been parsed.
```suggestion
            Assert.AreNotEqual(SceneManager.GetActiveScene().name, "Credits");
            actionDecoder.OnNewActionLine("&LOAD_SCENE:Credits\n");
```

That all aside: It's so rare to see simple tests these days, so this is super well written. 👍

ViMaSter (approved):
Reading this PR, I noticed just how simple the needed change for this feature is. As it should be, one might say, but that hasn't been the case for quite a while. I think this new architecture we're going for is paying off. Not to mention the added stability. 👀👌

Great work and something that actually moves the game forward, as usual. 👍

unity-ggjj/Assets/Tests/PlayModeTests/Scenes/CreditsTests.cs
PhoebeMitchell: I would assume SceneManager.LoadSceneAsync("Game") would either load the Game scene or give an error, but I've added your suggestion as a sanity check (as we're clearly not above those going by my other PR :P)

notexplosive (approved):
gameing

PR #254, opened 2022-04-06
bugfix/Cannot navigate main menu with keyboard after returning to main menu through pause menu

### LINKED ISSUE #240: [BUG] Cannot navigate main menu with keyboard after returning to main menu through pause menu
## Describe the bug
When you click the back to menu button in the pause menu you're brought back to the main menu, but no button is selected so you cannot navigate using the keyboard.

## Steps To Reproduce
Start the game in the Game scene.
Open the pause menu with the Escape key and click "Back to menu".
Try to navigate the menu with the keyboard. You cannot.
Try to navigate the menu with the mouse. Buttons are selected and you can now navigate using the keyboard again.

## Expected behavior
Navigation should be possible after returning to the main menu without using the mouse.

PhoebeMitchell:
## Summary

Removes some code that sets the currently selected game object to null when a menu is disabled. When returning to the main menu from the pause menu the new game button would become selected, but the then pause menu would be disabled setting the currently selected game object to null.

Though I can't imagine I originally added this code for nothing, I'm not quite sure what its purpose was, so please check to see if anything is broken.

## Testing instructions

Start the Game scene and go back to the main menu via the pause menu. Check that you can navigate without using the mouse.
Make sure the rest of the game still works.
Run the tests.

## Additional information

- `[ ]` Changes UI
  
closes #240

### DIFF
--- a/unity-ggjj/Assets/Scripts/Menu/Menu.cs
+++ b/unity-ggjj/Assets/Scripts/Menu/Menu.cs
@@ -69,15 +69,4 @@ public class Menu : MonoBehaviour
         }
         _initiallyHighlightedButton.Select();
     }
-
-    /// <summary>
-    /// Called when the menu is disabled. Should be used to set DialogueController to not busy.
-    /// </summary>
-    private void OnDisable()
-    {
-        if (EventSystem.current != null)
-        {
-            EventSystem.current.SetSelectedGameObject(null);
-        }
-    }
 }


ViMaSter (approved):
~~Have a bug? Just delete the code. Bug gone. 👍~~

I _think_ this was needed to support additionally loading a scene when there is an EventSystem present in the current **and** next scene and we've loaded the scene additively. For example, switching from the MainMenu to the Game, before we went down the monoscene route. This should be superfluous by now. 

Game->Main Menu->Game->Main Menu with just the keyboard works fine on Win x64 and the tests also pass on GitHub. Looks good to me!

notexplosive (approved):
good.

PR #255, opened 2022-04-08
243 figure out how to preserve evidence from one scene into the next

### LINKED ISSUE #243: Figure out how to preserve evidence from one scene into the next
## Description

There are two things I want:

**If I'm transitioning from scene_1.ink to scene_2.ink** I want the game to remember evidence/court record I obtained from scene1 as I enter scene2. I don't want scene_2 to need to re-add all the things I got in scene_1 (that's a technical debt timebomb).

BUT! at the same time...

**If I'm loading a scene_2.ink ad-hoc from the editor** I _still_ want it to know what evidence I obtained in scene_1.

## Acceptance Criteria

This story will be finished when evidence is preserved from scene to scene without duplicating AND a developer can still load any arbitrary scene and have the correct evidence.

PhoebeMitchell:
## Summary

The starting evidence at the beginning of a story is now added using Ink knots as discussed on Discord. Each Ink knot calls the knot for the previous chapter. This reduces duplicate actions.

Also fixes a bug where the court record was not cleared between stories leading to duplicate evidence.

## Testing instructions

Run the tests. Does the game run still?
Play through a script until the end. Open the evidence menu. Are there any duplicated items?

## Additional information

- `[ ]` Changes UI
  
closes #243


unity-ggjj/Assets/Resources/InkDialogueScripts/Case1/1-9-RossCrossExamination3.ink
```diff
@@ -2,26 +2,12 @@ INCLUDE ../Options.ink
...
-&ADD_RECORD:Ross
-
-&ADD_EVIDENCE:AttorneysBadge
-&ADD_EVIDENCE:PlumberInvoice
-&ADD_EVIDENCE:Switch
-&ADD_EVIDENCE:Jory_Srs_Letter
-&ADD_EVIDENCE:LivestreamRecording
-&ADD_EVIDENCE:JorysBackpack
-&ADD_EVIDENCE:StolenDinos
-&ADD_EVIDENCE:BentCoins
```
notexplosive, line 23: It makes me happy to see these big blocks of text go away.... but at what cost?

unity-ggjj/Assets/Resources/InkDialogueScripts/Case1/StartingEvidence.ink
```diff
@@ -0,0 +1,28 @@
+=== Part2StartingEvidence
```
notexplosive: Why does it start at 2?

Generally I think we could make these names more informative, maybe the names could imply what evidence they hold?

I was thinking something like:
`Part2StartingEvidence` -> `StartingEvidence`
`Part3StartingEvidence` -> `Jory_Switch_Plumber_Letter`
`Part4StartingEvidence` -> `TutorialBoy_Recording_Backpack` (lil dicey here because now I'm omitting information, where do you draw the line?)
`Part5StartingEvidence` -> `Ross_BentCoints`

Or maybe we just divide them into "evidence groups" and just call them `EvidenceGroup1` - `EvidenceGroup4`? Which isn't much better than "part."

Not prescribing anything, just thinking out loud.

notexplosive (approved):
We trade duplication for complexity, I think this is a good trade.

unity-ggjj/Assets/Resources/InkDialogueScripts/Case1/StartingEvidence.ink
PhoebeMitchell: The parts line up with the script numbers. Part2 -> 1-2, Part3 -> 1-3 etc. Knot names aren't allowed to have dashes though (afaik) so changed "1-" to "part".
Looking at the Lua build just now I realise the attorney's badge and Dan and Arin are present in the intro scene too so yeah it probably should be Part1.

Honestly I think PartXStartingEvidence is good. It's the starting evidence for part X. Admittedly that kind of breaks down after Part 5 when no more evidence is added. Just having the latest evidence added is a bit misleading as it omits that fact that they also add all preceding evidence.

ViMaSter (approved):
This seems like the best solution to me; the biggest reason for it being:

https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/1aa611a834ad9a1eac33ca9be766e01d07874b53/unity-ggjj/Assets/Scripts/NarrativeScript/NarrativeScriptPlayer/NarrativeScriptPlayerComponent.cs#L53-L55

With that...
- every story, starts from a clean slate
  - got there by the `DebugGameStarter`? Cool. Don't care.
  - got there by the in-game chapter select? Neat. Don't care either.
  - got there by progressing through the game? Good for you. Still don't care though.
  - got there because it's an automated test running? Perfect. Not my concern either.
- if a Story decides to be based on another, it **can** do that, we don't enforce it however
  - if we want to group those, we can too (like we do right now and it looks great)
- **there is no context, that isn't visible just from looking at Ink Stories alone**
  - not in the savefiles
  - nor in any other global object C# takes care of

I remember this being quite the discussion, but I think this solution you both came up with and implemented, serves our needs perfectly! 👍

PR #256, opened 2022-04-08
253 player can access chapter via case select menu

### LINKED ISSUE #253: Player can access chapter via case select menu
## Description
Players should be able to access chapters via the case select menu. The case select menu is already there, but the buttons need to be linked up to ink stories. Use the Lua build text files in Assets>Resources>Story to see where each chapter is supposed to start.

...

## Acceptance Criteria

This story will be finished when the player can access each chapter via the case select menu.

PhoebeMitchell:
## Summary

Assigns Ink stories to the chapter select menu buttons as well as code to make the chapter select buttons function.
Chapters roughly align to those in the Lua build.

Kapture 2022-04-09 at 00 46 28

## Testing instructions

Try to start the game from the chapter selection menu.

## Additional information

- `[ ]` Changes UI
  
closes #253


unity-ggjj/Assets/Scripts/SceneLoading/SceneLoader.cs
```diff
@@ -14,9 +14,9 @@ public class SceneLoader : MonoBehaviour, ISceneLoader
 {
     public bool Busy { get; private set; }
 
-    [Tooltip("Assign a narrative script to play on scene load")]
-    [SerializeField]
-    private TextAsset _narrativeScript;
+    [field: Tooltip("Assign a narrative script to play on scene load")]
+    [field: SerializeField]
+    public TextAsset NarrativeScript { get; set; }
```
ViMaSter, line 19: I actually really like this change.  
It's technically not required to have a public setter for game code; it's only editable via the Inspector because of `SerializeField`. (By having the `SerializeField`-attribute, it produces the same outcome, but technically still changes just how encapsulated the underlying field really is.)  
Logic would dictate, that exposing this field is a change only needed to make tests work, and changes that only affect test code, are generally a bad idea.

With that said: I actually prefer this over having to work around it with `TestTools.SetField`. Couldn't really tell you why though. : D    
Good to see a change to help me settle a mental debate! 👍

ViMaSter (approved):
We like... have a game now? 👀 

Always love user-impacting changes! <sup>(especially because I rarely produce them 🙃)</sup>

notexplosive (approved):
😎🎮    📺

gameing

PR #268, opened 2022-06-28
bugfix/217-Scripts-containing-only-knots-give-an-error-when-validate

### LINKED ISSUE #217: [BUG] Scripts containing only knots give an error when validated
## Describe the bug
If a script contains only knots (https://github.com/inkle/ink/blob/master/Documentation/WritingWithInk.md#3-knots) it won't compile to a a JSON file. This is because if you divert to a knot in another file, the compiled JSON will just include the contents of the knot in the JSON file.

However, when validating files _all_ Ink files. The files containing only knots cannot find an associated JSON file and so give an error.

## Steps To Reproduce
Create an Ink file containing only a knot. Witness an error when it is validated.

## Expected behavior
Technically these files do not need to be validated as they will be parts of other files that will be validated. However it could be nice if you could write a file like this and get instant validation like with other files.

PhoebeMitchell:
## Summary

closes #217 

## Testing instructions

Edit a script with no associated JSON asset (e.g. anything in Resources>InkDialogueScripts>Templates). Does it give an error when you edit it?

## Additional information

- `[ ]` Changes UI


ViMaSter (approved):

notexplosive (approved):
gotta go fast

PR #293, opened 2022-09-24
Read me fix

NightRoost:
## Summary
renumbered the list of requirments


ViMaSter (approved):

PhoebeMitchell (approved):

notexplosive:
FWIW: Markdown doesn't actually care about if the numbers are correct. You could do `1.` for every item in a list and it will still render as `1, 2, 3, 4`

```
1. Like
1. this
1. for
1. Example
```
renders as:

1. Like
1. this
1. for
1. Example

ViMaSter:
Integrated as part of #321, as the checks for PRs can't complete for forks. (Otherwise, anyone could extract secrets from our repository workflow.)

PR #329, opened 2022-12-18
RELOAD_SCENE action now restarts the player from the currently active script

### LINKED ISSUE #316: [BUG] Reload scene action doesn't reload the scene properly
## Describe the bug
the TMPHGameOver narrative script uses the RELOAD_SCENE action to restart a cross examination.

However, what it actually does it just reload the unity scene so the cross examination doesn't play properly.

## Steps To Reproduce
Go to part 3 and get a game over by presenting incorrect evidence.
Skip through the game over dialogue.
Witness how the cross examination doesn't restart.

## Expected behavior
The game should restart at the beginning of the cross examination.

ViMaSter:
## Summary
This PR resolves the issue described inside #316.
It also resolves an issue with the test pipelines where Code Coverage wasn't properly posted to codecov.io in all commits after 51f13304a7052e1443c79b8b861038df1524b621 (which updated the CI test runner).

Before this change, a call to `RELOAD_SCENE` loaded a new instance of the `Game` scene and unloaded the currently active instance afterward.  
This resulted in two possible outcomes both of which were unexpected.  
When encountering `RELOAD_SCENE`...
- **after starting the `Game` scene using the `DebugGameStarter` to jump to a specific scene**, the newly loaded `Game` scene would still contain that `DebugGameStarter` and restart the game from that script 
   (for example: start the game with the `DebugGameStarter` set to `1-5`, fail the cross-examination in `1-6` enough times until your game is over, and notice how the game will return to `1-5` instead)
- **after starting the `Game` scene using the `MainMenu`**, the game will actually freeze; the `Game` scene is reloaded, but the `SceneLoader` starting the game at `1-1` is actually part of the `MainMenu` and therefore missing

## Testing instructions
**Case 1:**
1. Start the game from the `New Game` button (you can change the button to start the game at `1-5`, 
to skip most linear content)
2. Fail the cross-examination inside `1-6`
3. Notice the game starts from the cross-examination, rather than freezing with "Lorem impsum..." placeholder text

**Case 2:**
1. Start the game from the `Game` scene; set the `DebugGameStarter` to `1-5`, to skip most linear content
2. Fail the cross-examination inside `1-6`
3. Notice the game starts from the cross-examination, rather than the witness testimony

## Additional information

- `[ ]` Changes UI
  - closes #316

### DIFF
--- a/unity-ggjj/Assets/Scripts/SceneLoading/SceneLoader.cs
+++ b/unity-ggjj/Assets/Scripts/SceneLoading/SceneLoader.cs
@@ -6,8 +6,6 @@ using UnityEngine.SceneManagement;
 
 /// <summary>
 /// Class that contains methods for loading scenes.
-/// ChangeScene has overloads for scene index and scene name.
-/// LoadScene coroutine keeps track of the progress of the scene loading.
 /// Unloads the current scene after new scene has loaded.
 /// </summary>
 public class SceneLoader : MonoBehaviour, ISceneLoader
@@ -67,6 +65,9 @@ public class SceneLoader : MonoBehaviour, ISceneLoader
     /// <summary>
     /// Called by a transition controller to load the next scene after a transition.
     /// </summary>
+    /// <remarks>
+    /// This wrapper method is required to control SceneLoader instances via UnityEvents mapped inside the editor
+    /// </remarks>
     public void LoadSceneCallback()
     {
         StartCoroutine(LoadSceneCoroutine());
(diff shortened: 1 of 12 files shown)


unity-ggjj/Assets/Scripts/SceneLoading/SceneLoader.cs
```diff
@@ -67,39 +78,82 @@ private void Transition()
...
+            {
+                continue;
+            }
+
+            if (debugGameStarter.narrativeScript == null)
+            {
+                continue;
+            }
+            
+            Debug.LogWarning($"The scene '{scene.name}' loaded by this {nameof(SceneLoader)} contained a '{nameof(DebugGameStarter)}'. As a {nameof(NarrativeScript)} was specified when loading this scene, this needs to be removed, otherwise '{debugGameStarter.narrativeScript.name}' (the {nameof(NarrativeScript)} specified inside the {nameof(DebugGameStarter)}) will incorrectly take precedence over '{NarrativeScript.name}' (the {nameof(NarrativeScript)} specified inside the {nameof(SceneLoader)})");
```
notexplosive, line 132: jesus what an error message (no action, just remarking)

notexplosive (approved):

unity-ggjj/Assets/Scripts/SceneLoading/SceneLoader.cs
ViMaSter: [_ImGui has entered the chat._](https://github.com/ocornut/imgui/blob/master/imgui.cpp#L14359)

Was unsure whether to make it a comment, but this was actually an issue I ran into multiple times, so I figured it was worth a warning.

PR #351, opened 2023-03-25
Allow setting of poses of off-screen actors

### LINKED ISSUE #233: [BUG] SET_POSE cannot set pose for offscreen actor after that actor is used in another scene
## Describe the bug
Arin appears in two BGScenes during trials in case1: TMPHCourt and Anime.
If Arin appears in Anime, then a warning will occur claiming the animator is on an inactive gameobject when `&SET_POSE:Point,Arin` in TMPHCourt

## Steps To Reproduce
Create an Ink script containing the following:

```
&SCENE:TMPHCourt
&SET_ACTOR_POSITION:1,Arin
&JUMP_TO_POSITION:2
&SCENE:Anime
&ACTOR:Arin
Example text
&SCENE:TMPHCourt
&SET_POSE:Point,Arin
```

## Expected behavior
The above script should be playable without warnings, and the actor should be able have poses set from off screen.

## Additional context
Current workaround: Call `&SET_ACTOR_POSITION:1,Arin` again after the Anime scene ends and things works as expected.
Would be nice if that wasn't necessary though.

ViMaSter:
## Summary
Fixed bug where poses can't be set if an actor is off-screen and caused a warning to occur.

N/A

## Testing instructions
1. Checkout this branch
2. Run the `ActorControllerTests::SetPoseOnInactiveActor()` test
3. Verify that the test passes
4. Revert the commit containing the fix (bd9c6c6bde03f495fe3281320c0ef994b7bb4693) and verify that the test fails

## Additional information
- `[x]` Changes UI
  - closes #233


notexplosive (approved):
yaaay!

PR #353, opened 2023-04-09
Resolve issue added with #351

### LINKED ISSUE #233: [BUG] SET_POSE cannot set pose for offscreen actor after that actor is used in another scene

ViMaSter:
## Summary
Not entirely sure how GitHub even allowed integration of it after the tests failed, but #351 introduced a bug that [reliably fails the playthrough tests](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/actions/runs/4544200609/jobs/8009912558) - and thereby the playthrough.

The change in #351 to resolve #233 temporarily ensures a named actor and **all parent GameObject up to the root** (i.e. `Dan` -> `TMPHAssistant` -> `BGSceneList`) are active and visible to work around [Unity's limitation of being unable to start animations on disabled GameObjects](https://issuetracker.unity3d.com/issues/if-animation-is-played-before-activating-inactive-game-object-it-will-never-play-animations-when-game-object-will-be-active). This was needed, if the actor is in another scene:
- BGSceneList (**enabled**)
  - TMPHCourt (**enabled**)
    - Arin (**enabled** - current `ActorController._activeActor`)
  - TMPHAssistant (**>disabled<**)
    - Dan (**enabled**, but **>implicitly disabled<**)

If the Actor and all containing GameObjects were **already enabled**, it would incorrectly disable the top-most GameObject.

This Pull Request changes it only be true, if the object we're tracking is actually not active.

## Testing instructions
1. Checkout the target branch of this Pull Request
2. Run the Playthrough tests
3. Notice them failing
4. Checkout the source branch of this Pull Request
5. Run the Playthrough tests + new `SetPoseOnActorThatIsAlreadyActive` test
6. Notice them succeeding
7. Revert the commit containing the fix (b6de9da57d2e685695de308a84214cb2f7bdb7a7)
8. Run the Playthrough tests + new `SetPoseOnActorThatIsAlreadyActive` test
9. Notice them failing

## Additional information

- `[ ]` Changes UI


unity-ggjj/Assets/Scripts/Actor/ActorController.cs
```diff
@@ -113,36 +113,29 @@ private Actor FindActorInInventory(string actorName)
...
         {
-            Debug.LogError("Actor has not been assigned");
-            return;
+            highest = highest.transform.parent.gameObject;
+        }
+        var wasDeactivatedBefore = !highest.activeSelf;
+        if (wasDeactivatedBefore)
+        {
+            highest.SetActive(true);
+        }
```
notexplosive: q: Do we need this if-check? Can we just always call `highest.SetActive(true)` at first? Either it's not active and we make it active (desirable) or it's already active in which case `SetActive(true)` is a no-op... right?

It looks like this code was already here before your changes, just dropping my 2c

notexplosive (approved):
bippity boppity

unity-ggjj/Assets/Scripts/Actor/ActorController.cs
ViMaSter: Sure don't! Good catch, removed it in 89dab4b65cb525202b580119adf9aeba11b5ca6!

PR #359, opened 2023-06-09
Feature/highlights for controls buttons

PhoebeMitchell:
## Summary

Adds a highlight to buttons in the controls menu so it is clear they are selected when rebinding them.

Also fixes a bug where input actions using the same key meant that key could not be bound to i.e. if you bound "select" to another key, you could not bind bind anything to the X key despite.

Also removes an apparently unused method.

controls

## Testing instructions

Try and rebind some stuff. Does the colour change?
Rebind "select" to another key. Try and rebind it back to X. Does it work?

## Additional information

- `[x]` Changes UI


lyndsiWilliams:
Heya! I pulled this branch down and tested it, it all works as designed! _But_ you're also able to change more than one control at a time. Is that intended? It could cause some issues if multiple controls are mapped to the same thing, it might be good to put a safeguard that only allows one key to be mapped to an option, and/or one keymap to be changed at a time.

multiControlChange

lyndsiWilliams (approved):
Awesome, that fixed it! I think the only thing missing is user feedback if they try to assign a control to a key that's already taken. [I've created an issue so it can be done in another PR](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/issues/365). Great job on this! ✨

PR #367, opened 2023-08-08
Refactor/sceneloader

PhoebeMitchell:
## Summary

Refactors the SceneLoader component to be simpler and more efficient when switching scenes.
SceneLoader has been split into
- SceneLoader: Currents any given scene with a transition if specified.
- GameLoader: Uses scene loader to load the game script with a given narrative script.

Transferring a narrative script between scenes is now handled using a ScriptableObject, as opposed to the method of additive loading scenes and searching all game objects for the NarrativeGameState object.

DebugGameStarter has been replaced by GameStarter which either starts the game from the StartGameSettings scriptable object instance, or from a debug narrative script if one has been assigned.

Some now redundant tests have been removed.

## Testing instructions

Run the game. Can you still start the game? Can you select a specific part of case 1 to play using the case select menu? Do the tests pass?

## Additional information

- `[ ]` Changes UI


unity-ggjj/Assets/Scripts/GameState/GameStarter.cs
```diff
@@ -0,0 +1,50 @@
...
+        private NarrativeGameState _narrativeGameState;
+    
+        private void Awake()
+        {
+            _narrativeGameState = GetComponent<NarrativeGameState>();
+        }
+
+        private void Start()
+        {
+            if (_debugNarrativeScriptTextAsset != null)
```
lyndsiWilliams, line 21:
```suggestion
            if (_debugNarrativeScriptTextAsset)
```
I think this should work the same without `!= null`

unity-ggjj/Assets/Scripts/GameState/GameStarter.cs
```diff
@@ -0,0 +1,50 @@
...
+            {
+                Debug.Log($"DebugGameStarter: Running script {_debugNarrativeScriptTextAsset.name}");
+                _narrativeGameState.NarrativeScriptStorage.NarrativeScript =
+                    new NarrativeScript(_debugNarrativeScriptTextAsset);
+                _narrativeGameState.StartGame();
+            }
+            else
+            {
+                if (_gameStartSettings == null)
```
lyndsiWilliams:
```suggestion
                if (!_gameStartSettings)
```

unity-ggjj/Assets/Scripts/GameState/GameStarter.cs
```diff
@@ -0,0 +1,50 @@
...
+                {
+                    Debug.LogWarning("No debug narrative script text asset or GameStartSettings instance assigned. Game will not start.");
+                }
+
+                var narrativeScriptTextAsset = _gameStartSettings.NarrativeScriptTextAsset;
+                if (narrativeScriptTextAsset != null)
```
lyndsiWilliams:
```suggestion
                if (narrativeScriptTextAsset)
```

unity-ggjj/Assets/Scripts/SceneLoading/SceneLoader.cs
```diff
@@ -1,159 +1,28 @@
...
-            if (narrativeGameStateInNewScene != null)
-            {
-                break;
-            }
-        }
-        
-        if (narrativeGameStateInNewScene == null)
-        {
-            throw new NullReferenceException($"The newly loaded scene '{scene.name}' has no root object containing a '{nameof(NarrativeGameState)}'-component\r\nThis is required to initialize the next '{nameof(NarrativeScript)}'");
+            if (_transition == null) { return; }
```
lyndsiWilliams, line 23:
```suggestion
            if (!_transition) { return; }
```

lyndsiWilliams (approved):
Beautiful refactor, great job on this! I found a couple of teensy non-blocking conditional changes that could be adjusted, otherwise it looks great!

...unless you're checking for null specifically in these instances, and not just truthy/falsy values. In which case, disregard those bits 😅

unity-ggjj/Assets/Scripts/SceneLoading/FadeToImageTransition.cs
```diff
@@ -42,8 +39,8 @@ private void Start()
     /// Call this method to begin the transition at the end of a scene.
     /// Used by SceneLoader which does not necessarily know how a scene transition is being handled.
     /// </summary>
-    public void Transition()
+    public void Transition(Action callback)
     {
-        _imageFader.StartFade(0, 1, _fadeTime, _onTransitionOutComplete.Invoke);
+        _imageFader.StartFade(0, 1, _fadeTime, callback);
     }
```
ViMaSter: Using `UnityEvent` for `In` transitions, but `Action` for `Out` transitions feels inconsistent. I personally would prefer to stick to one type of event/action handler class.

Generally, my preference would be:
1. `interface` with an object and call a method on it directly (**+** tests can stub/mock that reference)
2. directly reference an object and call a method on it directly (**-** direct coupling, **+** allows setting it up once per Unity Scene via the Inspector)
4. `Action`/`event` (**+** allows setting up no/multiple events per Unity Scene via code)
3. use UnityEvents (**+** allows setting up no/multiple events per Unity Scene via the Inspector, **-** having to look into each scene and code file to find all bindings)

With 3. and 4. ensuring all expected events happen feels difficult without full integration tests.
Having said that, refactoring to get rid of events also feels like scope creep for this PR. 

If we need to stick with UnityEvents for now because we have Inspector-controlled bindings, I would prefer using [`UnityEvent.AddListener`](https://docs.unity3d.com/ScriptReference/Events.UnityEvent.AddListener.html) here instead and keeping it a UnityEvent.

unity-ggjj/Assets/Scripts/SceneLoading/GameLoader.cs
```diff
@@ -0,0 +1,25 @@
+using UnityEngine;
+
+namespace SceneLoading
+{
+    public class GameLoader : MonoBehaviour
+    {
+        private const string GAME_SCENE_NAME = "Game";
+        
+        [field: SerializeField] public TextAsset NarrativeScriptTextAsset { get; set; }
```
ViMaSter, line 9: More explicit variable names; love it! 👌

unity-ggjj/Assets/Scripts/SceneLoading/GameStartSettings.cs
```diff
@@ -0,0 +1,26 @@
...
+    public class GameStartSettings : ScriptableObject
+    {
+        private TextAsset _narrativeScriptTextAsset;
+        public TextAsset NarrativeScriptTextAsset
+        {
+            get
+            {
+                var narrativeScriptTextAsset = _narrativeScriptTextAsset;
+                _narrativeScriptTextAsset = null;
+                return narrativeScriptTextAsset;
```
ViMaSter: Logic this complex warrants a method for me; as a user, I would expect a getter to be as free of consequences as it can be.  
Something like `GetAndClearNarrativeScriptTextAsset()` is long, but would surface the implementation details whenever a reader encounters it in another file.

unity-ggjj/Assets/Scripts/GameState/GameStarter.cs
```diff
@@ -0,0 +1,50 @@
...
+                    _narrativeGameState.NarrativeScriptStorage.NarrativeScript =
+                        new NarrativeScript(narrativeScriptTextAsset);
+                    _narrativeGameState.StartGame();
+                }
+                else
+                {
+                    Debug.LogWarning("No narrative script text asset assigned to GameStartSettings instance. Game will not start.");
+                }
+            }
+        }
```
ViMaSter: [Via the Code style guidelines](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/wiki/Code-Style-Guideline#:~:text=Methods%20should%20first%20assert%20data%20is%20in%20an%20expected%20format%20and%20either%20return%20or%20create%20appropriate%20log%20messages): There's a few places in the PR where flipping the conditions and returning early makes code a bit more readable in my opinion. This method could be converted to...

```suggestion
        private void Start()
        {
            if (_debugNarrativeScriptTextAsset != null)
            {
                Debug.Log($"DebugGameStarter: Running script {_debugNarrativeScriptTextAsset.name}");
                _narrativeGameState.NarrativeScriptStorage.NarrativeScript =
                    new NarrativeScript(_debugNarrativeScriptTextAsset);
                _narrativeGameState.StartGame();
                return;
            }

            if (_gameStartSettings == null)
            {
                Debug.LogWarning("No debug narrative script text asset or GameStartSettings instance assigned. Game will not start.");
            }

            var narrativeScriptTextAsset = _gameStartSettings.NarrativeScriptTextAsset;
            if (narrativeScriptTextAsset == null)
            {
                Debug.LogWarning("No narrative script text asset assigned to GameStartSettings instance. Game will not start.");
                return;
            }

            _narrativeGameState.NarrativeScriptStorage.NarrativeScript =
                new NarrativeScript(narrativeScriptTextAsset);
            _narrativeGameState.StartGame();
        }
```

ViMaSter (approved):
Getting rid of additive scene loading basically completely eliminates having to keep track of potential side effects from scene switching. And I'm bad at keeping track of things. This is great! 👀

unity-ggjj/Assets/Scripts/SceneLoading/FadeToImageTransition.cs
PhoebeMitchell: As far as I'm aware the onTransitionOutComplete event isn't even used so I went ahead and removed it

unity-ggjj/Assets/Scripts/SceneLoading/GameStartSettings.cs
PhoebeMitchell: Agreed, and I was thinking that while i wrote it. Not sure why I made it a property.
Also added in an exception to avoid attempting access when you're not supposed to.

unity-ggjj/Assets/Scripts/GameState/GameStarter.cs
PhoebeMitchell: Good catch. That's way simpler :+1:

unity-ggjj/Assets/Scripts/GameState/GameStarter.cs
PhoebeMitchell: As per the discussion on Discord I'm going to keep it as it is for now

unity-ggjj/Assets/Scripts/SceneLoading/GameStartSettings.cs
```diff
@@ -12,15 +12,24 @@ namespace SceneLoading
...
+
+        public TextAsset GetAndClearNarrativeScriptTextAsset()
         {
-            get
+            if (_narrativeScriptTextAsset == null)
             {
-                var narrativeScriptTextAsset = _narrativeScriptTextAsset;
-                _narrativeScriptTextAsset = null;
-                return narrativeScriptTextAsset;
+                throw new InvalidOperationException("Stored narrative script text asset can only be accessed once. If you wish to access the narrative script text asset multiple times a reference must be stored locally.");
```
ViMaSter, line 22: Perfect occasion for a detailed exception! 👌

ViMaSter (approved):

PR #438, opened 2024-08-09
Adds more line information to parser exceptions

ViMaSter:
## Summary
Adds more debug information by adding the line the `ActionDecoder` is currently parsing to each exception.

## Testing instructions
Deliberately create a parser exception, notice the debug logs offering no info about the offending line before this PR and offering line info after.

## Additional information

- `[ ]` Changes UI

### DIFF
--- a/unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoderBase.cs
+++ b/unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoderBase.cs
@@ -29,6 +29,10 @@ public abstract class ActionDecoderBase : IActionDecoder
             var method = GenerateInvocationDetails(actionLine, GetType());
             method.MethodInfo.Invoke(this, method.ParsedMethodParameters.ToArray());
         }
+        catch (MissingParserException)
+        {
+            throw;
+        }
         catch (MethodNotFoundScriptParsingException)
         {
             throw;
--- a/unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoderBase.cs
+++ b/unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoderBase.cs
@@ -29,6 +29,14 @@ public abstract class ActionDecoderBase : IActionDecoder
             var method = GenerateInvocationDetails(actionLine, GetType());
             method.MethodInfo.Invoke(this, method.ParsedMethodParameters.ToArray());
         }
+        catch (MethodNotFoundScriptParsingException)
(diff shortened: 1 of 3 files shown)


unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoderBase.cs
```diff
@@ -24,8 +24,27 @@ public abstract class ActionDecoderBase : IActionDecoder
...
-        var method = GenerateInvocationDetails(actionLine, GetType());
-        method.MethodInfo.Invoke(this, method.ParsedMethodParameters.ToArray());
+        try
+        {
+            var method = GenerateInvocationDetails(actionLine, GetType());
+            method.MethodInfo.Invoke(this, method.ParsedMethodParameters.ToArray());
+        }
+        catch (MissingParserException)
+        {
+            throw;
```
lyndsiWilliams, line 34: I'm unsure about error specifications in C#, but could all of these catches get descriptive errors like line 46 (`actionLine`s, `e.Message`s, etc)? Or does `throw` give enough details in itself with these specific catches?

ViMaSter: It does! Unity's debugging window prints the `.Message` and `.StackTrace` of the exception and then recursively prints the `.InnerException` using the same pattern, until none are left. The order is a bit strange, but no information is lost by a call to `throw`:
image
(Using `throw e;` instead [_would_ actually obfuscate information](https://theburningmonk.com/2010/02/throwing-exceptions-the-right-way/).)

The `MissingParserException` for example currently has [this message](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/blob/a0581d3827f9e0ad9726675de4ff81b8d7b9fe3d/unity-ggjj/Assets/Scripts/TextDecoder/ActionDecoderBase.cs#L125).

The list of `catch` calls captures all exceptions that _already_ have line information and rethrows them directly. (As there's nothing more to add.)  
For any that don't (like a generic `IndexOutOfRangeException`), they are being wrapped in a `ScriptParsingException` by the final `catch` which also adds information about the line throwing the exception.

lyndsiWilliams: Oh that's awesome! In that case this looks good to me, thanks for explaining! 😁

lyndsiWilliams (approved):

PR #448, opened 2024-08-16
Fix bug where the speak action did not set the currently speaking actor

### LINKED ISSUE #417: Support multiple characters on screen at once
## Description
Writers want the ability to display two characters on screen simultaneously. This is slightly more complex than just "showing 2 characters at once" since we'd need to adjust their positions so they're taking up the left/right of the screen, rather than both sitting centered.

## Acceptance Criteria

This story will be finished when we can achieve an effect similar to shown in this Mockup.

This includes a way for writers to invoke this behavior via a Narrative Script.

image

PhoebeMitchell:
## Summary
Makes it so the speaking actor does not need to be the currently active actor. This removes the need to call JUMP_TO_POSITION to make an actor speak when multiple characters are on screen.

closes #417

## Testing instructions
All actors should continue to speak as before. In the court room scene actors should play the speaking animation even when off screen.


ViMaSter (approved):
Bugfixes _and_ package updates? Nice. 👌

PR #460, opened 2024-08-24
Support for `Talk` and `Move` inside investigation sections

ViMaSter:
## Summary
Since this was a work-in-progress branch, viewing individual commits isn't super useful. I've broken down my changes by files below, I'd recommend looking at those first and then at the remaining files.

### Known issues (done to keep PRs _somewhat_ small)

- `Examine` and `Present` are currently unhandled. They might work, they might not, and they're neither covered nor implemented. They're outlined inside #456 however.
- After selecting `Move` or `Talk`, the player can't go back to the main menu, other than selecting an option and playing through it. The logic is currently unhandled and might open up weird edge cases, as (aside from the evidence menu) we haven't had UI that allowed you to "go back one step" in the game loop. I'll probably leverage whatever we've used for the main menu setting menus for that in another PR.

#### Minor stuff up front:

- Reloading the game in the background was pretty annoying, so I bound the in-game editor and hot-reloading to `Ctrl/Command+R`, rather than just `R`.
  - `unity-ggjj/Assets/Scripts/Input/DebugControls.cs`
  - `unity-ggjj/Assets/Tests/PlayModeTests/Suites/Scripts/RuntimeEditingTests.cs`

-  Adjusted the default audio levels; open up your average YouTube video, set your volume to a comfortable level, and the game should(!) no longer blow out your eardrums.
  This _is_ reflected in the settings menu however and probably shouldn't. Ideally the game starts at 80% master and 100% for the rest, no matter our internal mixing. I'll create a separate issue for that.  
  Also added ducking/side-chaining of the music whenever SFX play.  
  image
    - `unity-ggjj/Assets/GameAudioMixer.mixer`
  
---

#### The new game state: Investigation. This PR adds `Talking` and `Moving` based on #456.

- Reference script: https://gist.github.com/ViMaSter/7606b62d0405d173ddfa6e51fb2ed671  
(also added under `unity-ggjj/Assets/Tests/PlayModeTests/TestScripts/InvestigationUI.ink`). 
- Demo:
  
  https://github.com/user-attachments/assets/267ae133-3ada-4493-a3c6-42a3a73ec312

Technical details of the implementation:

[...]

PR #494, opened 2024-09-24
Show dialogue choice menus again

ViMaSter:
## Summary
Resolves a regression discovered by @PhoebeMitchell: Choice menus outside investigations were no longer showing up and players automatically continued as if they chose the first action.

Running…

```ink
&SCENE:TMPH_Court
&JUMP_TO_POSITION:Defense
&SET_POSE:Thinking
&THINK:Arin
Hm...
&HIDE_TEXTBOX

+ [Object!]
    -> playerObjects
+ [Nah, we gucci]
    -> playerDoesntObject

=== playerObjects ===
&OBJECTION:Arin
&PLAY_SFX:deskSlam
&PLAY_EMOTION:DeskSlamAnimation
&SPEAK:Arin
My client's behavior during the livestream isn't suspicious in the least!

-> END

=== playerDoesntObject ===
&SPEAK:Arin
Nah, it's fine. Polishing some coins doesn't prove he's guilty.
-> END
```
### Before

https://github.com/user-attachments/assets/48a6ad77-e5c7-409b-8912-3e8474ec8251

### After

https://github.com/user-attachments/assets/42ba2277-c641-46ee-8085-bd471733d0e8

## Testing instructions
1. Open the `Game` scene
2. Set the `GameStarter` to `ChoiceMenu`
3. Notice the choice menu opening after advancing the game
4. Copy the `ChoiceMenu` script to your Desktop
5. Switch to `develop`
6. Copy the `ChoiceMenu` script back to Unity and select it again
7. Notice no choice menu opening and Arin immediately objecting after advancing the game

## Additional information
The issue stems from [this change](https://github.com/Studio-Lovelies/GG-JointJustice-Unity/commit/720a14851fa2a739fe079a964f050512fc9667d7) to `ChoiceMenu.cs` introduced by me as part of #460.

The InvestigationState injects logic into the ChoiceMenu to handle things like the `Move` preview image or checkboxes for `Talk` options. This requires the menu to be populated right after calling `ChoiceMenu.Initialise()`.

However this makes the `ChoiceMenu` immediately listen to input. "The player pressing `X` to advance to the choices", becomes "the player pressing `X` to confirm a choice" as well, as they both happen in the same frame.

To resolve this, I have passed the logic from the investigation state into a callback the ChoiceMenu calls after the buttons are created after a one frame delay. This resolves the regression while retaining the new functionality. (79942516217eab8e93ad7ad87831ec0880fdef98)

I've also added an automated test to catch this in the future. Running this test on `develop` right now will reliably fail, while it succeeds on this branch. (16793b45c9161a2f10b8031cf3cd9f1d06e835b7)

A different test failed after my initial submission of this PR, as I didn't migrate all parts of code to the callback; this was remedied in 0ca1a92d53b045df7bc28ca5839656d0b6e1d50d.

- `[ ]` Changes UI


unity-ggjj/Assets/Scenes/Game.unity
```diff
@@ -9523,7 +9523,7 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   m_Material: {fileID: 0}
-  m_Color: {r: 1, g: 1, b: 1, a: 1}
+  m_Color: {r: 1, g: 1, b: 1, a: 0}
```
PhoebeMitchell, line 9526: What is this? Why is it invisible now.

unity-ggjj/Assets/Scripts/GameState/InvestigationState.cs
```diff
@@ -110,7 +107,7 @@ public void OpenTalkMenu()
                 
                 _examinedTalkChoices.Add(_narrativeGameState.SceneController.ActiveSceneName + "_" + menuItem.Text);
             });
-        }
+        });
     }
 
     [FormerlySerializedAs("_unknownLocationBackground")]
```
PhoebeMitchell: Not part of this PR but the formerly serialized attribute should be removed and this field should be moved to the top of the file.

unity-ggjj/Assets/Scripts/GameState/InvestigationState.cs
```diff
@@ -148,27 +141,34 @@ public void OpenMoveMenu()
                 {
                     return;
                 }
-                
+
+                var sceneImage = _investigationMoveMenu.transform.parent.Find("SceneImage").GetComponent<Image>();
```
PhoebeMitchell: _investionMoveMenu should have a direct reference to SceneImage's Image component

unity-ggjj/Assets/Scenes/Game.unity
ViMaSter: ~~Can't not read this as https://youtu.be/dTBY9SdWd9Y?t=2426~~

Since the initialization of the ChoiceMenu happens in a Coroutine now, the preview sprite changes only after the menu container is opened; it would show the unchanged default sprite (a giant white box) for 1 frame. To work around this, I set it invisible by default and explicitly set the opacity back only after changing the sprite.

unity-ggjj/Assets/Scripts/GameState/InvestigationState.cs
ViMaSter: Good catch; removed in 718fd1db4470e3437e5ef794520fc167439ae28b and relocated in 174f0a5bc9eb7539be2fbe237a4deaa7b8b53d75.

unity-ggjj/Assets/Scripts/GameState/InvestigationState.cs
ViMaSter: Resolved in 6cab5b8c0ca4383eee100b38c780f513d666397f. Adding a field used only once to the `ChoiceMenu` felt weird; I've created a new class `InvestigationChoiceMenu` instead which derives from `ChoiceMenu` and added that field to it.

unity-ggjj/Assets/Scripts/Menu/ChoiceMenu/InvestigationChoiceMenu.cs
```diff
@@ -0,0 +1,6 @@
+using UnityEngine.UI;
+
+public class InvestigationChoiceMenu : ChoiceMenu
+{
+    public Image _sceneImage;
```
PhoebeMitchell: Should be a property and in PascalCase

unity-ggjj/Assets/Scripts/Menu/ChoiceMenu/InvestigationChoiceMenu.cs
```diff
@@ -0,0 +1,6 @@
+using UnityEngine.UI;
+
+public class InvestigationChoiceMenu : ChoiceMenu
```
PhoebeMitchell: Typically elsewhere we've chosen to have a reference to the base component instead of inheritance e.g. `InvestigationChoiceMenu.ChoiceMenu.Foo`.... but i can think of at least one place we're using inheritance instead so eh don't mind.

PhoebeMitchell (approved):

unity-ggjj/Assets/Scripts/Menu/ChoiceMenu/InvestigationChoiceMenu.cs
ViMaSter: Well, there's really no technical requirement for inheritance, so I've changed it to composition in 2aeefcef3a158c0bfbb9aec87c976ef5efbb7117. :)

unity-ggjj/Assets/Scripts/Menu/ChoiceMenu/InvestigationChoiceMenu.cs
ViMaSter: Addressed in 2aeefcef3a158c0bfbb9aec87c976ef5efbb7117.

PR #497, opened 2024-10-30
Adds `&LOCK_MOVE_CHOICE` and `&LOCK_TALK_CHOICE` actions

ViMaSter:
## Summary
Adds `&LOCK_MOVE_CHOICE` and `&LOCK_TALK_CHOICE` actions.

## Testing instructions
1. Open the Game scene
2. Set the `GameStarter` to `InvestigationUI`
3. Select talk option 2 and notice talk option 3 being available
4. Select talk option 3 and notice talk option 2 missing afterwards
5. Select talk option 1 and notice talk option 2 reappearing

## Additional information

- `[ ]` Changes UI
  - #494


unity-ggjj/Assets/Scripts/GameState/InvestigationState.cs
```diff
@@ -22,42 +22,63 @@ public class InvestigationState : MonoBehaviour, IInvestigationState
...
     private List<Choice> _talkOptions;
     private List<Choice> _moveOptions;
     
+    private readonly List<string> _explicitlyLockedTalkChoices = new();
+    private readonly List<string> _explicitlyLockedMoveChoices = new();
     private readonly List<string> _unlockedTalkChoices = new();
     private readonly List<string> _unlockedMoveChoices = new();
     private readonly List<string> _examinedTalkChoices = new();
     private readonly List<string> _examinedMoveChoices = new();
     private readonly List<string> _examinedDetails = new();
```
notexplosive: idea: It looks like we have a bunch of parallel lists that are updated together to keep in sync. I feel like there's a better way we could represent this state.

Instead of having multiple list of parallel strings, what if we just have 3 lists of:
- Talk Choices
- Move Choices
- Examined Details

and each of those lists held richer objects that knew about their state (eg: `IsExplicitlyLocked` `IsUnlocked` etc).

unity-ggjj/Assets/Scripts/GameState/IInvestigationState.cs
```diff
@@ -3,10 +3,20 @@
...
+    {
+        Talk,
+        Move
+    }
+
+    public enum ChoiceTag
+    {
+        None,
+        Locked
+    }
```
notexplosive: suggestion: Move these enums out of IInvestigationState, and maybe prepend with the word `Investigation`? I feel like at the call site I find `IInvestigationState.ChoiceTag.Locked` harder to read than `InvestigationChoiceTag.Locked`

(totally feel free to push back if you disagree!)

notexplosive (approved):
couple broad thoughts about structure but the changes themselves look good!

unity-ggjj/Assets/Scripts/GameState/InvestigationState.cs
ViMaSter: Resolved in b7d0a7785bc15e3fea63125b068e62667683f2d9.

Attempted to also integrate `_talkOptions` and `_moveOptions`, but I didn't want to merge "global game state" and "current choice state" into one state container.

unity-ggjj/Assets/Scripts/GameState/IInvestigationState.cs
ViMaSter: Resolved in 29cad0380d4e27b2beb6d1170d6fdbc7ef6e1526.

unity-ggjj/Assets/Scripts/GameState/InvestigationState.cs
notexplosive: Very nice! feels way more readable to me! (also nice use of `record` despite the fact that unity is on a "middle-child" language version where `record` exists but `record` isn't cool yet)

